PR #15, opened 2023-03-16
Supporting matrix with multiple groups

airenzp:
Added menu option to display matrix for several groups.


xzhou82 (approved):
thanks it works! next may change column width to be automatic so the columns can be wider when the number of samples is few

PR #18, opened 2023-03-16
reorganize the container code and readme 

siosonel:
based on usage and GHP registry tests

@aacic you have another branch that may conflict with changes that I have, we should coordinate our changes in build.sh, for example.


container/README.md
```diff
@@ -63,41 +104,54 @@ npx proteinpaint-container server
...
 ./pack.sh
 
 # will run docker build
 ./build.sh
-```
 
-### Testing and Debugging
+TAG=latest # can change to a version number like 2.11.2
+IMAGE_NAME=full:$TAG # may use server:$TAG for server-only image
+docker pull ghcr.io/stjude/proteinpaint/$IMAGE_NAME
```
aacic, line 122: We would have to agree on the docker image name format.
It can be also like:

`ghcr.io/stjude/ppfull-$TAG`

Like here:
https://github.com/stjude/proteinpaint/pkgs/container/ppfull

@robinpaul85 @siosonel  which one is better? WDYT?

aacic (approved):

PR #21, opened 2023-03-17
handle cases where a global or local filter may have 2+ entries

siosonel:
in the root tvslst, and reuse a genomebrowser filterUI if applicable

@xzhou82 The actual issue is that `getFilterExcludingPill($id)` mistakenly assumed that only up to 2 entries can exist in a root filter data, a cohort filter and/or the `filterUiRoot`. We now see that a local filter can have 3 entries: the cohort filter, the global filter UI data, and the local filter UI data.

I did not push to master because there may be other use cases where only the global cohort filter may be needed for a different type of track or chart, so other entries in the root filter was intentionally excluded? This will need to be tested more, even though all filter and tvs integration test pass.

### DIFF
--- a/client/plots/genomeBrowser.controls.js
+++ b/client/plots/genomeBrowser.controls.js
@@ -28,6 +28,7 @@ class GbControls {
 		.holder
 		*/
 		this.type = 'gbControls'
+		this.filterUI = {}
 	}
 
 	async init(appState) {
@@ -60,7 +61,8 @@ class GbControls {
 	}
 
 	async main() {
-		if (this.state.config.snvindel?.details?.groups) {
+		const groups = this.state.config.snvindel?.details?.groups
+		if (groups) {
 			// is equipped with comparison groups, render group ui
 			this.render1group(0)
 			this.render1group(1)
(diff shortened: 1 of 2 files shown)


client/plots/genomeBrowser.controls.js
```diff
@@ -96,7 +98,7 @@ class GbControls {
 				}
 			})
 		const div = this.dom.variantFilterHolder.append('div').style('display', 'none')
-		filterInit({
+		this.filterUI.variant = await filterInit({
```
xzhou82, line 101: the `variant filter` should be stateless as the filter is consisted of VCF INFO fields but not dict terms, thus should not be affected by state change

siosonel: ok, I'll fix. Are there places besides mass app, where a cohort filter + other filter uses `filterInit()`? For example, would `block.mds2.vcf.numericaxis.AFtest.js` break with the change to `getFilterExcludingPill()`? I'm not familiar enough where to look for that usage.

xzhou82 (approved):
thanks a lot the refactoring works well

PR #23, opened 2023-03-17
Violin list samples

karishma-gangwani:
This should fix the list sample issue and also show values as 0.


xzhou82 (approved):

xzhou82:
when a dataset is unsecured, `List samples` shows for bar clicking, but will not show for violin brushing (it should)

PR #27, opened 2023-03-17
Tests 7

creilly8:
- Fix for failing scatter test
- test app prototype
- New vocab.unit and updated vocab.int tests


siosonel:
@creilly8 let me know if you need help with these requests
- please redo this commit, so that the message does not say "initial commit"? That message is typically found at the very start of the commit history
- as my prettier hook was not working when I worked on the sampleScatter code, are you able to split the changes into two: one to apply the prettier formatting, and then the next to apply your updates, so that it will be easier to ignore the formatting-only changes

siosonel (approved):

PR #35, opened 2023-03-18
Wait until pp server inside docker container starts running.

aacic:


aacic:
@robinpaul85 @siosonel 

The server in the docker container is started after 20 seconds on my computer, so when the run script is executed the server is still not online which is confusing, so I've improved this script so it waits until the server goes online.

container/run.sh
```diff
@@ -76,3 +76,24 @@ docker run -d \
 	$IMAGE_NAME
 
 echo "^ assigned container ID ^"
+
+container_id=$(docker ps -q -f name=pp)
+end_time=$((SECONDS+30))
+
+echo "Waiting for 'STANDBY AT PORT $HOSTPORT' in container logs..."
```
siosonel, line 83: @aacic should detect `'STANDBY AT PORT $EXPOSED_PORT'`. The HOSTPORT is for the URL that the user loads in his browser, the EXPOSED_PORT is the port that the express app listens to inside the container. 

For simplicity, we may force HOSTPORT to match the EXPOSED_PORT, but we'll need to document that `serverconfig.port` may be ignored inside the container if `serverconfig.url` is present and has a different port. Although, a user could also manually edit the serverconfig.port to match the port # in the serverconfig.url.

siosonel: @aacic We may also consider running an attached container instead of detached, so that they see the logs of when the server is ready/listening.

siosonel:
Updated the output to
```bash
$ ./run.sh ppfull:latest
[/Users/esioson/gb/tp] [3456] [ppfull:latest] [pp]
finding any matching container process to stop and remove ...
38ef2df0140b
Starting container process='pp' ...
e613e474a6fab7b2995243b770c57a5b4bff98ced5f0ba8e86945d1452d22d6f
^ assigned container ID ^
Waiting for server validation ...
generating public/bin for http://localhost:3456
null
DB connected for MSigDB: anno/msigdb/db
msigdb initiated as hg38-level termdb

Sun Mar 19 2023 23:09:33 GMT+0000 (Coordinated Universal Time) 

Validation succeeded.

STANDBY AT PORT 3000

************************************************************************
*
* Open the ProteinPaint app at https://localhost:3456 in you web browser
*
*************************************************************************

Hints:
- inspect logs with 'docker logs pp'
- ssh into the container with 'docker exec -it pp bash'
- stop the container with 'docker stop pp'

```

siosonel (approved):

PR #43, opened 2023-03-21
Making matrix wider and with sample ids

airenzp:


client/plots/sampleScatter.js
```diff
@@ -1111,14 +1111,14 @@ class Scatter {
...
 						}
-
 						config.divideBy = {
 							term: { name: 'groups', type: 'samplelst', values },
 							q: {
 								mode: 'custom-groupsetting',
 								groups: qgroups
 							}
 						}
+						config.settings.matrix.colw = 10
```
xzhou82, line 1121: can you make it dynamically decide width, rather than hardcoding? see if matrix has an attribute to do it

airenzp: setting this property to 0 does that, but when the number of samples increases, say to 120 samples, is looking like this. 
Screenshot 2023-03-21 at 2 13 34 PM
And if the width is below 8 it does not show the sample id. So I think with 10 should work for most cases, groups are not that big.

airenzp: Also the sampleids are not too small or too big with this size.

siosonel: @airenzp you may want to improve the `setAutoDimensions()` function in matrix.js, instead of hardcoding.

xzhou82 (approved):

PR #46, opened 2023-03-22
fix tsne error when samples are all filtered out

congyu-lu:


congyu-lu:
Sorry VS code did some auto formating, the only change is in line 133

client/plots/sampleScatter.js
```diff
@@ -130,7 +129,7 @@ class Scatter {
 		const s0 = this.data.samples[0]
```
karishma-gangwani, line 129: can you change to a meaningful variable name?

client/plots/sampleScatter.js
```diff
@@ -655,31 +639,27 @@ class Scatter {
...
 				offsetY += 25
 			}
 			offsetX += step
-
 		}
 		return offsetY
-		
 	}
 
 	hideCategory(G, tw, key, hide) {
```
karishma-gangwani, line 665: this one too. is G a svg element?

karishma-gangwani (dismissed):

congyu-lu:
Screen Shot 2023-03-22 at 1 03 36 PM
after applying filter, clicking new session will throw an error

airenzp (approved):

client/plots/sampleScatter.js
airenzp: s0 is the first sample, is visible from the value it gets and also from the name(short of sample0).  I added a comment to this line to explain it. Let me know if you want me to rename the variable though.

client/plots/sampleScatter.js
airenzp: this line was added because of reformatting, so not related to this PR. It is an svg group, that is called g in svg.

airenzp:
> Screen Shot 2023-03-22 at 1 03 36 PM
> 
> after applying filter, clicking new session will throw an error

Fixed now, needed to be prepared to have no samples, in several pieces of the code.

PR #48, opened 2023-03-22
Fixed lasso bug

airenzp:


congyu-lu:
image

apply a filter -> lasso select all the samples as group1 -> edit filter -> lasso select all the samples as group2, -> the samples selected are the same as group1

PR #49, opened 2023-03-23
Matrix zoom

siosonel:
Add this entry to serverconfig.features: `"matrix": ["zoom"]` in order to see and test the feature. Open a small matrix plot to test, the zoom may be triggered via:

1. dragging a rect outline on the matrix: KNOWN BUG (will be fixed with switch to canvas image): the mouse start and stop must be on a rendered "series" rectangle, starting or stopping on an empty column/cell will cause an error
2. dragging the slider
3. entering a number 
4. clicking on the minus or plus buttons

Screenshot 2023-03-23 at 2 02 46 PM

Test zoom controls: http://localhost:3000/testrun.html?name=zoom.unit

### DIFF
--- a/client/dom/test/zoom.unit.spec.js
+++ b/client/dom/test/zoom.unit.spec.js
@@ -1,0 +1,110 @@
+import tape from 'tape'
+import { select } from 'd3-selection'
+import { zoom } from '#dom/zoom'
+
+/*************************
+ reusable helper functions
+
+/**************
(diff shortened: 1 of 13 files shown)


xzhou82:
can add later: column gap should have a setting called "automatic", so it can automatically add 1px column gap when column width >2px

xzhou82 (approved):

PR #54, opened 2023-03-24
initial implementation to prevent unauthorized filtering down to low …

xzhou82:
…sample count

please first pull sjpp master, then test on this branch; tests still work i think

1. when the number of samples passing filter <10 it shows an error and halts the portal
2. barcharts are frozen as it should be. hasn't tested the other plots yet
3. next user needs to change the filter to reenable portal, and manually dismiss the error. there isn't an automated way to dismiss it
4. for sjlife, if user signs in with access, then the user is allowed to filter to less than 10

TODO
1. mass nav: when triggering the error through adding a filter from barchart, the filter ui does not rerender to show the "offending" tvs. but it should so that user can delete it (also click Redo should work?)
2. should disable New session button when error is triggered, so not to let saving

### DIFF
--- a/client/mass/app.js
+++ b/client/mass/app.js
@@ -37,6 +37,7 @@ class MassApp {
 			holder: opts.holder, // do not modify holder style
 			topbar: opts.holder.append('div'),
 			errdiv: opts.holder.append('div'),
+			filterBelowMinSampleWarning: makeFilterBelowMinSampleWarning(opts.holder),
 			plotDiv: opts.holder.append('div')
 		}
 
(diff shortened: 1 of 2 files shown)


siosonel:
@xzhou82 The number of samples from the filterCTE query is only the upper limit of samples with data. There may even be less samples if not all of the filtered samples were annotated for all of the requested terms. The applicable number of filtered samples is the count right before a `get_rows*()` or `getData()` returns. This means that different chart sandboxes may have an error or not, depending on a chart inside it having >= min number of samples, so it's not a single error that applies to  all the charts.

xzhou82:
thanks a lot yes that makes sense. i will close this pr for now so we can reimplement it

PR #55, opened 2023-03-24
Fix docker build.

aacic:


siosonel:
some client-side changes may require matching backend changes, and vice-versa, so it would be incorrect to specify only one of releaseTag.server and releaseTag.front . Either both must be specified or none at all for app-full.js, I'll update the code after testing

siosonel (approved):

PR #56, opened 2023-03-24
Table row buttons

airenzp:
Draft to add action buttons to table rows

### DIFF
--- a/client/mds3/sampletable.js
+++ b/client/mds3/sampletable.js
@@ -84,7 +84,7 @@ export async function displaySampleTable(samples, args) {
 		const height = screen.height
 		const colButton = {
 			text: 'Disco',
-
+			//disabled: (i) => false,
 			callback: async (event, i) => {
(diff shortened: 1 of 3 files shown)


client/mds3/sampletable.js
```diff
@@ -75,8 +76,25 @@ export async function displaySampleTable(samples, args) {
 	const params = { rows, columns, div: args.div, resize: rows.length > 10 }
 	//if (args.maxWidth) params.maxWidth = args.maxWidth
 	//if (args.maxHeight) params.maxHeight = args.maxHeight
+
+	params.columnButtons = []
```
xzhou82, line 80: columnButtons should be conditional; only set if there are buttons to add for a dataset
for ash dataset it generates a blank column with no buttons as the dataset does not have single sample data

airenzp: Fixed now

xzhou82 (approved):

PR #64, opened 2023-03-28
Render matrix terms as series data

siosonel:
... to limit the number of rendered canvas images. Previously, rendering samples as series data creates many more images.


xzhou82 (approved):
works!

PR #70, opened 2023-03-29
Nav.groups

xzhou82:
the `GROUPS` tab is added to mass nav header
the ui uses table.js to display created groups
initial tests showed it's working for both pnet and sjlife
filter tests all passed
created groups work independently from global filter and subcohort
groups are tracked in state (state.groups[]) and can be saved
more functionalities will be added to `Launch plot`

some code duplication with sampleScatter, including function for creating samplelst tw, and launching new plots. later can improve

### DIFF
--- a/client/dom/table.js
+++ b/client/dom/table.js
@@ -259,6 +259,10 @@ export function renderTable({
 		}
 		for (const [colIdx, cell] of row.entries()) {
 			const td = rowtable.append('td').attr('class', 'sjpp_table_item')
+
+			// attach <td> for external code to modify
+			cell.__td = td
+
 			const column = columns[colIdx]
 			if (column.editCallback && cell.value) {
 				td.on('click', () => {
--- a/client/filter/filter.js
+++ b/client/filter/filter.js
@@ -387,6 +387,71 @@ class FilterRxComp extends Filter {
 
 export const filterRxCompInit = getCompInit(FilterRxComp)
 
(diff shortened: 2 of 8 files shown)


gavrielm:
The "Launch plot with Group 1" button makes it seem like a plot will be launched that is filtered for samples in group 1, but really group 1 is treated as an overlay term:
image

gavrielm:
Possibly rename the button to "Use Group 1 in chart"?

xzhou82:
i think it would be fine

gavrielm:
Maybe we should notify user that sample size of group gets reduced as a result of intersection. For example (see "Age > 10" group):

image

xzhou82:
yes good point

xzhou82:
should allow to user to see a venn when multiple groups are selected

gavrielm:
Yeah venn would be good. The total # patients also differs when sample groups are used:

image

gavrielm (approved):
I will approve. Works nicely. Can either add venn diagram option to this PR or next one.

siosonel (approved):

congyu-lu:
Works great in MBMETA, small issues noticed:
1. "negate" a joined filter gives a very large sample number. for example: 
Screen Shot 2023-03-29 at 5 03 44 PM
2. Is there a way to delete a group created directly instead of deleting its filter components one by one?

PR #76, opened 2023-03-31
Added needed imports after refactor and fix for assigning color

airenzp:
Fix to assign color to the sample for gene terms


xzhou82 (approved):
it works thanks a lot

PR #77, opened 2023-04-03
Scatter legend fix

airenzp:
Improved gene term legend as suggested. Fixed bug in color assignment,  when mutations are hidden _wildtype_ and _not tested_ categories should be included to assign color. Also showed hide/show in menu when needed.


congyu-lu:
Works great! A minor issue: There are 73 dots for CTNNB1 MISSENSE,  but the legend is MISSENSE: 74S

image

xzhou82 (approved):

PR #82, opened 2023-04-04
Showing search results on the fly

airenzp:
When typing letters matching samples are displayed. If you press Enter and there is a valid group of samples (not 0 and not all the samples from the cohort), the group is created/kept. If you press Escape you cancel the search.


xzhou82:
after canceling search, all ref dots are gone

xzhou82 (approved):

PR #83, opened 2023-04-04
Integrate disco plot code

aacic:


client/plots/disco/events.js
```diff
@@ -0,0 +1,283 @@
+import { tooltip } from '#src/client'
```
xzhou82, line 1: see if you can replace tooltip with `import {Menu} from '#dom/menu`

client/plots/disco/template.controls.js
```diff
@@ -0,0 +1,617 @@
+import { Menu } from '#src/client'
```
xzhou82, line 1: change to importing from '#dom/menu'

client/plots/disco/dt.disco.js
```diff
@@ -0,0 +1,282 @@
...
+import { sayerror, hideerrors } from '#src/client'
+
+// these are shared across instances
+const mlabel = {}
+for (var key in mclass) {
+	mlabel[mclass[key].label] = mclass[key]
+	mlabel[key] = mclass[key]
+}
+
+const fusionColocationColors = {
```
xzhou82, line 33: this is defined in multiple scripts

aacic: I've removed it from this script. It was not used.

client/plots/disco/events.js
aacic: I've tried to replace it with Menu, but the tooltip has additional methods like: `getPosChange`. I would say we keep it for now and I'll remove it when I refactor the whole disco plot. I can add a TODO for now.

aacic: Disco plot-related code is extracted to the helper.js script in the disco folder.

xzhou82 (approved):

PR #85, opened 2023-04-06
Barchart color

airenzp:
Reading default colors for categorical, numerical and gene terms. Supporting color change for all. Using getColors (in common.js) to assign color of none provided.

### DIFF
--- a/client/dom/html.legend.js
+++ b/client/dom/html.legend.js
@@ -1,9 +1,12 @@
 import { select } from 'd3-selection'
+import { Menu } from '#dom/menu'
+import { rgb } from 'd3-color'
 
 export default function htmlLegend(legendDiv, viz = { settings: {}, handlers: {} }, barDiv) {
 	const isHidden = {}
 
 	function render(data) {
+		console.log(data)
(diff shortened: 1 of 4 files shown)


congyu-lu:
Works great! Why there is a color picker for Statistical Significance?
image

siosonel:
So, maybe we always set a rect stroke in the barchart, which defaults to the fill color, but when white or "whitish" color, then make the stroke the color that you used (rgb(179, 179, 179)) . That way, we also maintain a consistent color schema as assigned in the terms table or for gene variant mclass

Screenshot 2023-04-06 at 3 12 30 PM

airenzp:
Thanks for the comments. @congyu-lu that item should not be editable. I will fix it soon. Thanks!

airenzp:
well, just fixed it now.

congyu-lu:
> well, just fixed it now.

I guess the white square is not necessary? below is what it looked like before. 
image

xzhou82 (approved):
thanks i will merge now. if any issue remains we can fix in new branches

PR #91, opened 2023-04-08
All logscale 3

karishma-gangwani:
Rendering still needs some tweaks and brushing/filtering for negative values seems off. need to discuss that. but label clicking and brushing on positive values works and scale was added to the controls.


client/plots/violin.interactivity.js
```diff
@@ -143,7 +145,16 @@ export function setInteractivity(self) {
 		for (const [c, k] of Object.entries(data.samples)) {
 			const sampleIdObj = {}
 			if (data.refs.bySampleId[c]) {
-				sampleIdObj[data.refs.bySampleId[c]] = k[term.$id].value
+				if (self.config.settings.violin?.unit === 'log' && k[term.$id].value != 0) {
+					sampleIdObj[data.refs.bySampleId[c]] = roundValue(
+						getBaseLog(
```
xzhou82, line 150: you can use `Math.log2()`

karishma-gangwani: well, I intentionally did this in case they wanted a different log base. It's a lot neater this way. If it changes then we won't have to replace the code everywhere and just update the db file.

PR #94, opened 2023-04-10
Bug fix for getFilteredSampleCount()

creilly8:


siosonel:
@creilly8 This will need a bit more testing, the samplecount should not be 0 in the filter nav tab.

Screenshot 2023-04-10 at 5 31 53 PM

siosonel:
@creilly8 I pushed a fixed in case data[0].samplecount is present

siosonel (approved):

PR #97, opened 2023-04-11
remove derelict flag of _is_genotype and join_on_clause

xzhou82:
i found out the `index` parameter of get_term_cte cannot be replaced by tw, thus only the minor cleanup
cuminc uses get_rows and CI works

### DIFF
--- a/server/src/termdb.sql.js
+++ b/server/src/termdb.sql.js
@@ -448,8 +448,6 @@ q{}
 	.term[0,1,2]_id 			// supported parameters for barchart data  
 	.term[0,1,2]_q 				// supported parameters for barchart data
 		the _q{} is managed by termsetting UI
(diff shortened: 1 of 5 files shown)


server/src/termdb.sql.js
```diff
@@ -544,7 +539,6 @@ export async function get_term_cte(q, values, index, filter, termWrapper = null)
 	} else {
 		throw 'unknown term type'
 	}
-	if (index != 1) CTE.join_on_clause = `ON t${index}.sample = t1.sample` // will be ignored if no join clause is created by sql constructor
```
siosonel, line 547: @xzhou82 I think `get_rows_by_two_keys()` may still need this `join_on_clause`

siosonel (approved):

PR #107, opened 2023-04-14
Color bars

airenzp:
Hiding color choices when there is no overlay and showing default color input when plain color wanted. Done in collaboration with @siosonel .

### DIFF
--- a/client/plots/barchart.js
+++ b/client/plots/barchart.js
@@ -129,6 +129,7 @@ class Barchart {
 				},
 				{
 					label: 'Color bars',
+					title: 'May use server-assigned color if available',
 					type: 'checkbox',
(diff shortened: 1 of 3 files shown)


client/plots/barchart.js
```diff
@@ -128,19 +128,23 @@ class Barchart {
 					]
 				},
 				{
-					label: 'Color bars',
+					label: 'Multicolor bars',
+					title: 'Color bars ussign preassigned colors from the dataset or a random color',
```
siosonel, line 132: Typo on `ussign'. But also, if there are no preassigned colors, then it will use the `defaultColor`, not a random color

siosonel (approved):

PR #108, opened 2023-04-14
wip create custom terms from Group filters

xzhou82:
tried a few CI and all can pass
todos
1. (done) upon adding a new term, state is changed but app is not notified (the term pill does not show in Group ui)
2. new term seems to be generating wrong charts
3. in termdb app.js,  show custom terms based on usecase (filter by term type)

i think it's fine to merge this, will continue to fix all issues next. if breaking things can't be fixed by monday, can go with nuclear option to disable "Groups" from nav.js altogether until after AACR

### DIFF
--- a/client/filter/filter.js
+++ b/client/filter/filter.js
@@ -1,8 +1,8 @@
-import { getInitFxn, getCompInit, Bus } from '../rx'
+import { getInitFxn, getCompInit, Bus } from '#rx'
 import { select } from 'd3-selection'
-import { Menu } from '../src/client'
+import { Menu } from '#dom/menu'
 import { TVSInit } from './tvs'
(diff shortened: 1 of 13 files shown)


siosonel (approved):

airenzp:
Nice that this is moving forward. Thanks! Looks good. The button names maybe could be simplified?  I am thinking something like:
Add new group --> Add group
Create new variable with 2 groups --> Create dictionary variable.  Users are more familiar with the dictionary maybe?) and then a tooltip with the details like 'Creates a dictionary variable using the groups selected...'
Name the new variable--> Name. Maybe this input could be moved to the left of the button, so that it looks more like a form. 

We can talk about it if it is not clear.

PR #114, opened 2023-04-19
Tvs numeric range

airenzp:
This draft PR improves the previous numeric filter UI to use a single input to enter the range values. Checking all the use cases still. The input is now a separated component to be reused, numericRangeInput.


xzhou82 (approved):
thank you so much.
one issue to be fixed next: at numeric tvs showing density plot, after updating it through <input>, drag on the slider in density plot is not updating the range in <input>

PR #124, opened 2023-04-23
Matrix in firefox and safari

siosonel:
- Fix zoom area not rendering in Firefox
- Fix dom canvas rendering in Safari which does not support OffscreenCanvas
- Fix grid rendering for quantitative/continuous terms
- More control options for grid outline, canvas min-pixel width detection
- Improved auto-width computation, legend offset


xzhou82 (approved):
thanks

PR #128, opened 2023-04-24
Display error if validation failed. Fix in unbounded case

airenzp:
Should allow to filterout edge cases, like the x=0 reported by Gav on the average dose to heart. Should show error message if tvs validation failed.

### DIFF
--- a/client/dom/numericRangeInput.js
+++ b/client/dom/numericRangeInput.js
@@ -46,8 +46,8 @@ export class NumericRangeInput {
 		//When an error is thrown the previous range is restored
 		else this.range = range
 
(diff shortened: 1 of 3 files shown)


gavrielm:
Thanks for the fix. I am wondering if the labels of the filters should not show a stop value when the filter is one-sided (i.e.`0 < x`). For example, I set these filters to `0 < x`, but it looks like they are bounded by a stop value:

image

airenzp:
Thanks for the tests @gavrielm , now the limit on the unbounded side is removed

xzhou82 (approved):

xzhou82:
looks great, thanks!

gavrielm:
Thank you!

PR #129, opened 2023-04-24
Numeric filter equal

airenzp:
Supporting x=value case in the numeric filter. Rebased on top of PR fixing the bug reported by @gavrielm to contain the latest changes. You can either merge that one first or close it and merge this one directly, once approved.


gavrielm:
`x = 0` and `0 = x` do not work for me

airenzp:
Now the case x=0 should be supported @gavrielm. Also supported edit of tvs with ranges of the form x=value, after adding a numeric filter edit tvs with x=value ranges was not supported. Bug fixes. Improvements in the UI

xzhou82 (approved):
thanks!

PR #134, opened 2023-04-25
Violin tvs 2

karishma-gangwani:
I made some changes to the rendering. I don't think log transforming the values on the backend is needed (checked ggplot2 rendering as per Xin's suggestion). the list samples are also working now (thanks to Edgar). I also made a small change to skip zero values for descriptive statistics computation based on Gavriel's suggestions. Let me know if this looks good.


xzhou82 (approved):
thanks it's looking correct now. keep going

PR #138, opened 2023-04-26
Using termwrappers in the custom terms and allowing term events to ha…

airenzp:
@siosonel  and I worked in this branch to: pass sample names when possible. Use termwrappers in the custom terms and allow term events in the charts or the tree to handle term wrappers. These changes were needed to pass the original termwrapper that contains q settings to, for example,  hide sample names. So far our tests worked and all the tests are passing, but please test more to cover all the use cases where a term is passed and may need a term wrapper.

### DIFF
--- a/client/mass/charts.js
+++ b/client/mass/charts.js
@@ -237,7 +237,7 @@ function setRenderers(self) {
 					// summary/survival/cuminc all expect config.term{} to be a termsetting object, but not term (which is confusing)
 					// thus convert term into a termwrapper (termsetting obj)
 					// tw.q{} is missing and will be fill in with default settings
-					const tw = { id: term.id, term }
+					const tw = term.q ? term : { id: term.id, term }
 					action.config[chart.usecase.detail] = tw
 					self.dom.tip.hide()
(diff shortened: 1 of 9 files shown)


client/mass/charts.js
```diff
@@ -237,7 +237,7 @@ function setRenderers(self) {
 					// summary/survival/cuminc all expect config.term{} to be a termsetting object, but not term (which is confusing)
 					// thus convert term into a termwrapper (termsetting obj)
 					// tw.q{} is missing and will be fill in with default settings
-					const tw = { id: term.id, term }
+					const tw = term.q ? term : { id: term.id, term }
```
siosonel, line 240: @airenzp it may be better to check for `term.term ? term : ...` instead of `term.q ? term : ...`, since a termwrapper is guaranteed to have {term}, but `q` may be filled-in only later by `fillTermWrapper()` or `fillTW()`.

xzhou82 (approved):
tested to work, thanks. feel free to merge

clicking a custom term from Dictionary app tree does break

PR #139, opened 2023-04-27
Pending fixes in the groups

airenzp:
Added some pending fixes to the PR from yesterday. Showing sample names was disabled always, fixed. Improved groups GUI as discussed previously. Used .term to check if term wrapper, as @siosonel commented out. Please check.

### DIFF
--- a/client/mass/groups.js
+++ b/client/mass/groups.js
@@ -82,18 +82,23 @@ class MassGroups {
 
 	async groups2samplelst(groups) {
 		const samplelstGroups = []
+		let disabled = true
 		for (const g of groups) {
 			const samples = await this.app.vocabApi.getFilteredSampleCount(g.filter, 'list')
 			const items = []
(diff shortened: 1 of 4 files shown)


client/mass/groups.js
```diff
@@ -143,7 +148,9 @@ class MassGroups {
 		if (this.state.customTerms.length == 0) {
 			this.dom.customTermDiv
 				.append('div')
-				.text('No custom variables. Use above controls to create new ones.')
+				.text(
+					'No custom variables. Use above controls to create new ones. Custom variables are added to dictionary terms'
```
xzhou82, line 152: please replace as `are added to dictionary.`
add missing period at the end
just say dictionary as that's where they appear together with official terms
also on the ui let's print "variable" but not "term"

xzhou82 (approved):
thanks

PR #143, opened 2023-04-27
Run rust build after npm_ci.

aacic:


siosonel:
may be good to have the option to not display stats in mass plots, so that integrations tests will not trigger R, Rust code for tests that do are directly testing the stats table rendering or results

siosonel (approved):

PR #145, opened 2023-05-01
Filter group

siosonel:
Full test: http://localhost:3000/testrun.html?dir=filter

Manual tests: 
1. Should now show `+AND`, previously this was hidden since it matched the group's join value. Click on `+AND` to add filters with the same join value as the negated subgroup
Screenshot 2023-05-01 at 10 13 09 AM

2. Should now have a menu option to easily switch the group's join value between 'AND' and 'OR', useful when negating a group which will require reversing the group's disjunction or conjunction
Screenshot 2023-05-01 at 10 15 32 AM


congyu-lu (approved):
works great!

xzhou82 (approved):

PR #151, opened 2023-05-03
Added logic to show plots on top of other plots as they are opened. 

airenzp:
As suggested by @siosonel, added logic to show a new plot on top of other plots as they are opened. Also did some refactoring moving the menu creation to the groups. Edgar wanted that the new plots could be visible to the user.


siosonel:
@airenzp It works when opening survival and matrix, the plot shows directly above the scatter plot that launched the new plots. But when I click `Summarize`, the plot still opens all the way to the very top.

PR #153, opened 2023-05-03
Groups cuminc

airenzp:
Added support for cuminc and other fixes, like the ordering of the plots. Did some refactoring related and test fixes. Please test if works as expected in the scatter and in the groups. Edgar helped me to solve some issues in this PR (eg using supportedChartTypes instead of allowedTermTypes)

### DIFF
--- a/client/mass/groups.js
+++ b/client/mass/groups.js
@@ -4,13 +4,7 @@ import { filterInit, getNormalRoot, filterPromptInit, getFilterItemByTag } from 
 import { select } from 'd3-selection'
 import { appInit } from '#termdb/app'
 import { renderTable } from '#dom/table'
-import {
-	getSamplelstTW,
-	addMatrixMenuItems,
-	openSummaryPlot,
-	addPlotMenuItem,
-	showTermsTree
-} from '#termsetting/handlers/samplelst'
+import { getSamplelstTW } from '#termsetting/handlers/samplelst'
 import { get$id } from '../termsetting/termsetting'
 import { getActiveCohortStr } from './charts'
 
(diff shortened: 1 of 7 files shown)


client/plots/sampleScatter.interactivity.js
```diff
@@ -459,31 +458,13 @@ export function setInteractivity(self) {
...
 			})
-		addMatrixMenuItems(self.dom.tip, menuDiv, tw, self.app, self.state)
-		if (self.state.allowedTermTypes.includes('survival')) {
-			const survivalDiv = menuDiv
-				.append('div')
-				.attr('class', 'sja_menuoption sja_sharp_border')
-				.style('position', 'relative')
-				.html('Survival analysis&nbsp;&nbsp;&nbsp;›')
+		addMatrixMenuItems(self.dom.tip, menuDiv, tw, self.app, self.id, self.state)
+		if (this.state.supportedChartTypes.includes('survival'))
```
siosonel, line 462: I think this should be `self.state...`, not `this.state...`

client/plots/sampleScatter.interactivity.js
```diff
@@ -459,31 +458,13 @@ export function setInteractivity(self) {
...
+			addPlotMenuItem('survival', menuDiv, 'Compare survival', self.dom.tip, tw, self.id, this)
 
-			survivalDiv.on('click', async e => {
-				const state = {
-					nav: { header_mode: 'hide_search' },
-					tree: { usecase: { target: 'survival', detail: 'term' } }
-				}
+		if (this.state.supportedChartTypes.includes('cuminc'))
```
siosonel, line 465: also `self.state...`, not `this.state....`

gavrielm:
I added a check to `cuminc.js` that makes sure that the cuminc term is a condition term. This will catch the error before R script gets triggered.

siosonel:
@airen: can be done in a separate branch: these callbacks in the screenshot below, for the group menu options, do not belong in `handlers/samplelst.js`, maybe should be in `mass/groups.js` instead? But also, some of these functions assume that `.attr('class', 'sja_menuoption sja_sharp_border')` is going to be used. I think the code organization can be improved.

Screenshot 2023-05-03 at 2 49 47 PM

xzhou82:
I agree these launch plot functions should be in mass/groups, as they're
not termsetting logic

siosonel (dismissed):

airenzp:
Moved the functions now. I dont know what you mean with this "But also, some of these functions assume that .attr('class', 'sja_menuoption sja_sharp_border') is going to be used" but we can talk about it later.

client/mass/groups.js
```diff
@@ -493,7 +493,7 @@ export async function openSummaryPlot(term, samplelstTW, app, id, newId) {
 export async function showTermsTree(div, callback, app, parentMenu, state = { tree: { usecase: { detail: 'term' } } }) {
 	const menu = new Menu({ padding: '5px', offsetX: 170, offsetY: -34 })
 	menu.showunderoffset(div.node())
-	const termdb = await import('../../termdb/app')
+	const termdb = await import('../termdb/app')
```
siosonel, line 496: @airenzp there is already an `import { appInit } from '#termdb/app'` at the beginning of `groups.js`, no need for dynamic import of termdb/app here

airenzp: Fixed now

siosonel (approved):

PR #154, opened 2023-05-03
Refactor to use function to add common items

airenzp:
In the group menus some refactoring could be done to remove duplicated code. Requested by @siosonel .


xzhou82 (approved):
works, thanks

PR #156, opened 2023-05-04
Fixed bug using a custom variable for the shape

airenzp:
Fixed bug we found when selecting a custom variable for the shape.


xzhou82 (approved):
thanks

PR #159, opened 2023-05-04
Added scatter groups to app groups

airenzp:
Now the scatter groups are added and removed to/from the groups. Eventually I assume we want to replace the plot groups with these groups, but that will require more changes. So I leave it for another PR


xzhou82 (approved):
thanks

PR #162, opened 2023-05-05
Fixed bugs creating groups

airenzp:
I tested a bit more the group changes from yesterday. When creating a group the filter was containing all the samples, not just the ones from the group, as it was using a samplelst created with all the groups. Fixed now. Also creating a group from the list samples was failing. Fixed now.


xzhou82 (approved):
thanks

PR #164, opened 2023-05-05
fix the ordering of condition term values in the barchart

siosonel:
Test with condition and numeric terms. The fix had do with the backend not querying the db for `termid`, if available.
@gavrielm the fix is mostly done in the backend, no need to add the condition term when setting excludes
@xzhou82 no need to add `order` in the grade entries for condition term.values


xzhou82 (approved):
thanks for the fix

PR #165, opened 2023-05-08
Scatter divide by

airenzp:
Supporting divide by from the scatter plot. Updated summary plot to pass divideBy term. Updated server side to return multiple charts. Supported all term types: numeric, categoric or geneVariant.

### DIFF
--- a/client/plots/sampleScatter.interactivity.js
+++ b/client/plots/sampleScatter.interactivity.js
@@ -112,7 +112,6 @@ export function setInteractivity(self) {
 		const hidden = tw.q.hiddenValues ? key in tw.q.hiddenValues : false
 		const menu = new Menu({ padding: '5px' })
 		const div = menu.d.append('div')
-
 		div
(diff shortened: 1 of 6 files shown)


airenzp:
I found that the hiding effect I was showing you on the meeting(Clicking on a category hides all the samples with that category) is because I mark the samples with that category as hidden. So is not a bug. And since you thought it made sense I leave it as it is.

airenzp:
The summary scatter also seems to be working as expected, when using mbmeta, that has the metastasis stage as numeric.(PNET does not and that is why I did not see the scatter tab I think)

airenzp:
I need to make the tool icons like zoom or search to work on all the charts. I am marking the PR as draft again.

airenzp:
Added some more fixes. Now the zoom icons work, the category is displayed and there are some styling improvements.

xzhou82 (approved):

PR #172, opened 2023-05-10
Scatter regression

airenzp:
This is a draft PR to plot regression curves on the scatter plot. To plot the curve you need to select a regression from the burger menu.

NOTE: From this repo or the parent repo (where proteinpaint is a submodule), run `npm install d3-regression --workspace=[proteinpaint/]client`, see the [related Monorepo instructions](https://docs.google.com/document/d/1UX80FqtNb3eRhj5mfHLwZLGFmktyHwNjrmC3EI6IOYE/edit#heading=h.w1rk41cxkbbl)

### DIFF
--- a/client/termdb/TermdbVocab.js
+++ b/client/termdb/TermdbVocab.js
@@ -868,6 +868,23 @@ export class TermdbVocab extends Vocab {
 		return await dofetch3('termdb', { headers, body })
 	}
 
+	async getLowessCurve(opts) {
+		// the scatter plot may still render when not in session,
+		// but not have an option to list samples
+		const headers = this.mayGetAuthHeaders()
+
+		// dofetch* mayAdjustRequest() will automatically
+		// convert to GET query params or POST body, as needed
+		const body = {
+			genome: this.state.vocab.genome,
+			dslabel: this.state.vocab.dslabel,
+			getLowessCurve: 1,
+			coords: opts.coords
+		}
+
+		return await dofetch3('termdb', { headers, body })
+	}
+
 	async getCohortsData(opts) {
 		const body = {
 			getCohortsData: 1,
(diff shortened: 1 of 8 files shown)


client/termdb/TermdbVocab.js
```diff
@@ -868,6 +868,23 @@ export class TermdbVocab extends Vocab {
 		return await dofetch3('termdb', { headers, body })
 	}
 
+	async getLowessCurve(opts) {
```
siosonel, line 871: @airenzp may consider doing this on the client-side, for example using https://github.com/HarryStevens/d3-regression? Since the data points are already available from the server response, it seems that a client-side stat library may be more appropriate.

xzhou82:
good idea

airenzp:
Fixed rendering and passed two arrays to R. Will try now the d3-library.

airenzp:
The new version using d3-regression now works, and is much faster. There is no lowess implementation so I used a loess regression  instead, which seems quite similar. In order to test it the d3-package needs to be installed. This package is added in a PR from sjpp.

airenzp:
I thought we could allow the user to select the type of regression instead. Supporting now two types of regression, loess and polynomial, as they render in a similar way.

siosonel:
> The new version using d3-regression now works, and is much faster. There is no lowess implementation so I used a loess regression instead, which seems quite similar. In order to test it the d3-package needs to be installed. This package is added in a PR from sjpp.

@airenzp I updated this PR's description on how to install, the package should not be part of the sjpp repo.

airenzp:
Added the lowess regression from R and is actually very performant

xzhou82 (approved):

xzhou82 (approved):
works great!

PR #173, opened 2023-05-10
group all uncomputable values under one violin legend group

siosonel:
also fix and improve the integration tests

### DIFF
--- a/client/mass/test/violin.integration.spec.js
+++ b/client/mass/test/violin.integration.spec.js
@@ -347,7 +347,7 @@ tape('test label clicking, filtering and hovering', function(test) {
 
 		await labelClicking(violin, violinDiv) //test filter on label clicking
 		await testFiltering(violin, violinSettings, violinDivData) //test filtering by providing tvs.lst object
-		testLabelHovering(violin, violinDiv)
+		await testLabelHovering(violin, violinDiv)
 
 		if (test._ok) violin.Inner.app.destroy()
 		test.end()
@@ -423,7 +423,7 @@ tape('test label clicking, filtering and hovering', function(test) {
 		const filterUiRoot = getFilterItemByTag(violin.Inner.state.termfilter.filter, 'filterUiRoot')
 		const filter = filterJoin([filterUiRoot, tvslst])
 		filter.tag = 'filterUiRoot'
-		violin.Inner.app.dispatch({
+		await violin.Inner.app.dispatch({
 			type: 'filter_replace',
 			filter
 		})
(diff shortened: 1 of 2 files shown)


client/mass/test/violin.integration.spec.js
```diff
@@ -347,7 +347,7 @@ tape('test label clicking, filtering and hovering', function(test) {
 
 		await labelClicking(violin, violinDiv) //test filter on label clicking
 		await testFiltering(violin, violinSettings, violinDivData) //test filtering by providing tvs.lst object
-		testLabelHovering(violin, violinDiv)
+		await testLabelHovering(violin, violinDiv)
```
siosonel, line 350: @karishma-gangwani not using await was causing error messages to be displayed

client/mass/test/violin.integration.spec.js
```diff
@@ -673,24 +673,30 @@ tape('test uncomputable categories legend', function(test) {
...
 
 		await testUncomputableCategories(violin, legendDiv)
-
+		if (test._ok) violin.Inner.app.destroy()
 		test.end()
 	}
 
 	async function testUncomputableCategories(violin, legendDiv) {
-		const categories = await detectGte({ elem: legendDiv.node(), selector: '.legend-row', count: 9 })
 		const keys = Object.keys(violin.Inner.data.uncomputableValueObj)
```
siosonel, line 681: Unless the order of legend entries is guaranteed, it is best to not hardcode the index position in which the legend entry may appear in the array as returned by `detectGte` array.

creilly8 (approved):

client/plots/violin.renderer.js
```diff
@@ -429,7 +429,6 @@ function getLegendGrps(self) {
 		t1 = self.config.term,
 		t2 = self.config.term2,
 		headingStyle = 'color: #aaa; font-weight: 400'
-	let title, name
```
siosonel, line 432: @karishma-gangwani Please avoid sharing variables between functions or loops, unless those variables are really meant to be shared state. In this case, the title/name is unique to a legend entry, there is no reason to share that variable value.

client/plots/violin.renderer.js
```diff
@@ -450,18 +449,19 @@ function getLegendGrps(self) {
...
 		if (term?.term.values) {
+			const items = []
 			for (const k in term.term.values) {
 				if (self.data.uncomputableValueObj?.[term.term.values[k]?.label]) {
-					const items = []
 					items.push({
 						text: `${term.term.values[k].label}, n = ${self.data.uncomputableValueObj[term.term.values[k].label]}`,
 						noIcon: true
 					})
-					title = `${term.term.name}`
```
siosonel, line 460: @karishma-gangwani This code was creating one legend group for each uncomputable value. I moved it outside the loop so that all of the uncomputable values will be under one group.

client/plots/violin.renderer.js
```diff
@@ -478,8 +478,7 @@ function getLegendGrps(self) {
 					hiddenOpacity: 1
 				})
 			}
```
siosonel, line 480: See my comment above on not sharing variable across functions and loops, unless necessary.

client/mass/test/violin.integration.spec.js
```diff
@@ -673,24 +673,30 @@ tape('test uncomputable categories legend', function(test) {
...
-		const category1 = categories[8].innerText.split(',')[0] + ',' + categories[8].innerText.split(',')[1]
-		const category2 = categories[9].innerText.split(',')[0]
+		const categories = await detectGte({ elem: legendDiv.node(), selector: '.legend-row', count: 9 })
+		const uncomputableLegend = categories.filter(c => keys.find(k => c.__data__.text.startsWith(k)))
```
creilly8, line 683: @siosonel What about iterating through uncomputableLegend to make the "Uncomputable category..." tests? Then it's only one `test.equal` to write. What do you think?

siosonel: @creilly8 Yes, the tests can be optimized that way

PR #175, opened 2023-05-11
use the runpproteinpaint function, as exported from the front helper …

siosonel:
…code, in mds3 test. Will manually run integration test on this branch


creilly8 (approved):
Thanks for the fix

PR #178, opened 2023-05-12
download the barchart legend as part of the svg image

siosonel:
Per Sandeep's request


congyu-lu:
gender overlay with age group, in the downloaded figure the legend is moved to the right and also this error: 
Screen Shot 2023-05-12 at 8 55 34 AM

divide by any term in the downloaded figure the legend is not there.

xzhou82 (approved):
thanks it works. it does report an error when downloading
<svg> attribute height: Expected length, "667.592.5".

xzhou82:
thanks for finding it. sorry merged too fast

PR #179, opened 2023-05-12
Violin legend fixes

karishma-gangwani:
this should fix the pvaluetable rendering issues, hide uncomputable values from rendering, hide selected plot and values from pvaluetable and display an intuitive message saying 'click to unhide plot' in the legend. Please let me know if you see any other issues regarding these.


gavrielm (approved):
Works great! Thanks!

xzhou82 (approved):
thanks!

PR #180, opened 2023-05-12
download function to fix survival plot download errors

congyu-lu:


client/common/svg.download.js
```diff
@@ -0,0 +1,84 @@
...
+		for (const prop of titles[i].styles) {
+			if (prop.startsWith('font')) title.style(prop, titles[i].styles.getPropertyValue(prop))
+		}
+
+		select(mainG).attr('transform', 'translate(' + corner.x + ',' + (corner.y + tboxh) + ')')
+		svg.appendChild(mainG)
+	})
+
+	//const svg_name = self.config.term.term.name + ' barchart'
+	const svg_name = ' barchart'
```
siosonel, line 82: The (optional) filename should be an argument, or default to self.chartType or a non-chart-specific generic name
```js
export function downloadChart(self, div, mainG, svgStyle, filename='') {
    ...
   const svg_name = filename || self.chartType || 'download'
   ...
}
```
It's confusing to see "barchart" in the filename for a downloaded survival plot.

congyu-lu: Sorry, I just changed it

siosonel (approved):

PR #181, opened 2023-05-15
Hiding terms menu when needed. Cleanup

airenzp:
When showing the terms tree to open a plot from the groups the menu was not hidden if clicking out or changing to another option. Fixed now. Also done some cleanup related.


xzhou82 (approved):
thanks

PR #186, opened 2023-05-17
Geneset search

airenzp:
This draft PR adds a function to render the geneset search discussed and creates a test to use it. It can be tested from http://localhost:3000/testrun.html?name=genesetEdit. I am opening it to share what I have so far and see if this is what we meant. Please check. I will continue to work on it until we think is ready.


client/dom/genesetSearch.js
```diff
@@ -0,0 +1,61 @@
+export function initGenesetSearch({ holder, genome, callback, geneList, mode = 'mutation', vocabApi }) {
+	const div = holder.append('div').style('width', '50vw')
+	div
+		.style('border-style', 'solid')
+		.style('border-width', '2px')
+		.style('border-color', '#eee')
+		.style('padding', '5px')
+	const headerDiv = div.append('div')
+	const inputSearch = headerDiv.append('input').attr('placeholder', 'Search genes')
```
siosonel, line 9: @airenzp this should use `import { addGeneSearchbox } from '#dom/genesearch'`

client/dom/test/genesetSearch.spec.js
```diff
@@ -0,0 +1,127 @@
...
+tape('\n', function(test) {
+	test.pass('-***- dom/geneDictionary -***-')
+	test.end()
+})
+
+tape('Search genes test', function(test) {
+	test.timeoutAfter(3000)
+
+	runTests()
+
```
siosonel, line 99: does not need to use a `runTests()` wrapper function

siosonel:
@airenzp I think it's a good start, just need to more closely match the button, input, layout as shown in the drawing.

xzhou82:
let's name it `geneSetEdit` as the purpose is to create or edit a gene set

airenzp:
Thanks for the comments. I will continue to work on it

PR #192, opened 2023-05-23
Enable typescript

creilly8:


siosonel (changes requested):
@creilly8 I think we should enable typescript only for server side for now, so no need to have client-side code changes in this PR branch? Not sure how close @aacic is with the disco plot changes, but I would assume he would activate typescript for client-side when his branch is ready.

Also, I think there also has to be a typescript devDependency in server/package.json.

Lastly, I'd prefer not committing the empty line additions in server/src/app.js.

creilly8:
Thanks @siosonel. Those changes are made.

siosonel (dismissed):
Thanks, works

siosonel (approved):

PR #193, opened 2023-05-23
Geneset edit

airenzp:
Draft to edit genesets from the matrix

### DIFF
--- a/client/dom/genesearch.js
+++ b/client/dom/genesearch.js
@@ -60,6 +60,8 @@ function argument object {}
 	triggered when a valid hit is found, and has been written to RESULT object (see below)
 	no parameter is supplied
 
+.hideHelp: true
+	if true, hide the text msg on the right of <input>
 
 ***********************************
 by calling addGeneSearchbox(), it redirectly returns a RESULT object detailed below without await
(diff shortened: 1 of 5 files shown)


airenzp:
I am marking this PR as ready for review as I think it addresses most of the things we discussed. Note that for deleting genes now one needs to select the genes and then click the delete button on top. I felt this way was better than adding a close button to all the genes. Please check.

siosonel:
@airenzp There should be more margins between the header, middle, and footer divs in the geneset edit menu, and also a border in the middle div. Also set a max-width of maybe 500 px so that the menu does not get too wide. It should really match the drawing more closely. 

Also, I'm not sure about using a `Delete` button at the top, clicking on the displayed gene with an 'x' should already remove that gene. I can try to commit these ideas in a branch of top of yours, or you can work on these in your current branch, let me know.

airenzp:
Thanks for the comments. If it wont take you too long I think it is better for you to adjust the styling and the buttons to match the scheme.  Also I dont know how to put an x on top of the button and probably for you is easier. If something takes too long please let me know and I can work on that.

airenzp:
Also I used a bigger size because when displaying the other buttons and choices, that are now commented out until supported, more space was needed. So keep in mind those buttons will be there later.

xzhou82 (approved):
thanks, will merge now due to big changes from ts work

PR #203, opened 2023-06-01
New dataset types

creilly8:
- New `server/shared/types` dir
- New `../genome.ts`, `dataset.ts`, and `index.ts`
- All `pp/server/dataset/*` files typed
- Minor update to `server/app.js`

Must checkout transpileDatasets in sjpp repo to transpile files appropriately.


siosonel:
> @siosonel For the unit tests to pass, it looks like the transpiled files must be included in the PR. Ok to put them back?
> 
> Maybe not. This might be failing because the code for transpiling isn't in sjpp master.

@creilly8 This should not be related to sjpp.

@aacic can you take a look? I think last week you did made sure that genome ts files are compiled before running CI tests? If true, we'd need something similar for `dataset/*.ts` files

siosonel (approved):

PR #205, opened 2023-06-02
Scatter 3d

airenzp:
In this draft PR I enable the rendering of a 3D plot if the divideBy term is continuous. As I had already started working on this I felt I should finish the idea. Please check. I could not find a good use case, where the rendering of a 3rd axis makes a big difference though. I also fixes some small bugs found in between.


xzhou82 (dismissed):

xzhou82 (approved):

client/plots/sampleScatter.js
```diff
@@ -314,7 +328,7 @@ export async function getPlotConfig(opts, app) {
 		}
 		// may apply term-specific changes to the default object
 		const result = copyMerge(config, opts)
-		return result
+		if (result.term0?.q?.mode == 'continuous' && !app.hasWebGL()) return result
```
xzhou82, line 331: @airenzp it's returning undefined for premade plots

siosonel: @airenzp this has to throw here, won't be able to recover a save session with 3d plot when the browser does not support WebGL

airenzp: Yes, added the throw now, thanks for the heads up.

PR #212, opened 2023-06-07
Vp user controlled width height

karishma-gangwani:
This will allow the user to modify the width and height of the plots.

### DIFF
--- a/client/mass/test/violin.integration.spec.js
+++ b/client/mass/test/violin.integration.spec.js
@@ -265,7 +265,7 @@ tape('term1 as numeric and term2 categorical, test median rendering', function(t
 })
 
 tape('test basic controls', function(test) {
-	test.timeoutAfter(3000)
+	test.timeoutAfter(4000)
 	runpp({
 		state: {
 			nav: {
--- a/client/plots/controls.config.js
+++ b/client/plots/controls.config.js
@@ -169,7 +169,9 @@ function setNumberInput(opts) {
 		]
 
 	// debounce by default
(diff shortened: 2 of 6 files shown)


karishma-gangwani:
> 

so remove min and max thickness?

siosonel:
@xzhou82 may be able to combine the `q.definedPlotThickness` into the `q.plotThickness` input. If left blank, then auto-compute, otherwise use it's value.

@karishma-gangwani I suggest leaving the min, maxThickness in the configuration, but do not expose it as an input. You can put back in the plotThickess as a separate config option and display as a control menu input.

siosonel:
@karishma-gangwani actually, maybe remove the minThickness, should be at least 3x the radius. The maxThickness could remain but be unexposed to the user, later on it can be exposed if requested.

karishma-gangwani:
> @karishma-gangwani actually, maybe remove the minThickness, should be at least 3x the radius. The maxThickness could remain but be unexposed to the user, later on it can be exposed if requested.

okay. let me see

xzhou82:
i'm not sure if it's good idea to have the `Max thickness` control. [here](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22hrtavg%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22diaggrp%22}}],%22nav%22:{%22activeTab%22:1}}) changing value 150 to 100 has no effect, it's counter intuitive

there should just be an input showing the current thickness (even if automatically decided), and allow me to adjust it

also CI breaks

xzhou82 (approved):

PR #215, opened 2023-06-08
Fixed skipped cuminc tests

gavrielm:
Fixed the skipped cuminc tests in `client/mass/test/cuminc.integration.spec.js`. These tests previously failed because I used the incorrect input data to generate the `precomputed_cuminc` table in TermdbTest db. Using the correct input data allows the tests to pass. Please test at: http://localhost:3000/testrun.html

Made a minor change in `client/plots/cuminc.js` to not require the presence of term2 when looking for skipped series (possible that the term1 series could be skipped)


siosonel (approved):
works

PR #217, opened 2023-06-08
matrix fixes

siosonel:
- create a matrix tw.settings.barh as needed to support change to continuous mode
- also show the dt, origin label…s on mouseover to disambiguate the seemingly conflicting not tested, wildtype values on mouseover

Screenshot 2023-06-08 at 5 17 41 PM


xzhou82 (approved):
thanks
 `tw.settings.barh` should be customizable on ui

PR #219, opened 2023-06-09
fix barChart ranking error and add to commons.js morigin germline and…

congyu-lu:
… somatic

Can remove morigin.germline and morigin.somatic if we revert back to 'G'/'S' in neuro onc datasets.


siosonel (approved):

xzhou82:
how to replicate the ranking err?

xzhou82 (approved):

PR #220, opened 2023-06-09
Groups cleanup

airenzp:
Alternative groups storage and logic considering if the mass UI is visible or not. Added validation when renaming groups to ensure names are unique.

### DIFF
--- a/client/mass/groups.js
+++ b/client/mass/groups.js
@@ -318,11 +318,8 @@ async function updateUI(self) {
 			{
 				text: 'Delete',
 				callback: (e, i) => {
-					groups.splice(i, 1)
-					self.app.dispatch({
-						type: 'app_refresh',
-						state: { groups }
-					})
+					const group = groups[i]
+					self.app.vocabApi.deleteGroup(group.name)
 				}
 			}
 		],
--- a/client/mass/groups.js
+++ b/client/mass/groups.js
@@ -302,13 +302,18 @@ async function updateUI(self) {
 		columns: [
 			{
 				label: 'NAME',
-				editCallback: (i, cell) => {
-					// group name is changed
-					groups[i].name = cell.value
-					self.app.dispatch({
-						type: 'app_refresh',
-						state: { groups }
-					})
+				editCallback: async (i, cell) => {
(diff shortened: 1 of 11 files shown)


client/termdb/Vocab.js
```diff
@@ -173,17 +173,32 @@ export class Vocab {
...
-		})
+		const i = this.state.groups.findIndex(i => i.name == name)
+		if (i != -1) {
+			const group = this.state.groups[i]
+			this.state.groups.splice(i, 1)
+			await this.app.dispatch({
+				type: 'app_refresh',
+				state: { groups: this.state.groups }
+			})
+			const config = this.state.plots.find(p => p.id === group.plotId)
```
xzhou82, line 193: thanks looks ok to me. @siosonel what do you think about dispatching twice in a row

siosonel: I don't think two dispatches are needed. 
- When changing a tw group for a chart's config, then dispatch only `plot_edit`
- When renaming or editing a group, that could affect other charts, then should have a `group_edit` action in the store (would require updating `reactsTo()` of each chart class), OR use app_refresh (wouldn't require updating reactTo()).

airenzp: Solved now. Edgar helped me to review the reacting logic.

client/mass/groups.js
```diff
@@ -302,13 +302,18 @@ async function updateUI(self) {
...
-				editCallback: (i, cell) => {
-					// group name is changed
-					groups[i].name = cell.value
-					self.app.dispatch({
-						type: 'app_refresh',
-						state: { groups }
-					})
+				editCallback: async (i, cell) => {
+					const newName = cell.value
+					const index = self.state.groups.findIndex(group => group.name == newName)
```
siosonel, line 307: I think this check should be done in the `mass/store.js`, in case other places allow group name edits, then no need to duplicate this validation check

xzhou82 (approved):
thanks a lot it works

client/mass/groups.js
xzhou82: can look at it later

PR #226, opened 2023-06-14
Fixed bug when rendering multiple charts with regression

airenzp:


karishma-gangwani:
Screenshot 2023-06-15 at 8 44 57 AM

thanks for your fixes Airen. It looks like divideBy is not working. and label shows as 'default'

karishma-gangwani (approved):
works! thanks Airen

airenzp:
@karishma-gangwani could not see the last changes in this PR until she stopped the server and restarted. Doing rs did not worked. I leave a note in case it happens again.

PR #227, opened 2023-06-14
Migrate termsetting to typescript

creilly8:
1. Enable typescript on client 
2. Migrated `client/termsetting` to ts
3. New termdb, termsetting, filter, and vocab types

All [unit ](http://localhost:3000/testrun.html?name=*.unit)and [integration](http://localhost:3000/testrun.html?name=*.integration) tests in the browser worked for me. The termsetting unit and integration tests worked for me as well. 

Test with:
- termsetting.integration: http://localhost:3000/testrun.html?name=termsetting.integration
- termsetting.unit: http://localhost:3000/testrun.html?name=termsetting.unit


siosonel:
@creilly8 works for me so far, after fixing a few very minor issues. 

It may be good to separate out termsetting, filter types that are only used in the client side into `client/types/[ termsetting | filter].ts`. So the types that are in `shared/types` are really those that are used by both server and client code.

I also wonder why `shared/types/*.js` files are being generated, I would have assumed those are not needed at all with webpack doing the bundling + transpilation. Not a big deal, but p

creilly8:
Sure @siosonel, I can separate out those types. Lets do that in another PR. 

I suspect `tsc watch` reads wild cards differently. I remember reading something about that in typescript documentation. I'll try to find that again.  

I updated the import statements to match the paths map then reran the unit and integration tests. Everything passed for me.

siosonel (approved):

aacic (approved):

PR #232, opened 2023-06-16
New disco plot type script

aacic:


xzhou82 (approved):

client/plots/disco_new/mapper/CnvArcsMapper.ts
```diff
@@ -0,0 +1,190 @@
...
+        return chromosome.startAngle + ((chromosome.endAngle - chromosome.startAngle) * (Number(data.stop) / chromosome.size));
+    }
+
+    getColor(value) {
+
+        const lossColor = this.settings.cnv.lossColor
+        const ampColor = this.settings.cnv.ampColor
+        const cappedLossColor = this.settings.cnv.cappedLossColor
+        const cappedAmpColor = this.settings.cnv.cappedAmpColor
+
```
siosonel, line 108: @aacic The variable reassignments above are not needed, since they are only used once within this function, so there is no readability gained by having this extra layer of indirection. For example, just return the settings directly
```js
         if (value < this.lossCapped) {
            return this.settings.cnv.cappedLossColor
        } ...
```

If a setting is used in more than one place within the same function, then it may be simpler to assign the variable alias to the parent config, like this
```js
        const s = this.settings.cnv
        
        if (value < this.lossCapped) {
            return s.cappedLossColor
        } ...
``

File: client/plots/disco_new/mapper/CnvArcsMapper.ts [hunk 2]
```diff
@@ -0,0 +1,190 @@
...
+
+        if (value < this.lossCapped) {
+            return cappedLossColor
+        } else if (value <= 0) {
+            return lossColor
+        } else if (value <= this.gainCapped) {
```
siosonel, line 114: @aacic The condition here seems incorrect: is this supposed to be `else if (value >= this.gainCapped)`? Otherwise, if it's really less than, then it would seem random that a value could be matched against `< this.lossCapped` first. I could be wrong, without having read the input value as being either negative or positive, but it seems that the mutation class should be considered here.

client/plots/disco_new/mapper/CnvArcsMapper.ts
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    private calculateInnerRadius(data: Data) {
+        if (this.gainOnly) {
+            return this.settings.rings.cnvInnerRadius
+        }
+
+        if (this.lossOnly) {
+            let outerRadius = this.settings.rings.cnvInnerRadius + this.settings.rings.cnvWidth
```
siosonel, line 127: @aacic since the value didn't change within the variable scope, should have declared as `const` instead of `let`

client/plots/disco_new/mapper/CnvArcsMapper.ts
```diff
@@ -0,0 +1,190 @@
...
+        } else if (value <= 0) {
+            return lossColor
+        } else if (value <= this.gainCapped) {
+            return ampColor
+        } else if (value > this.gainCapped  ) {
+            return cappedAmpColor
+        }
+    }
+
+    private calculateInnerRadius(data: Data) {
```
siosonel, line 121: @aacic for readability/conciseness, may want to use `const s = this.settings.rings

siosonel:
@aacic In general, I find the mapper code/indirection very hard to follow. For example

```js
// CnvArcsMapper.ts, line 41
const gain = new CnvLegend("Max", (cnvMaxValue > 0)?CnvType.Gain: CnvType.Loss, this.getColor(cnvMaxValue), cnvMaxValue)

						// CnvLegend.ts, line 9
						constructor(text: string, cnvType: CnvType, color: string, value: number) {
						  this.text = text
						  this.cnvType = cnvType;
						  this.color = color;
						  this.value = value;
						}

						// why not use an object literal??? no need for a constructor
						const gain = {
							text, 
							cnvType: 'Gain', // should use the common.js dt/mclass encoding here?? 
							color, 
							value // cnvMaxValue
						}


// CnvArcsMapper.ts, line 45
this.cnvClassMap.set(CnvType.Gain, gain)

					 // has a Typescript Type as key,
					 // and the gain legend as value, 
					 // based on cnvClassMap: Map<CnvType, CnvLegend> = new Map() // CnvArcsMapper.ts, line 10

					 // why not use 
					 {Gain: gain}
					 // or even simpler,
					 {gain}

// Then as accessed/used in LegendJSONMapper.ts line 64
private mapCnv(legend: Legend, legendJSON: Array<any>, order: number) {
        const gain = legend.cnvClassMap!.get(CnvType.Gain)
        // basically accesses the contructed object from line 9 above, 
        // which could have been a reference to an object literal instead

         if (gain && loss && cap) {
            let cnvOrder = 0
            const cnvItems: Array<any> = []
            if (gain.value > 0) {
                cnvItems.push(
                    {
                        termid: legend.cnvTitle,    // "termid" can be misleading here, usually this is used within the context of termsetting terms 
                        key: CnvType.Gain,          // should use the common.js dt/mclass encoding here?? 
                        text: `Max: ${gain.value}`,
                        color: gain.color,
                        order: cnvOrder++,
                        border: "1px solid #ccc"
                    }

                    // that's a lot of backtracking, following indirection to find out what gain means for the legend entry here
                )
            }
```

client/plots/disco_new/mapper/CnvArcsMapper.ts
aacic: This was fixed by the PR:
https://github.com/stjude/proteinpaint/pull/234
--fix option for eslint converts automatically let to const.

client/plots/disco_new/mapper/CnvArcsMapper.ts
aacic: This is fixed by PR:
https://github.com/stjude/proteinpaint/pull/235

aacic:
I'll come up with a nicer way to deal with this.

PR #233, opened 2023-06-16
Add .eslintrc.js

aacic:


siosonel:
@aacic I think we should turn off some rules that add a lot of warning/noise to the linter output. This is until we are able to convert most of the `any` type to a better defined type.
```js
"@typescript-eslint/no-explicit-any": "off"
```

server/genome/hg38.ts
```diff
@@ -1,4 +1,4 @@
-const clinvar = require('../dataset/clinvar')
+import { clinsig }  from '../dataset/clinvar'
```
siosonel, line 1: @aacic Thanks very much for this! This is great that we are finally implementing syntax quality control. Hopefully we can add to the `utils/hooks/pre-commit`.

aacic: I've tried to add following lines to .pre-commit-config.yaml
```
- repo: https://github.com/typescript-eslint/typescript-eslint
  rev: v5.59.11
  hooks:
    - id: typescript-eslint
```
But, I get this error:
```
An error has occurred: InvalidManifestError: 
=====> /Users/aacic/.cache/pre-commit/repoxxrx39jw/.pre-commit-hooks.yaml is not a file
Check the log at /Users/aacic/.cache/pre-commit/pre-commit.log
```
It seems that pre-commit doesn't work well with typescript-eslint. I would say we try to set-up Husky in another PR.

siosonel:
@aacic I got something working in the `setup-eslint-prettier-v2` branch. I added a few things to `utils/hooks/pre-commit`, and installed prettier v^2.8.8 which supports typescript reformatting. It may require an installation, from either the pp repo or the parent repo which has the pp repo as a submodule

```bash
npm install @typescript-eslint/eslint-plugin@latest --save-dev
``` 

If the `setup-eslint-prettier-v2` looks good, we can merge to your branch.

aacic:
Merged with: https://github.com/stjude/proteinpaint/pull/234

PR #236, opened 2023-06-20
Matrix single sample

congyu-lu:


client/plots/matrix.interactivity.js
```diff
@@ -105,7 +95,7 @@ export function setInteractivity(self) {
 					.append('div')
 					.attr('class', 'sja_menuoption sja_sharp_border')
 					.text(k)
-					.on('click', (event) => {
+					.on('click', event => {
```
siosonel: I think `const pane = newpane(); ... /* use pane.body as holder */` should be used, with a close button instead of a menu.

siosonel: Screenshot 2023-06-21 at 12 34 56 AM

siosonel (approved):

PR #238, opened 2023-06-21
replace newpane with adhoc sandbox

congyu-lu:


xzhou82 (approved):
thanks next can allow clicking on sample label to show the plot options
also the prettier/linting is changing styles back and forth

PR #239, opened 2023-06-21
replace menu with adhoc sandbox

congyu-lu:


xzhou82:
thanks, please show sample name in sandbox header, as in line 308 of sampletable.js
for both scatter and matrix

congyu-lu:
> thanks, please show sample name in sandbox header, as in line 308 of sampletable.js for both scatter and matrix

Thanks, added sample names in sandbox header

client/plots/matrix.interactivity.js
```diff
@@ -116,6 +117,7 @@ export function setInteractivity(self) {
 					.text('Disco plot')
```
xzhou82, line 117: line 113 `if()` should not be nested under singleSampleGenomeQuantification

congyu-lu: Thanks, fixed for both matrix and scatterplot

client/plots/matrix.interactivity.js
```diff
@@ -84,38 +84,45 @@ export function setInteractivity(self) {
 	}
 
 	self.mouseclick = function(event) {
-		if (self.state.termdbConfig.queries?.singleSampleGenomeQuantification) {
+		if (
+			self.state.termdbConfig.queries?.singleSampleGenomeQuantification ||
+			self.state.termdbConfig.queries?.singleSampleMutation
+		) {
 			const sample = event.target.__data__
 			sample.sample_id = sample.row?.sampleName
```
xzhou82, line 92: this is not right

xzhou82 (approved):
thanks i now realize a matrix with lots of samples won't show sample names upon zooming in, thus preventing from clicking sample name. but this is a separate issue

PR #243, opened 2023-06-21
Fix fusion tool tip.

aacic:


xzhou82 (approved):
do you know the reason of the massive prettify/linting changes, it's impossible to see what's actually changed

PR #250, opened 2023-06-23
GDC matrix now shows sample names if zoom is big enough

airenzp:
GDC samples returned from the server contain a sample field, not a sampleName, now it is read this field in function sampleLabel if sampleName is not provided


xzhou82 (approved):
thanks! will look into server returned data

PR #256, opened 2023-06-23
Type error fixes in categorical, condition, and geneVariant handlers

creilly8:


gavrielm (approved):
Tested condition term and it works well

PR #258, opened 2023-06-26
reduce array value to a single value as a quick fix

xzhou82:
please also pull sjpp master
test at http://localhost:3000/?genome=hg38&gene=ENST00000345146&mds3=GDC, click on a recurrent mutation to see "Primary diagnosis" used as outer ring values


karishma-gangwani (approved):
works! thanks

PR #274, opened 2023-06-29
Fixed issues with geneset

airenzp:
Added check to avoid duplicating genes. Fixed bug when selecting the active group.


xzhou82 (approved):
thanks

PR #275, opened 2023-06-29
Geneset menu position

airenzp:
Showing geneset at the same position of the parent popup


xzhou82 (approved):
looks okay i hope the menu won't flash when clicking `Edit`

PR #288, opened 2023-07-06
Replace legacy `/static/` urls in app drawer files

creilly8:
- replaced the pecan legacy URL with `../ppdemo/..` URLs
- Checked the links within the cards, should be no issues


siosonel (approved):
app cards work

PR #299, opened 2023-07-13
Refactor label logic for disco plot

aacic:
* Enabled typescript unit tests
* Introduced object literals instead of the Label class and various other classes
* Refactored label logic so labels for fusion events are added and prioritized


siosonel:
@aacic 

The changes to use object literal is much clearer, thanks! The test also works, but I did have to commit a few changes to allow your soec filename to show up in testrun.html and for http://localhost:3000/testrun.html?name=LabelsMapper.unit to work.

I moved the `client/plots/test/disco/mapper/LabelMapperTest.unit.spec.ts`, to `client/plots/disco/mapper/test/LabelMapper.unit.spec.ts` for the following reasons:

a. it's generally better to keep the spec file near the code file that's it's testing
b. in the PP codebase, the current practice is to have a test folder under the folder that holds code that could be tested. 
c. The other common way I've seen, like in GDC, is having the spec file in the same folder as the code that is being tested, so right next to a similarly named fiile. The reason that we don't do (c) is the code dir could become cluttered with helper code/data files.

Moving forward (maybe in different branch/PR), I hope you can reorganize the disco code files under "subject" directories rather than MVC "functional" dirs. So for example, under `plots/disco` after my last commit we have the more scattered
```text
mapper/
  - LabelMapper.ts
  - test/
      - LabelMapper.spec.ts]
renderer/
  - LabelRenderer.ts
viewmodel/
  - Label*.ts
```
preferred grouping of code files by "subject"
```text
label/
  - label.mapper.ts
  - label.renderer.ts
  - label.viewmodel.ts
  - label.types.ts 
  - test/
      - label.mapper.unit.spec.js
      - ...
```

siosonel:
@aacic Moving forward (maybe in different branch/PR), I hope you can reorganize the disco code files under "subject" directories rather than MVC "functional" dirs. So for example, under `plots/disco` after my last commit we have the more scattered
```text
mapper/
  - LabelMapper.ts
  - test/
      - LabelMapper.spec.ts]
renderer/
  - LabelRenderer.ts
viewmodel/
  - Label*.ts
```
preferred grouping of code files by "subject"
```text
label/
  - label.mapper.ts
  - label.renderer.ts
  - label.viewmodel.ts
  - label.types.ts 
  - test/
      - label.mapper.unit.spec.js
      - ...
```

xzhou82 (approved):

PR #300, opened 2023-07-13
Seperated .q{} types

creilly8:
- New type files for .q{} as well as eventual term, termsetting, and other relevant, specific types (e.g. vocabApi, dom). 
- Applied to `handler/*.ts` files

All unit and integration tests pass for me.


siosonel:
@creilly8 at first glance, I would prefer even more separation. I can help work on that as needed, except I may not get to it by today or tomorrow.

1. There is really only one "type" of `TermsettingInstance`. One of ny examples from yesterday (copied below) is wrong, explained by #2 - the instance type does not actually change as its instance.term.type changes.
```js
type CategoricalTermsettingInstance extends TermsettingInstance`
```
2. The `TermsettingInstance.handlerByType` object already disambiguates the different handler types.
3. The only issue here is that `this.handler = <CategoricalHandler | NumericHandler | > ...`, so I think the typescript compiler may have a hard time detecting issues at transpile time. But as pointed out yesterday, our spec.ts files can use the type definitions to make sure that arguments and results are as expected, so it will all other tests in the spec file more reliable.

server/shared/types/categorical.ts
```diff
@@ -0,0 +1,26 @@
+import { TermWrapper, BaseQ } from './termdb'
+
+/*
+--------EXPORTED--------
+CategoricalConditionQ
+CategoricalTW
+
+*/
+
+export type CategoricalConditionQ = BaseQ & {
```
siosonel, line 10: @creilly This is incorrect and/or confusing. Categorical tw.q should not be mixed in with condition tw.q, only the latter can have bar_by_children, value_by_*.

creilly8: Ahh, I see. I can go through it again and make those changes.

creilly8:
@siosonel, Maybe we should chat about your comments regarding the termsetting instance. The advantage of separating them out forces termsetting type properties to exist rather than be optional within the handlers. I was able to remove non-null assertions because of it. 

But I think all of this is related to the lack of specificity. Once all of the differences are well defined, then we can implement more elegant solutions.

siosonel (approved):
discussed with @creilly8 additional improvements in future branches/PR

creilly:
please take me off this thread I have nothing to do with this

PR #306, opened 2023-07-14
to change the neuro onc datasets matrix CNV panel hover over and legend

congyu-lu:


congyu-lu:
Please test review this PR together with sjpp: [neuro_onc_CNV_hover_over_and_legend](https://github.com/stjude/sjpp/tree/neuro_onc_CNV_hover_over_and_legend)

siosonel:
@xzhou82 Congyu and I created a `termdb.matrix.settings.ignoreCnvValues` option to force matrix plot to not use a chromatic scale even when a geneVariant `dt=4` has a `"value"` key-value. Do you agree with this solution, or should there *not* be a numeric `"value"` to indicate qualitative CNV data, so `"class": "CNV_amp" | "CNV_loss"` would have be sufficient like before? If yes, then no change in the pp and sjpp code needed. I think using numeric CNV value is confusing when it's really indicating a qualitative measure, not quantitative.

xzhou82 (approved):

Issue #313, opened 2023-07-16
Big gap between matrix controls and column label


siosonel:
Separated from https://github.com/stjude/proteinpaint/issues/293

LOGESH-B:
I am interested to do this

PR #314, opened 2023-07-16
Matrix label clip

siosonel:
Fixes https://github.com/stjude/proteinpaint/issues/313

To test, http://localhost:3000/testrun.html?name=matrix.integration, edit the last test in the spec file to `test.only`. Previously, the `Acute lymphoblastic leukemia` sample group label would be clipped. 

You can also test the `* Label Max Length` input to adjust in the Samples, Genes, Variable control menus.

The identified root cause for this issue is within the `adjustSvgDimensions()`, :
- the label g position is translated so that none of the label characters touches the control panel div above it
- however, the clip-path rect y and height do not account for the preceding adjustment, so it inadvertently clips the column labels

The solution is to also adjust the clip-path rect y and height in `adjustSvgDimensions()`.


xzhou82 (approved):
thanks a lot

karishma-gangwani:
image

I am getting some bundling errors for GDC since yesterday.

siosonel:
@karishma-gangwani The prod GDC API is down until this evening. I was able to test using the GDC VPN, where the uat API is active.

Screenshot 2023-07-16 at 10 43 03 AM

PR #317, opened 2023-07-17
Display the active divide-by label for matrix sample columns

siosonel:
Addresses JIRA # SV-2240.

The divide by label moves based on Cell Layout -> Group Label position inputs such as `s.collabelpos == 'top'`. Use the [integration test](http://localhost:3000/testrun.html?name=matrix.integration) or any dataset with matrix plots with a divide by term.
 
Screenshot 2023-07-16 at 8 52 44 PM


xzhou82 (approved):
very useful. thanks!

PR #320, opened 2023-07-17
Matrix sort improvements

siosonel:
- removed the asListed sample option: will require a `setting.sampleOrder` option to be easy to understand and implement
- relabeled the sample sorting options: will edit the gdc dataset file later to remove the duplicate "custom" option
- add sample grouping options

To test, just click around the new/improved sample sorting options.

Screenshot 2023-07-17 at 1 14 10 PM

### DIFF
--- a/client/plots/matrix.config.js
+++ b/client/plots/matrix.config.js
@@ -44,6 +44,7 @@ export async function getPlotConfig(opts = {}, app) {
 				sampleNameFilter: '',
 				sortSamplesBy: 'a',
 				sortOptions: getSortOptions(app.vocabApi.termdbConfig, controlLabels),
+				sortSampleGrpsBy: 'name', // 'hits' | 'name' | 'sampleCount'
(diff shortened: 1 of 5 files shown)


karishma-gangwani:
Is custom sort the same as the second option? I don't see any change when the second option is selected. 
By case name, id only shows ssm columns. There's no cnv. Is that correct?

siosonel:
> Is custom sort the same as the second option? I don't see any change when the second option is selected. 
Yes, they are the same. I will remove the custom sort which is configured in the GDC dataset js file, so that sort option will not show up afterwards.

> By case name, id only shows ssm columns. There's no cnv. Is that intended?
I see CNV with the `All GDC` filter. Also, it may be coincidence that "ealier" sample names don't have CNV data.

Screenshot 2023-07-17 at 2 44 44 PM

xzhou82 (approved):

PR #324, opened 2023-07-18
Use named import.

aacic:
I got an error while running:
cd container
./pack.sh
This should fix it.


siosonel:
@aacic I had already committed this change yesterday, because I also ran into the same issue. I can either approve/merge as needed, or please close if not.

aacic:
> @aacic I had already committed this change to master yesterday, because I also ran into the same issue. I can either approve/merge as needed, or please close if not.

I think we should merge this, I don't see the change in the master branch.

siosonel (approved):

PR #90, opened 2023-04-07
Tests.9

creilly8:
- Removed `sleep()` from a few tests
- New `sandbox.unit.spec`, 27 tests total
- New `menu.unit.spec`, 25 tests total
- Added missing todo comments in `menu.js` & `destroy()` for testing

PR #326, opened 2023-07-18
Improved menus as mentioned in sjpp issue #90

airenzp:
Removed padding and adjusted menu positions. Closes issue  stjude/sjpp#90. Also showed sample name on sample menu from the scatter plot.

### DIFF
--- a/client/dom/table.js
+++ b/client/dom/table.js
@@ -155,10 +155,7 @@ export function renderTable({
 	if (showLines) theadRow.append('td').style('width', '1vw')
 
 	if (buttons || noButtonCallback) {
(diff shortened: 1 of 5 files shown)


xzhou82 (approved):
thanks, follow up issues:
1. cosmetic: custom term button showing in GROUPS should have hover effect; click a custom term, the cascading menu should have same appearance as filter tvs pill menu (if easily doable)
2. in scatter, when multiple dots are overlapping, hover can show all dots in tooltip. click should do the same, to show all samples with plot options (disco) for each sample. otherwise user won't be able to access disco for those hidden behind

PR #345, opened 2023-07-22
cache all mappings to case uuid

xzhou82:
## Description

backend now caches mapping to case uuid from 3 types of ids (case submitter, sample submitter, aliquot)
when selecting cases from gdc lollipop or oncomatrix, it will fetch case id from cache, and no longer performs on the fly api query which crashes with hundreds of input ids

this affects both lollipop and oncomatrix

please help me test that sample selection from both allows to create valid GFF cohort

## Checklist


siosonel (approved):
It works as tested in both my local example html and v2 portal dev environment

PR #356, opened 2023-07-25
Option for a reset button in the recover component

siosonel:
## Description
Support a rewrite option and rewrite unit tests for the recover component.

Test with
- http://localhost:3000/testrun.html?name=recover.unit
- http://localhost:3000/example.gdc.matrix.html?maxGenes=3: only activated for GDC, reset/restore button is hidden by default

## Checklist


karishma-gangwani (approved):
works. thanks @siosonel. just one question. Why don't we use this for other datasets as well? I think it's a nice feature to be able to restore the initial session.

siosonel:
> works. thanks @siosonel. just one question. Why don't we use this for other datasets as well? I think it's a nice feature to be able to restore the initial session.

@karishma-gangwani Yes, I can add the reset/restore button as default for other datasets/plots as well, but later. It was easier to do for GDC since the embedded matrix is not part of MASS UI and also doesn't use a sandbox, where undo/redo icons are used instead of text as button labels. So will just need to figure what icons to use when using a sandbox and similar concerns, will do in another branch.

siosonel:
@karishma-gangwani @xzhou82 I had changed to draft after finding that the maxHistoryLen was causing the initial state to be dropped, now fixed to always have `history[0] = self.origState`.

xzhou82 (dismissed):

xzhou82 (dismissed):

PR #383, opened 2023-08-02
Numeric handler

creilly8:
## Description
Route all numeric terms through one `handlers/numeric.ts` rather than multiple scripts. In the future, merge `numeric.toggle.ts` code into new `numeric.ts` and simplify. 

All tests pass but I noticed problems with random regression analysis scenarios. I'm not able to come up with a scenario now that fails. @xzhou82, Would you please test with real regression scenarios?

## Checklist


xzhou82 (approved):
thank you so much; did a bunch of tests and all work

PR #423, opened 2023-08-16
ignore hidden values when conducting association tests

congyu-lu:
## Checklist

### DIFF
--- a/client/termdb/TermdbVocab.js
+++ b/client/termdb/TermdbVocab.js
@@ -173,6 +173,7 @@ export class TermdbVocab extends Vocab {
 		if ('minSampleSize' in opts) params.push(`minSampleSize=${opts.minSampleSize}`)
 
 		if (opts.term2) {
+			//send the hidden group labels to server to ignore them when computing association test pvalues
 			const hiddenValues = {
 				term1: opts.term.q.hiddenValues
 					? Object.keys(opts.term.q.hiddenValues).map(h => opts.term.term.values?.[h]?.label || h)
--- a/client/termdb/TermdbVocab.js
+++ b/client/termdb/TermdbVocab.js
@@ -172,6 +172,17 @@ export class TermdbVocab extends Vocab {
 		if ('grade' in opts) params.push(`grade=${opts.grade}`)
 		if ('minSampleSize' in opts) params.push(`minSampleSize=${opts.minSampleSize}`)
 
+		if (opts.term2) {
+			const hiddenValues = {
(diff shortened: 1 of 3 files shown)


client/termdb/TermdbVocab.js
```diff
@@ -172,6 +172,17 @@ export class TermdbVocab extends Vocab {
 		if ('grade' in opts) params.push(`grade=${opts.grade}`)
 		if ('minSampleSize' in opts) params.push(`minSampleSize=${opts.minSampleSize}`)
 
+		if (opts.term2) {
```
gavrielm, line 175: Would be useful to have a short comment here explaining why this is done

server/src/termdb.barchart.js
```diff
@@ -634,9 +634,20 @@ Output: The function has no return but appends the statistical results to the pr
 	]
 }
 */
-async function computePvalues(data, ds) {
+async function computePvalues(data, ds, hiddenValues) {
+	// remove hidden values from deep copy of data
+	const charts = JSON.parse(JSON.stringify(data.charts))
+	for (const chart of charts) {
+		chart.serieses = chart.serieses.filter(s => !hiddenValues.term1.includes(s.seriesId))
```
gavrielm, line 641: I don't think you should replace the same array with filtered array. Would be better to create new array.

server/src/termdb.barchart.js
```diff
@@ -634,9 +634,20 @@ Output: The function has no return but appends the statistical results to the pr
...
+		for (const s of chart.serieses) {
+			s.data = s.data.filter(d => !hiddenValues.term2.includes(d.dataId))
```
gavrielm, line 643: Same issue as above

server/src/termdb.barchart.js
```diff
@@ -634,9 +634,20 @@ Output: The function has no return but appends the statistical results to the pr
 	]
 }
 */
-async function computePvalues(data, ds) {
+async function computePvalues(data, ds, hiddenValues) {
+	// remove hidden values from deep copy of data
+	const charts = JSON.parse(JSON.stringify(data.charts))
```
gavrielm, line 639: You should just create a new object from scratch called `charts`. You can then iterate through `data.charts` and populate `charts` with the relevant information.

congyu-lu: Thanks, I will do that

gavrielm (approved):
Works well. Thank you for implementing this!

PR #463, opened 2023-08-23
Sample dictionary

airenzp:
## Description

This is a first draft to show sample data in the dictionary. It can be tested clicking on a sample from a scatter plot.  When the user has access to the data a menu option is added upon click to open the dictionary. If possible shows the disco plot and the methylation arrays to the right. It also allows to download all the data in tsv format.
I have tested it with several datasets. Testing the SJLife dataset is possible if a # is added to the SJLife credentials in the server config and a dynamic scatter is built to click on a sample.

## Checklist

### DIFF
--- a/client/plots/dictionary.js
+++ b/client/plots/dictionary.js
@@ -5,7 +5,7 @@ class MassDict {
 	constructor(opts) {
 		this.type = 'tree'
 		const div = opts.holder.append('div').style('display', 'flex')
-		const holder = div.append('div').style('padding', '20px').style('width', '30%')
+		const holder = div.append('div').style('padding', '20px')
 		const contentDiv = div.append('div')
 		this.dom = {
 			holder,
@@ -44,10 +44,11 @@ class MassDict {
 		if (this.sample && this.showContent) {
 			this.dom.holder
 				.style('border-right', '1px solid gray')
+				.style('min-width', '500px')
 				.style('overflow', 'scroll')
 				.attr('class', 'sjpp_hide_scrollbar')
 			this.dom.contentDiv
-				.style('width', '50%')
+				.style('width', '60%')
 				.style('min-height', '500px')
 				.style('display', 'flex')
 				.style('flex-direction', 'column')
--- a/client/plots/dictionary.js
+++ b/client/plots/dictionary.js
@@ -89,7 +89,7 @@ class MassDict {
 				for (const k in this.state.termdbConfig.queries.singleSampleGenomeQuantification) {
 					const div = this.dom.contentDiv.append('div').style('padding', '20px')
 					const label = k.match(/[A-Z][a-z]+|[0-9]+/g).join(' ')
-					div.append('div').style('font-weight', 'bold').text(label)
+					div.append('div').style('padding-bottom', '20px').style('font-weight', 'bold').text(label)
 
 					const ssgqImport = await import('./plot.ssgq.js')
 					await ssgqImport.plotSingleSampleGenomeQuantification(
@@ -97,7 +97,7 @@ class MassDict {
 						this.state.vocab.dslabel,
 						k,
 						this.sample,
-						div,
+						div.append('div'),
 						this.app.opts.genome
 					)
 				}
--- a/client/plots/dictionary.js
+++ b/client/plots/dictionary.js
@@ -4,7 +4,7 @@ import { appInit } from '#termdb/app'
 class MassDict {
 	constructor(opts) {
 		this.type = 'tree'
-		const div = opts.holder.append('div').style('display', 'flex')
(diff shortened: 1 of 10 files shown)


client/termdb/tree.js
```diff
@@ -422,7 +389,7 @@ function setRenderers(self) {
 
 		const isSelected = self.state.selectedTerms.find(t => t.name === term.name && t.type === term.type)
 		let text = term.name
-		if (term.isleaf && self.sampleData[term.id]) text += ` (${self.sampleData[term.id]})`
+		if (term.isleaf && self.sampleByTermId[term.id]) text += ` (${self.sampleByTermId[term.id]})`
```
siosonel, line 392: @airenzp should use optional chaining here, `sampleByTermId` may not be present.

Also rename `sampleByTermId` to `sampleDataByTermId`.

server/src/termdb.server.init.js
```diff
@@ -454,10 +454,14 @@ export function server_init_db_queries(ds) {
...
 	}
 
-	q.getSingleSampleData = function (sampleId) {
-		const sql = cn.prepare(
-			`select term_id, value from anno_categorical where sample=? union all select term_id, value from anno_float where sample=? union all  select term_id, value from anno_integer where sample=?`
-		)
+	q.getSingleSampleData = function (sampleId, term_ids) {
+		const ids = `('${term_ids.join("','")}')`
+
+		const query = `select term_id, value from anno_categorical where sample=? and term_id in ${ids}
```
siosonel, line 460: @airenzp Should not put variables directly in the sql string, should use the interpolated values as listed in the array argument to `sql.all(...)`. This is vulnerable to sql-injection attack.

airenzp: Should be solved now.

server/src/termdb.server.init.js
```diff
@@ -455,14 +455,14 @@ export function server_init_db_queries(ds) {
...
-		const query = `select term_id, value from anno_categorical where sample=? and term_id in ${ids}
-		union all select term_id, value from anno_float where sample=? and term_id in ${ids}
-		union all  select term_id, value from anno_integer where sample=? and term_id in ${ids} `
-
+		let ids = '?,'.repeat(term_ids.length)
+		if (term_ids.length) ids = ids.slice(0, ids.length - 1)
+		else ids = ''
+		const query = `select term_id, value from anno_categorical where sample=? and term_id in (${ids})
+		union all select term_id, value from anno_float where sample=? and term_id in (${ids})
+		union all  select term_id, value from anno_integer where sample=? and term_id in (${ids}) `
```
siosonel, line 463: @airenzp What happens if there is no `term_ids.length`, it seems the query will return zero results? There should not be a term_id condition in that case. I think this should be rewritten like this, including specifying a default term_ids argument value of an empty array:

```js
       q.getSingleSampleData = function (sampleId, term_ids=[]) {
	  const termClause = !term_ids.length ? '' : `and term_id in (${term_ids.map(t =>'?').join(','))})`
 		const query = `select term_id, value 
 			from anno_categorical 
 			where sample=? ${termClause}
			union all 
			select term_id, 
			value from anno_float 
			where sample=? ${termClause}
			union all  
			select term_id, value 
			from anno_integer 
			where sample=? ${termClause}`

		const rows = sql.all([sampleId, ...term_ids, sampleId, ...term_ids, sampleId, ...term_ids])
		return rows
	}
```

siosonel (changes requested):
Please see my comment and suggestion when term_ids is an empty array.

server/src/termdb.server.init.js
xzhou82: Airen and i talked about this today. i decided rather than running this sync query for sjlife which is scary, should instead read premade text files with data for single samples, as we're doing for mutation>disco. thus no need to modify this query

siosonel: @xzhou82 The anno tables are indexed by sample, there is no table scan, the query is very fast. Here, we even limit the number of terms to be queried, so even faster. No need to complicate with a separate file, that file read would likely be slower than the sqlite read.

client/plots/dictionary.js
```diff
@@ -4,13 +4,20 @@ import { appInit } from '#termdb/app'
...
-			holder: opts.holder.style('padding', '20px'),
+			holder,
+			contentDiv,
 			header: opts.header
 		}
 	}
 
 	async init(appState) {
+		const config = appState.plots.find(p => p.id === this.id)
+		this.sample = config.sample
```
xzhou82, line 19: later please add comment on the top, about these optional attributes and logic

xzhou82 (dismissed):

xzhou82 (approved):

xzhou82 (approved):

PR #479, opened 2023-08-25
Matrix term label hover over

congyu-lu:
## Checklist

### DIFF
--- a/client/plots/matrix.js
+++ b/client/plots/matrix.js
@@ -659,14 +659,12 @@ export class Matrix {
 		for (const t of this.termOrder) {
 			const countedSamples = new Set()
 			t.counts = { samples: 0, hits: 0 }
-			if (t.tw.term.type == 'geneVariant' || t.grp.name == 'Cytogenetics') {
-				t.counts.subGroupCounts = {}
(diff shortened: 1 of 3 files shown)


client/plots/matrix.js
```diff
@@ -659,6 +659,16 @@ export class Matrix {
 		for (const t of this.termOrder) {
 			const countedSamples = new Set()
 			t.counts = { samples: 0, hits: 0 }
+			if (t.tw.term.type == 'geneVariant' || t.grp.name == 'Cytogenetics') {
```
xzhou82, line 662: shouldn't hardcode grp name

congyu-lu: Removed 'Cytogenetics'

xzhou82:
over row label, hover and click should not share the same menu, otherwise you will never be able to edit. create a new one for hover

congyu-lu:
> over row label, hover and click should not share the same menu, otherwise you will never be able to edit. create a new one for hover

changed to hovering over won't dismiss edit menu

xzhou82 (approved):
thanks it works nicely. i would like you to add comments to explain the logic and purpose, where it's not obvious

PR #499, opened 2023-08-30
declare queries.snvindel.populations in Mds3 typescript

xzhou82:
## Description

it's not doing the job on my computer. restarting backend won't help
test by introducing errors into sjpp/dataset/sjlife2.hg38.ts

## Checklist

### DIFF
--- a/server/shared/types/dataset.ts
+++ b/server/shared/types/dataset.ts
@@ -171,6 +171,35 @@ type VariantFilter = {
 	terms: FilterTermEntry[]
 }
 
+// one set of AC and AN info fields to retrieve data for this population
+type PopulationINFOset = {
+	key?: string // optional name for identifying this set, when the population is ancestry-stratified and a population has multiple sets
+	infokey_AC: string // required info field
+	infokey_AN: string // required info field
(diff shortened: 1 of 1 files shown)


server/shared/types/dataset.ts
```diff
@@ -171,6 +171,35 @@ type VariantFilter = {
...
+/* define method to retrieve allele AC/AN in a population, by using bcf INFO fields; population could be ancestry-stratified
+two types of population are supported:
+- ancestry-stratified
+  allowto_adjust_race can be set to true
+  sets[] has >1 elements
+- not stratified
+  allowto_adjust_race cannot be set to true
+  sets[] has only 1 element
+*/
+type Population = {
```
creilly8, line 191: @xzhou82 exporting Population and adding directly to the ```populations``` fixes this issue (i.e. it's forced as existing instead of undefined as runtime). That works for now until we can implement mapped types.

xzhou82: thanks for now i won't change like that. let's merge as is and figure out how to get ts to validate elements of an optional array. there must be a way

creilly8 (approved):

PR #550, opened 2023-09-05
Visualizing multiple samples from scatter lasso

airenzp:
## Description

To test it please open a scatter plot, select a few samples and click on `Show samples`. A samples table is now dynamically built and autoaligned. I display on top the samples selected instead of all the samples as I think it may be of use( selecting samples allows the user to focus on specific samples). I discussed with @siosonel details about the implementation. 
## Checklist

### DIFF
--- a/client/termdb/tree.js
+++ b/client/termdb/tree.js
@@ -224,7 +224,6 @@ class TdbTree {
 		for (const term of terms) term_ids.push(term.id)
 		for (const sample of this.state.samples) {
 			const data = await this.app.vocabApi.getSingleSampleData({ sampleId: sample.sampleId, term_ids })
-			console.log(data)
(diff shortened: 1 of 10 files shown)


xzhou82:
bug in single sample view: expand the first branch, then inside the children terms expand a sub branch. then, collapse the first branch to see the defect

client/termdb/tree.js
```diff
@@ -321,30 +335,87 @@ function setRenderers(self) {
 		select(this).style('display', 'none')
 	}
 
+	self.renderSampleTable = function () {
```
siosonel, line 338: @airenzp I suggest moving the sample table renderer functions at the end of `self.renderers()`, since these logically happen only after the tree menu is rendered and we want to keep the tree branch renderer function closer together.

PR #567, opened 2023-09-07
fix the user-created matrix gene row label hover tooltip error

congyu-lu:
## Checklist


xzhou82:
thanks for the fix, but this causes inconsistency (327 indicated in row label, 370 from tooltip)
image
what is the rule of excluding dt=cnv in the 327 count? can that rule be applied to tooltip so that cnv class count is excluded from 370 as well, and the cnv class breakdown can be listed separately from mutation classes?

xzhou82 (approved):

PR #591, opened 2023-09-14
cache gdc case ids with exp data

xzhou82:
## Description

new data are silently cached on back when server launches, should not affect any mds3/matrix features. mds3 tape tests pass

the `features."stopGdcCacheAliquot":5` flag of serverconfig allows finer control of the caching behavior on your local instance:

1. set value to `true` to disable caching
3. set to integer to only cache ?x1000 cases
    this will be printed at console (depending on integer value)

> GDC: Done caching sample IDs 150 s
> 	 8207 aliquot IDs to sample submitter id
> 	 21958 different ids to case uuid.
> 	 615 cases with exp data.

3. delete the flag to cache all cases
    this will be printed at the console
> GDC: Done caching sample IDs 147 s
> 	 265088 aliquot IDs to sample submitter id
> 	 485023 different ids to case uuid.
> 	 18799 cases with exp data.

the expression api is only visible to those with IP whitelisted by gdc, if not the `checkExpressionAvailability()` function should be skipped, same should happen when it's deployed to our servers. thus i want Congyu to test it to make sure it does not break anything

## Checklist

### DIFF
--- a/server/src/termdb.gdc.js
+++ b/server/src/termdb.gdc.js
@@ -860,15 +860,16 @@ async function cacheSampleIdMapping(ds) {
 		casesWithExpData: new Set()
 	}
 
+	// caching action is fine-tuned by the feature toggle on a pp instance; log out detailed status per setting
 	if ('stopGdcCacheAliquot' in serverconfig.features) {
 		// flag is set
 		if (Number.isInteger(serverconfig.features.stopGdcCacheAliquot)) {
 			// flag value is integer (suppose to be positive integer)
-			// allow to test on dev machine
+			// allow to run a short test on dev machine
 			console.log('GDC: running limited sample ID caching')
 		} else {
 			// flag value is not integer, do not run this function at all
-			console.log('GDC sample IDs are not cached!')
+			console.log('GDC: sample IDs are not cached!')
 			return
 		}
 	} else {
@@ -899,10 +900,10 @@ async function cacheSampleIdMapping(ds) {
 
 		await checkExpressionAvailability(ds)
 
-		console.log('GDC: Done caching sample IDs', Math.ceil((new Date() - begin) / 1000), 's')
+		console.log('GDC: Done caching sample IDs. Time:', Math.ceil((new Date() - begin) / 1000), 's')
 		console.log('\t', ds.__gdc.aliquot2submitter.cache.size, 'aliquot IDs to sample submitter id')
 		console.log('\t', ds.__gdc.map2caseid.cache.size, 'different ids to case uuid.')
-		console.log('\t', ds.__gdc.casesWithExpData.size, 'cases with exp data.')
+		console.log('\t', ds.__gdc.casesWithExpData.size, 'cases with gene expression data.')
 	} catch (e) {
 		console.log('Error at caching: ' + e)
 	}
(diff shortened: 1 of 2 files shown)


server/src/termdb.gdc.js
```diff
@@ -936,12 +1006,45 @@ async function fetchIdsFromGdcApi(ds, size, from, aliquot_id) {
...
+	// hardcodes this url since the api is only in uat now. replace with path.join() when it's in prod
+	const url = 'https://uat-portal.gdc.cancer.gov/auth/api/v0/gene_expression/availability'
+
+	{
+		/*
+		when the api is released in production, delete this check and move it into testGDCapi()
+		exp api is only available on uat now, thus it works if the pp server ip is whitelisted, otherwise won't work and will skip this function
+		*/
+		const c = await testRestApi(url)
+		if (c == 'ERR_NON_2XX_3XX_RESPONSE') {
```
xzhou82, line 1029: @siosonel i use error code with string value to detect if the api is accessible. as i saw that from a non-whitelisted IP it returns a different code. do you think it's okay?

server/src/termdb.gdc.js
```diff
@@ -715,15 +720,15 @@ async function getOpenProjects(ds) {
...
-	if (re?.data?.aggregations?.['cases.project.project_id']?.buckets) {
-		for (const b of re.data.aggregations['cases.project.project_id'].buckets) {
-			// key is project_id value
-			if (b.key) ds.gdcOpenProjects.add(b.key)
-		}
-		console.log('GDC open-access projects:', ds.gdcOpenProjects.size)
-	} else {
-		console.log("getting open project_id but return is not re.data.aggregations['cases.project.project_id'].buckets")
+	if (!Array.isArray(re?.data?.aggregations?.['cases.project.project_id']?.buckets)) {
+		console.log("getting open project_id but return is not re.data.aggregations['cases.project.project_id'].buckets[]")
```
karishma-gangwani, line 724: the message is not very clear.

server/src/termdb.gdc.js
```diff
@@ -837,11 +888,21 @@ async function cacheSampleIdMapping(ds) {
...
 		}
 
+		await checkExpressionAvailability(ds)
+
 		console.log('GDC: Done caching sample IDs', Math.ceil((new Date() - begin) / 1000), 's')
-		console.log('\t', ds.aliquot2submitter.cache.size, 'aliquot IDs to sample submitter id')
-		console.log('\t', ds.map2caseid.cache.size, 'different ids to case uuid.')
+		console.log('\t', ds.__gdc.aliquot2submitter.cache.size, 'aliquot IDs to sample submitter id')
+		console.log('\t', ds.__gdc.map2caseid.cache.size, 'different ids to case uuid.')
+		console.log('\t', ds.__gdc.casesWithExpData.size, 'cases with exp data.')
```
karishma-gangwani, line 905: use the full form for exp - expression.

server/src/termdb.gdc.js
xzhou82: it's just a complain for broken data which should not happen

server/src/termdb.gdc.js
xzhou82: improved

karishma-gangwani (approved):

PR #600, opened 2023-09-17
clean up oncomatrix launcher that cgconly option is no longer generat…

xzhou82:
…ed by this code

## Description

i was about to make a launchGDCexp.js so I reviewed launchGdcMatrix.js and did some clean up. no UI change is introduced

## Checklist

### DIFF
--- a/client/src/launchGdcMatrix.js
+++ b/client/src/launchGdcMatrix.js
@@ -1,14 +1,12 @@
 import { dofetch3 } from '#common/dofetch'
-import { make_one_checkbox } from '#dom/checkbox'
 import { appInit } from '#plots/plot.app.js'
 import { fillTermWrapper } from '#termsetting'
 
 /*
-
 test with http://localhost:3000/example.gdc.matrix.html
 runpp({ launchGdcMatrix:true })
 
-designed to work for Oncoprint app in GDC Analysis Tools Framework
+designed to work for OncoMatrix app in GDC Analysis Tools Framework
 
 initiates a "plot" app instance that wraps around the matrix component
 and returns the plot app API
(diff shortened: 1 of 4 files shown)


client/src/launchGdcMatrix.js
```diff
@@ -219,16 +207,16 @@ async function getGenes(arg, gdcCohort, CGConly, maxGenes = 50) {
...
+		geneFilter: matrix.geneFilter
 	}
-	if (CGConly) body.CGConly = 1
 	const data = await dofetch3('gdc_filter2topGenes', { body })
 	if (data.error) throw data.error
 	if (!data.genes) throw 'no top genes found using the cohort filter'
 	return await Promise.all(
 		// do tempfix of "data.genes.slice(0,3).map" for faster testing
-		data.genes.map(async i => {
+		data.genes.slice(0, 3).map(async i => {
```
xzhou82, line 219: must delete before merging

client/src/launchGdcMatrix.js
```diff
@@ -137,14 +148,11 @@ export async function init(arg, holder, genomes) {
 							chartType: 'matrix',
 							settingsKey: 'maxGenes',
 							callback: async value => {
```
xzhou82, line 150: i think this max number gene limit should not be defined here as a custom input. it should be coded in matrix and applied to other dataset. can assign to Congyu?

xzhou82:
@airenzp need help to update `dom/genesetEdit.js` around line 50 on the optional CGC checkbox. please pull sjpp master to see the change (sorry accidentally committed to master, was to make a PR)

>                 topMutatedGenes: {
> -                       arguments: [{ id: 'CGConly', label: 'Use only cancer census genes', type: 'boolean', value: true }]
> +                       arguments: [{ id: 'geneFilter', label: 'Use only cancer census genes', type: 'string', value: 'CGC' }]

karishma-gangwani (approved):

karishma-gangwani (approved):
all tests pass.

xzhou82:
CGC only option is no longer showing due to sjpp change, i would like Airen to fix it in this branch
also want to document args {settings, matrix} from Edgar

client/src/launchGdcMatrix.js
```diff
@@ -1,31 +1,43 @@
...
+		// and is supplied by the OncoMatrixWrapper in the GFF repo, no need to supply here
+		// return { op: 'in', content: { field: 'cases.disease_type', value: ['Gliomas'] } }
 
 	.genes[]
 		future: list of genes to launch matrix
 
+	.termgroups[]
+		TODO 
+	
+	.settings{}
```
xzhou82, line 36: @siosonel could you comment on the reason for needing two attributes `arg.settings.matrix{}, arg.opts.matrix{}`

siosonel: @xzhou82 
1. `arg.settings.matrix`: any state that can be JSON-encoded/hydrated as part of `getPlotConfig().settings.matrix`. This is meant to hold plot-instance-specific state that only affects rendering and not data requests, but that distinction is not as clear with plots that use server-side rendering like violin plot.

2. `arg.plots.matrix`:  any options that cannot be JSON-stringified like callbacks. These options will also be applied to all rehydrated or new plots of the same type, so not plot-instance-specific.

client/src/launchGdcMatrix.js
siosonel: @congyu-lu we can discuss how to move this code to matrix.

airenzp:
Hi. I am looking at it Xin. I see the parameter type changed from a boolean to a string, so do you expect a string to be entered instead of a boolean?

airenzp:
I pushed a fix to handle the parameter and added a PR in sjpp that should be merged with this one.

PR #602, opened 2023-09-18
Fixed last bugs reported by Karishma

airenzp:
## Description

Closes issue stjude/sjpp#129. Some bugs where closed in a previous PR and now this is done: The scale is placed on the legend properly. The similar categories are differentiated. Please check.

## Checklist

### DIFF
--- a/client/plots/sampleScatter.js
+++ b/client/plots/sampleScatter.js
@@ -204,7 +204,8 @@ class Scatter {
 			chartType: 'sampleScatter',
 			settingsKey: 'minDotSize',
 			title: 'Minimum area, in square pixels',
-			min: 0
+			min: 9,
+			max: 3000
 		}
 		const maxDotSizeOption = {
 			label: 'Max area',
@@ -212,7 +213,8 @@ class Scatter {
 			chartType: 'sampleScatter',
 			settingsKey: 'maxDotSize',
 			title: 'Maximum area, in square pixels',
-			min: 0
+			min: 9,
+			max: 3000
 		}
 		const orientation = {
 			label: 'Scale order',
--- a/client/plots/sampleScatter.js
+++ b/client/plots/sampleScatter.js
@@ -33,7 +33,8 @@ NOTE
 "sample" and "category" attributes here are hardcoded
 
 */
-
+export const minDotSize = 9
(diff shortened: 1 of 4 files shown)


karishma-gangwani:
Thanks @airenzp I see a bug with the scaling legend when you zoom in. Do we want to make that a fixed size? Because the dots can get really large on the scale.

airenzp:
Thanks for the feedback Karishma. I added validations to limit the particles size on zoom and when editing the min and max values. Also the scale legend should be placed properly now in the different setups. Please let me know if you find any pending issues.

client/plots/sampleScatter.js
```diff
@@ -204,15 +204,17 @@ class Scatter {
...
 		}
 		const maxDotSizeOption = {
 			label: 'Max area',
 			type: 'number',
 			chartType: 'sampleScatter',
 			settingsKey: 'maxDotSize',
 			title: 'Maximum area, in square pixels',
-			min: 0
+			min: 9,
+			max: 3000
```
karishma-gangwani, line 217: why is max 3000? I think that's too large. in all-pharma, I can see the dot size increasing and overlapping the scale label.

client/plots/sampleScatter.js
```diff
@@ -204,15 +204,17 @@ class Scatter {
...
 		}
 		const maxDotSizeOption = {
 			label: 'Max area',
 			type: 'number',
 			chartType: 'sampleScatter',
 			settingsKey: 'maxDotSize',
 			title: 'Maximum area, in square pixels',
-			min: 0
+			min: 9,
```
karishma-gangwani, line 216: also the min. and max. should not be the same I think since this is a scaling. Do you think we should add some separate limits that can be reversed based on the scale order?

karishma-gangwani (approved):

client/plots/sampleScatter.js
airenzp: I just saw your comments Karishma. The min and max are just a validation to avoid that nothing breaks. And I think now nothing breaks or looks out of place. Thanks for the feedback. I am happy to fix bugs but we cant make it perfect at once. I need to work on some other stuff. But feel free to improve it!

PR #653, opened 2023-09-27
Fix disco plot bundling issue.

aacic:
## Checklist

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": "./"
+    "baseUrl": "./",
+    "target": "es2016"
   },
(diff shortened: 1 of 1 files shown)


siosonel (approved):

xzhou82:
what's the manifestation, why everything works on local and prp1 and only breaks on prp3, can you document what you learned

siosonel:
> what's the manifestation, why everything works on local and prp1 and only breaks on prp3, can you document what you learned

@xzhou82 
- `prp3` uses releaseFetcher, which fetches the published `@sjcrh/proteinpaint-client` package that was bundled with rollup, plus proteinpaint-front which was bundled by webpack.
- local and `prp1` uses bundles from webpack only, not rollup. For sure webpack defaults to browser bundles so it's less likely to break client-side features, but rollup may default to esm which the tsconfig can override for client-side bundles.

PR #672, opened 2023-09-29
always render consequences over CNV for oncoprint

congyu-lu:
## Description
closing sjpp issue (173) 

## Checklist

### DIFF
--- a/client/plots/matrix.js
+++ b/client/plots/matrix.js
@@ -1117,7 +1117,20 @@ export class Matrix {
 				}
 
 				const key = anno.key
-				const values = anno.filteredValues || anno.values || [anno.value]
+
+				let values
+				if (anno.filteredValues) {
+					if (t.tw.term.type !== 'geneVariant' || s.cellEncoding != 'oncoprint') values = anno.filteredValues
+					else {
+						const sortedFilteredValues = []
+						for (const dt of [4, 1, 3]) {
+							const v = anno.filteredValues.filter(v => v.dt === dt)
+							sortedFilteredValues.push(...v)
+						}
+						values = sortedFilteredValues
+					}
+				} else values = anno.values || [anno.value]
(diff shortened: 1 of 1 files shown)


xzhou82 (approved):

xzhou82 (approved):
thanks a lot. in OncoPrint mode, at a cell with multiple dt=1 it should still render all of them as equal-width slivers so there will be a visual indication of such and we don't need logic to prioritize which event to show (right now it may show intronic rather than missense in such case), also it won't incur performance penalty as it's canvas

client/plots/matrix.js
```diff
@@ -1117,7 +1117,20 @@ export class Matrix {
 				}
 
 				const key = anno.key
-				const values = anno.filteredValues || anno.values || [anno.value]
+
+				let values
```
siosonel, line 1121: @congyu-lu why doesn't lines 607-610 or 613 in `classifyValues()` work? In other words, I think we should apply `this.stackSiblingCellsByClass` to the filteredValues also, and not do it within `getSerieses()`

```js
                   else {
			// dt=1 are SNVindels, dt=4 CNV, dt=3 Gene Expression
			// will render only one matching value per dt
			for (const dt of [4, 1, 3]) {
				const v = filteredValues.find(v => v.dt === dt)
				if (v) renderedValues.push(v)
			}
		}
		// group stacked cell values to avoid striped pattern *** !!! apply to filteredValues? !!!!
		if (tw.term.type == 'geneVariant') renderedValues.sort(this.stackSiblingCellsByClass)
```

congyu-lu: Lines 607-610 only work for renderedValues. I will move the code sort filteredValues based on dt to classifyValues and also apply this.stackSiblingCellsByClass to filteredValues

PR #720, opened 2023-10-12
disable listing samples temp.

karishma-gangwani:
## Checklist

### DIFF
--- a/client/plots/barchart.events.js
+++ b/client/plots/barchart.events.js
@@ -436,24 +436,13 @@ function handle_click(event, self, chart) {
 			})
 		}
 	}
-	if (self.config.displaySampleIds) {
-		options.push({
-			label: 'List samples',
-			callback: () => listSamples(event, self, seriesLabel, dataLabel, chart)
-		})
-	}
-
-	// list samples for a category
-	// this option is always added
-	/*
-	// after discussing with yutaka this is disabled for now
-
-
-	options.push({
-		label: 'List samples',
-		callback: menuoption_listsamples
-	})
-	*/
+	//disable sample listing temporarily
+	// if (self.config.displaySampleIds) {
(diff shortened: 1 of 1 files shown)


xzhou82 (approved):

siosonel:
@xzhou82 @karishma-gangwani should this have been done in the `sjlife.hg38.js` dataset file, instead of commenting out the code? Also not sure if I need to fix something with auth, if a user is logged into vizcom and the `termdbConfig.displaySampleIds` flag is true, then that option will show up - I assume this is what happened and not that the user was not logged in and was still able to display sample IDs.
```js
   cohort: {
        ...,
	termdb: {
             displaySampleIds: true, // change to false
```

karishma-gangwani:
@siosonel I see. I would think as a user if I am not logged in I shouldn't see that button because it doesn't do anything or if I click the button, it should show me a message to log in to view the samples. since I was logged in for **mbmeta** on prp1, I could see the button but clicking the button didn't load the samples.

siosonel:
@karishma-gangwani Thanks, I now see that confusing mbmeta `List samples` menu option, I only tested with vizcom SJLife after seeing this commit. I work on fixing this.

karishma-gangwani:
thanks @siosonel

siosonel:
@karishma-gangwani Not having the listing is related to this quick fix, since lthe sample ID-name mapping was removed from the barchart response data due to performance issues. So, the code to show the `List sample` needs to be commented as was done here, and reactivated at the same time that the todo (or alternative solution) in the screenshot is done.

Screenshot 2023-10-12 at 4 53 08 PM

karishma-gangwani:
sounds good @siosonel violin also uses that so I can try to work on enabling this.

PR #751, opened 2023-10-23
Restoring list samples in barchart

airenzp:
## Description

Restored list samples using getData with a filter.

## Checklist

### DIFF
--- a/client/plots/barchart.events.js
+++ b/client/plots/barchart.events.js
@@ -516,6 +516,7 @@ async function listSamples(event, self, seriesId, dataId, chartId) {
 		maxHeight: '40vh',
 		resize: true
 	})
+	menu.show(event.clientX, event.clientY, false)
 
 	function getTvs(termIndex, value) {
 		const term = termIndex == 0 ? self.config.term0 : termIndex == 1 ? self.config.term : self.config.term2
@@ -532,8 +533,6 @@ async function listSamples(event, self, seriesId, dataId, chartId) {
 		}
 		return tvs
 	}
-
-	menu.show(event.clientX, event.clientY, false)
 }
 
 function menuoption_add_filter(self, tvslst) {
--- a/client/plots/barchart.events.js
+++ b/client/plots/barchart.events.js
@@ -437,12 +437,12 @@ function handle_click(event, self, chart) {
 		}
 	}
 	//disable sample listing temporarily
-	// if (self.config.displaySampleIds) {
-	// 	options.push({
-	// 		label: 'List samples',
-	// 		callback: () => listSamples(event, self, seriesLabel, dataLabel, chart)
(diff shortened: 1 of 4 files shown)


siosonel:
@airenzp I think the way that the violin plot does this is lighter, where the sample list is only for the clicked data. It does that by adding the data values of the clicked segment/area as tvs filters to the global filter. That approach avoids downloading sample ID lists when it's not needed yet; in contrast, adding sample IDs to the `/termdb-barsql?includeSample=true` will still have the slowdown and bloated payload when there are many more samples > 10000.

```js
  // from violin.interactivity.js
        
         self.listSamples = async function (event, t1, t2, plot, start, end) {
		const tvslst = self.getTvsLst(t1, t2, plot, start, end)
		const term = t1.q?.mode === 'continuous' ? t1 : t2
		const filter = {
			type: 'tvslst',
			join: 'and',
			lst: [self.state.termfilter.filter, tvslst],
			in: true,
		}
		const opts = {
			terms: [term],
			filter,
		}
		//getAnnotatedSampleData is used to retrieve sample id's and values (see matrix.js).
		const data = await self.app.vocabApi.getAnnotatedSampleData(opts)
		self.displaySampleIds(event, term, data)
	}
```

airenzp:
Thanks for the feedback @siosonel. I used the idea of the filter to request only the samples listed. Please check. Also I noticed that the list samples option is not displayed on the violin.

xzhou82:
to add to Edgar's comment, Yutaka has a pending request to not allow to filter down to less than 10 samples in sjlife, for fear of making the data identifiable.
in the barchart return data, if it supplies sample ids for every category, despite the ids are "deidentified", you can still know what id(s) are for the small categories, even category with a simple sample. that's practically identifying an sample and defeat the purpose of protection.

our summary plots can only return summaries, and not any sample-level detail, not in any way allow to associate annotation to an individual sample. this ensures sample-level data protection. a clear line in the sand. drill it to everyone's mind

karishma-gangwani:
> Thanks for the feedback @siosonel. I used the idea of the filter to request only the samples required. Please check. Also I noticed that the list samples option is not displayed on the violin. @karishma-gangwani maybe you want to check this line `displaySampleIds: app?.getState()?.termdbConfig?.displaySampleIds ? true : false, `

Thanks for the pr @airenzp I am able to see the list samples on barchart and violin both. Maybe you can show me what you're seeing.

karishma-gangwani (approved):

xzhou82 (approved):
thanks

airenzp:
When I use SJLife @karishma-gangwani, and have commented out the authentication check on the serverconfig, so it assumes I am logged in, the list samples is not shown on the violin. I can show you.

PR #760, opened 2023-10-25
fix matrix legend filter issues

congyu-lu:
## Description

After "Hide samples with event" (hard). legends are crossed out, after "Do not show event"(soft), legends are greyed out. 

Removed "show only" for geneVariant term, but kept "show all"

Still show legends after all samples are removed or hidden

Rename "filter" to "hide" for non-genevariant legends

Legends are no longer clickable after legend groups are hidden.

will optimize tvs structure to get ride of hard/soft flag in another branch after figuring out best structure for soft filter

## Checklist

### DIFF
--- a/client/dom/svg.legend.js
+++ b/client/dom/svg.legend.js
@@ -121,6 +121,7 @@ export default function svgLegend(opts) {
 		const g = select(this)
 			.attr('transform', 'translate(' + currlinex + ',' + currliney + ')')
 			.style('opacity', settings.itemOpacity)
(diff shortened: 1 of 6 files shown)


siosonel (changes requested):
@congyu-lu Everything works as decribed in the PR, thanks! The only thing is the order of the legend entries moves around after clicking to hide. Is that intended, or should there be sorting to avoid that?

siosonel (approved):

PR #767, opened 2023-10-26
Rounding decimals to set a filter value

### LINKED ISSUE #684: Should remove automatic rounding from tvs
[sjlife portal](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22SJLife%22}) -> `FILTER tab` -> Search for term `PGS000001 breast carcinoma (Derived from European ancestry; Computed for European ancestry) (MAF>1%)` -> set filter to `x>0.75` -> Will automatically round to `x>0.8`

Should not round entered values. Same issue is observed in `GROUPS` tab.

airenzp:
## Description

Allowed two decimals when setting a filter value. Closes #684 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Fixes:
 - Disco: file name of downloaded svg preserves context including sample name and gene.
 - Disco: by default it now prioritize CGC genes in hg38.
 - Various scatterplot fixes: tooltip dot scale, divideBy bug
+- Fix the session recovery within an embedder portal
 
 
 ## 2.30.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 Fixes:
 - In mass sampleview, expanding tree branch everytime no longer triggers rerendering of disco/ssgq plots
+- Disco: file name of downloaded svg preserves context including sample name and gene.
+- Disco: by default it now prioritize CGC genes in hg38.
+- Various scatterplot fixes: tooltip dot scale, divideBy bug
 
 
 ## 2.30.1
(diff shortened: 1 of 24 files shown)


xzhou82:
please make a change and detect term type. if type=integer, do not set toFixed().
[test at wgs_sample_age](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:3},%22termfilter%22:{%22filter%22:{%22type%22:%22tvslst%22,%22join%22:%22%22,%22in%22:true,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22wgs_sample_age%22},%22ranges%22:[{%22start%22:10,%22stop%22:20,%22stopinclusive%22:true}]}}]}}}) which is integer. drag brush should not print decimal ages

if type=float then use toFixed(2) for now. this is still problematic. if a term value are like 0.0001 and 0.0002 etc, the ui won't work. i think we can add some logic to compute a reasonable precision level by density plot min/max values

gavrielm:
`FILTER` -> `Age (years) at Cancer Diagnosis` -> set filter to `x>5` and press ENTER -> UI shows `5 < x < 24.79`.

I think the range should be unbounded (i.e., `5<x`), which is how it is on master

gavrielm:
I committed the following changes:
- Use the roundValue() module to round the values (I made this module a long time ago)
- Round min and max in the same way so that unbounded ranges can be displayed as unbounded in the UI

airenzp:
@gavrielm as a proof of concept I did a PR on top of this one so you can see/test the changes I suggest. Please check.

PR #769, opened 2023-10-27
prototype a session tracking feature with clearer options

siosonel:
## Description
Also, prefer local cache/file storage, in addition to the option to save in the server or share. ~~This is a draft, will do more integration testing later and include release notes if this approach looks okay to pursue~~.

NOTE: The `Share` option behaves almost the same as the current `Save Session` feature, but notes around expiration may have been lost. Someone else can work on that in another PR, since there is also an open question from @creilly8 on this. 

Test Scenarios for `Open` and `Save`: for each scenario, test that you can save to/open from browser, file, and if applicable, server.

A. Unrpotected ds: comment `serverconfig.dsCredentials.SJLife` (e.g., change to `dsCredentials['#SJLife']`):
  - use [a simple SJLife URL params](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22SJLife%22})
  - there should be no server options in the `Open` and `Save` menus 

B. Protected ds: enable `dsCredentials.SJLife`
  - use http://localhost:3000/example.auth.html
  - there should **be** server options in the `Open` and `Save` menus: I think this matches the requirements in Xin's drawing, but using button + menus instead of tab + tray
  - test while logged in and not
  - when saving to server, you should see the file saved in `[cachedir]/sessionsByCred/...`

C. Same as B, but simulate an embedded use case
  - `sudo vim /etc/hosts` and add the entry `127.0.0.1 viz.localhost`
  - change `dsCredentials['SJLife']` to `dsCredentials['*']` - all embedders will use the same jwt secret
  - from the `sjpp` dir, run `./proteinpaint/server/test/fake-jwt.js SJLife > public/fake-jwt.json '*'`
  - use http://viz.localhost:3000/example.auth.html (different subdomain)
  - manually test like in B
  - Also test that the `Share`-generated URL opens within `viz.localhost`, and not in `localhost`

The goal is to emulate familiar behavior, such as the `File: > open | save | share | ...` options that are very typical in desktop apps.

Screenshot 2023-10-27 at 6 14 49 PM
Screenshot 2023-10-27 at 6 15 42 PM
Screenshot 2023-10-27 at 6 16 02 PM
Screenshot 2023-10-27 at 6 16 19 PM

## Checklist

### DIFF
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -3,6 +3,7 @@ import { recoverInit } from '../rx/src/recover'
 import { searchInit } from './search'
 import { chartsInit } from './charts'
 import { groupsInit } from './groups'
+import { sessionBtnInit } from './sessionBtn'
 import { select } from 'd3-selection'
 import { dofetch3 } from '#common/dofetch'
 import { Menu } from '#dom/menu'
@@ -96,6 +97,12 @@ class TdbNav {
 					getState: appState => appState,
 					reactsTo: action => action.type != 'plot_edit',
 					maxHistoryLen: 5
+				}),
+				sessionBtn: sessionBtnInit({
+					app: this.app,
+					button: this.dom.saveBtn,
+					massSessionDuration: this.opts.massSessionDuration,
+					sessionDaysLeft: this.app.opts.sessionDaysLeft || null
 				})
 			})
 			this.mayShowMessage_sessionDaysLeft()
@@ -269,11 +276,9 @@ function setRenderers(self) {
 		self.dom.tds = table.selectAll('td')
 		self.subheaderKeys = self.tabs.map(d => d.subheader)
 
-		self.dom.saveBtn = self.dom.sessionDiv
-			.append('button')
-			.style('margin', '10px')
-			.text('Save Session')
-			.on('click', self.getSessionUrl)
+		self.dom.saveBtn = self.dom.sessionDiv.append('button').style('margin', '10px').text('Session ▼')
+
+		//.on('click', self.getSessionUrl)
 
 		if (self.sessionDaysLeft != null) {
 			//Only show if called from `mass-session-id` URL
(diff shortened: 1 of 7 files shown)


karishma-gangwani (approved):
Thanks @siosonel. I like you added the share option. I think if this is going to be released then you could replace the TODO with 'feature not enabled yet' or something.

xzhou82:
looks nice. right now it no longer saves on server. i still want an option to save on server, client saving won't work when one person browse portal on multiple devices, and sharing is a hassle

xzhou82:
eventually we will let user see all their saved sessions when they are logged in via viz, from any device.

creilly8:
@siosonel, I uploaded a mass session file from master and worked it great. The download and upload worked well for me too. Unless user feedback suggests another layout, I think this looks great. 

Side note: There's already a share button at the top of every viz in vizcom, albeit very small and in the upper left hand corner. Maybe utilize that? It's probably too much work for too little gain but could be an option later. If anything it would be a good test run for building an API to save to the customer's server rather than our own.

creilly8:
I'm also a big fan of the upload because the exported file downloads to Documents or Desktop.

creilly8:
@xzhou82 Does this mean we will keep session files longer than 30 days (the current limit)? In other words, will this new feature impact how we store session files on the server?

siosonel:
> @xzhou82 Does this mean we will keep session files longer than 30 days (the current limit)? In other words, will this new feature impact how we store session files on the server?

@creilly8 I coded the `Share` feature to behave almost the same as the current `Save Session` feature, so it may still be good to keep track of session storage expirations.

xzhou82 (approved):
thanks

siosonel:
> @siosonel, I uploaded a mass session file from master and worked it great. The download and upload worked well for me too. Unless user feedback suggests another layout, I think this looks great.
> 
> Side note: There's already a share button at the top of every viz in vizcom, albeit very small and in the upper left hand corner. Maybe utilize that? It's probably too much work for too little gain but could be an option later. If anything it would be a good test run for building an API to save to the customer's server rather than our own.

@creilly8 Using the embedder's share button will be preferred, but yes, it would require some coordination with vizcom so something we could design and plan

PR #771, opened 2023-10-27
Deciding decimal places on numeric filter dynamically

airenzp:
## Description

Now the decimal places are chosen dynamically. If the term type is integer is set to  0, otherwise is set considering that the rounded value is still between the mix and max values.

## Checklist

### DIFF
--- a/client/filter/tvs.density.js
+++ b/client/filter/tvs.density.js
@@ -62,6 +62,7 @@ export function addBrushes(self, new_brush_location) {
 }
 
 function applyBrush(self, elem, brush) {
+	console.log(self)
 	if (!brush.elem) brush.elem = select(elem)
 	const range = brush.range
 	const plot_size = self.num_obj.plot_size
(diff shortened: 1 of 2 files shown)


airenzp:
I need to check this further, still not covering some cases, like for this term  PGS001705... with very small values

client/filter/tvs.density.js
```diff
@@ -128,6 +128,12 @@ function applyBrush(self, elem, brush) {
 
 	function getDecimalPlaces(value, min, max) {
 		let digits = 2
+		if (Math.abs(Math.trunc(value)) == 0) {
```
gavrielm, line 131: Could this be simplified to `if (Math.abs(value) < 1)`?

airenzp: Actually there is no need for Math.abs here, I can remove it.

airenzp: Done

PR #784, opened 2023-10-31
Cross-origin session recovery in a new tab

siosonel:
## Description
Improvements: 
- There is now a menu option to delete cached sessions.
- The open UI now has a table of session names/IDs, the table renderer is also used in the delete option
- Clicking on a newly created `Share` URL link will open a new tab with the embedder's URL, no need to open the proteinpaint host site as an intermediary browser tab 

Test Scenarios for `Open` and `Save`: for each scenario, test that you can save to/open from browser, file, and if applicable, server.

A. Unrpotected ds: comment `serverconfig.dsCredentials.SJLife` (e.g., change to `dsCredentials['#SJLife']`):
  - use [a simple SJLife URL params](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22SJLife%22})
  - there should be no server options in the `Open` and `Save` menus 

B. Protected ds: enable `dsCredentials.SJLife`
  - use http://localhost:3000/example.auth.html
  - there should **be** server options in the `Open` and `Save` menus
  - test while logged in and not
  - when saving to server, you should see the file saved in `[cachedir]/sessionsByCred/...`

C. Same as B, but simulate an embedded use case
  - `sudo vim /etc/hosts` and add the entry `127.0.0.1 viz.localhost`
  - change `dsCredentials['SJLife']` to `dsCredentials['*']` - all embedders will use the same jwt secret
  - from the `sjpp` dir, run `./proteinpaint/server/test/fake-jwt.js SJLife > public/fake-jwt.json '*'`
  - use http://viz.localhost:3000/example.auth.html (different subdomain)
  - manually test like in B
  - Also test that the `Share`-generated URL opens within `viz.localhost`, and not in `localhost`

## Checklist

### DIFF
--- a/client/common/dofetch.js
+++ b/client/common/dofetch.js
@@ -209,8 +209,8 @@ function mayAdjustRequest(url, init) {
 		return url
 	}
 
-	if (method != 'GET') {
-		throw `unsupported init.method='${method}': must be undefined or GET or POST`
+	if (method != 'GET' && method != 'DELETE') {
+		throw `unsupported init.method='${method}': must be undefined or GET or POST or DELETE`
 	}
 
 	if (init.body) {
--- a/client/common/embedder-helpers.js
+++ b/client/common/embedder-helpers.js
@@ -1,0 +1,25 @@
+export function corsMessage(res) {
+	const embedder = res.state?.embedder
+	console.log()
+	const messageListener = event => {
+		console.log(3, event.origin === embedder.origin, event.origin, embedder.origin)
+		if (event.origin !== embedder.origin) return
+		console.log(4, event)
+		if (event.data == 'getActiveMassSession') {
+			window.removeEventListener('message', messageListener)
+			child.postMessage({ state: res.state }, embedder.origin) //; console.log()
+			setTimeout(window.close, 500)
+		}
+	}
+	window.addEventListener('message', messageListener, false)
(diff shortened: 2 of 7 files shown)


karishma-gangwani:
Thanks Edgar. I am testing it. Is this view also supposed to have a back button?
image

xzhou82:
i think it works well. i have only minor comments on the cosmetic styles to improve ui readability
1. the "go back" link at the panel top should use a class e.g. `sja_clbtext2` to indicate they are clickable, compared to the input prompt which are not clickable
2. may not show `Delete` option when there's no session saved in localstorage.
3. in delete menu, use dom/table.js to render the session table

creilly8 (approved):

siosonel:
I pushed a few more fixes, such as the new-tab recovery of a saved embedded browser session in a new tab. Also, new tabs are now target=_blank, so the same tab will not be reused for different sessions.
 
@karishma-gangwani and @xzhou82: I fixed the back button in the share menu, it was unintentionally being cleared. 
@xzhou82: I improved the styles

xzhou82 (approved):
looks great. use the dom/table.js next and can't wait to try it out on viz!

karishma-gangwani (approved):

PR #786, opened 2023-10-31
Allowed to search sample in sampleView

### LINKED ISSUE #772: Add search to filter samples in sample view

airenzp:
## Description

Allowed to search sample in sampleView using an input with autocomplete. Used datalist with limited number of items for autocomplete. Closes #772 

## Checklist

### DIFF
--- a/client/plots/sampleView.js
+++ b/client/plots/sampleView.js
@@ -44,7 +44,7 @@ class SampleView {
 		this.termsByCohort = {}
 
 		const div = this.dom.sampleDiv
-		await this.setSampleSelect(config, div)
+		await this.setSampleSelect(config)
 		const state = this.getState(appState)
 		const q = state.termdbConfig.queries
 		const hasPlots = q ? q.singleSampleGenomeQuantification || q.singleSampleMutation : false
@@ -56,7 +56,6 @@ class SampleView {
 		if (this.dom.header) this.dom.header.html(`Sample View`)
 
 		if (config.samples) {
-			const sampleDiv = this.dom.sampleDiv
 			const sampleLabel = sampleDiv.insert('label').style('vertical-align', 'top').html('Samples:')
 
 			const select = sampleDiv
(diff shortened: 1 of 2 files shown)


xzhou82:
thanks Airen. how is the autocomplete prompt rendered? looks cool
also at [sample view of pnet](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg19%22,%22dslabel%22:%22PNET%22}), when i search by `6767` there's no hit, but SJPNET067670_D1 exists in tsne

xzhou82 (approved):

airenzp:
> thanks Airen. how is the autocomplete prompt rendered? looks cool also at [sample view of pnet](http://localhost:3000/?noheader=1&mass=%7B%22genome%22:%22hg19%22,%22dslabel%22:%22PNET%22%7D), when i search by `6767` there's no hit, but SJPNET067670_D1 exists in tsne

Fixed now Xin. I found there is a datalist tag added for autocomplete. It is relatively new and it does not support very long lists, but filtering it works

airenzp:
Also, datalist has the search integrated, so when the list is not too long it can be used without any workaround, populating it with all the options. I found  for for the dataset [BALL-scrna](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22BALL-scrna%22}) it was breaking

xzhou82:
ok thanks for finding out. the visual style is appealing. if the method
won't scale with big list (and we don't want to load entire big list to
client to search), then this may not be our top choice

airenzp:
It does scale in the way I implemented it. You can test it. Just was explaining by default it just requires to add a datalist referenced by the input.

xzhou82 (approved):

congyu-lu (approved):
works great!

client/plots/sampleView.js
```diff
@@ -112,35 +99,84 @@ class SampleView {
...
+				.append('input')
+				.attr('list', 'sampleDatalist')
+				.attr('placeholder', sampleName)
+				.style('width', '400px')
+				.text(sampleName)
+
+			const datalist = this.dom.sampleDiv.append('datalist').attr('id', 'sampleDatalist')
+
+			input.on('keyup', e => {
+				const limit = 100
```
karishma-gangwani, line 117: Ah, I was wondering why I am only seeing a 100 samples on typing in. It works nicely. I guess my concern would be how would a user know which samples to look for? I typed in 'PA' and only 100 samples show up even though there's more than 100 samples with 'PA'.

karishma-gangwani (dismissed):

client/plots/sampleView.js
airenzp: I increased the limit to 1000, there will be cases for large datasets where this would not list all the samples, but otherwise the rendering breaks.

airenzp: Well I went ahead and set the limit to 10000, so it would list always all the possible samples for all our datasets, except the BALL-scrna, as the datasets have less than this number of samples.

karishma-gangwani (dismissed):
thanks @airenzp

xzhou82:
at the beginning, input box should always show a dropdown arrow, without having to type something first

xzhou82 (approved):

PR #844, opened 2023-11-16
Fix for babel-loader excluding test ts files

creilly8:
## Description
- Revert change in webpack.test.config.js
- New type file for tape tests
- Added typing to hic unit and integration tests

## Checklist


siosonel:
@creilly8 maybe todo later, not urgent: install and use https://www.npmjs.com/package/@types/tape, instead of manually creating types. Or if we don't want to install another dependency, then copy the declared types from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/tape/index.d.ts

creilly8:
> @creilly8 maybe todo later, not urgent: install and use https://www.npmjs.com/package/@types/tape, instead of manually creating types. Or if we don't want to install another dependency, then copy the declared types from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/tape/index.d.ts

Thanks Edgar. This is a better solution to simply install the types. Give me a moment

package.json
```diff
@@ -57,6 +57,7 @@
     "typescript": "^5.1.3"
   },
   "dependencies": {
+    "@types/tape": "^5.6.4",
```
siosonel, line 60: @creilly8 I think this should be a devDependency, otherwise codebases that uses `@sjcrh/proteinpaint-client`, such as the GFF repo, will end up installing tape types even though it will not be used in that codebase.

siosonel (approved):

PR #851, opened 2023-11-17
fix matrix plot get cell from event position issues

congyu-lu:
## Checklist

### DIFF
--- a/client/plots/matrix.dom.js
+++ b/client/plots/matrix.dom.js
@@ -135,4 +135,7 @@ export function setMatrixDom(opts) {
 		this.lastActiveLabel = this.activeLabel
 		delete this.activeLabel
 	}
+
+	window.onscroll = this.scrollStopHandler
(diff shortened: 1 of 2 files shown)


client/plots/matrix.dom.js
```diff
@@ -135,4 +135,7 @@ export function setMatrixDom(opts) {
 		this.lastActiveLabel = this.activeLabel
 		delete this.activeLabel
 	}
+
+	window.onscroll = this.scrollStopHandler
+	this.dom.holder.node().closest('.sjpp-output-sandbox-content').onscroll = this.scrollStopHandler
```
siosonel, line 140: @congyu-lu I just realized that this may break when embedded in GDC, as it doesn't use a `sandbox` div with that classname. I'll convert this PR to draft until we find a way to make it work when embedded.

siosonel (approved):

PR #939, opened 2023-12-08
gdcmaf.rs modification

josephpowi:
## Description
add documentation to the script
print out information when some of required input are missed.

## Checklist

### DIFF
--- a/rust/src/gdcmaf.rs
+++ b/rust/src/gdcmaf.rs
@@ -1,3 +1,15 @@
+/*
+	This script download cohort maf files from GDC, combine them into a single file, and output the sorted file based on chromsome and Start_Position.
+
+	Input JSON:
+		host: GDC host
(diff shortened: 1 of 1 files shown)


robinpaul85:
Maybe use a crate such as [memuse](https://docs.rs/memuse/latest/memuse/) for testing memory usage. Use a json flag when true invokes testing memory usage so that it is useful fir testing, but not used in production.

robinpaul85:
If using memuse, add it as [development dependency](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html) and write specific tests documenting the memory usage.

xzhou82 (approved):

xzhou82:
it's decided no longer need to sort entire file by chr/pos, thus RAM use should not be an issue anymore

PR #981, opened 2023-12-13
Added fillTermWrappers function to do one single request in the PrOFILE and fill the terms

airenzp:
## Description

fillTermWrappers reuses the fillTermWrapper logic doing one single request to fill the terms. This allows to further speed up the plots rendering in the PrOFILE and it can be used later on.

## Checklist


client/termsetting/termsetting.ts
```diff
@@ -877,6 +877,47 @@ export async function fillTermWrapper(tw: TermWrapper, vocabApi: VocabApi, defau
...
+	vocabApi: VocabApi,
+	defaultQByTsHandler?: DefaultQByTsHandler
+) {
+	{
+		const ids: Array<string | undefined> = []
+		for (const tw of tws) ids.push(tw.id)
+		const terms = await vocabApi.getterms(ids)
+
+		for (const tw of tws) {
+			tw.term = terms[tw.id || tw.term.id]
```
siosonel, line 891: @airenzp 
- it's easy to use the wrong function when the only difference is the `s` at the end of one of the function names. Could it be renamed `fillTwLst()`?
- why is the single `fillTermWrapper()` not being called inside here, could this code duplication be avoided?

xzhou82: i agree. is the function returning list of term wrapper objects or term json objects? they are different. if latter, should name it `getTerms()` so it's apparently different than existing `getterm()`

server/routes/termdb.termsbyids.ts
```diff
@@ -0,0 +1,57 @@
+// import { gettermbyidRequest, gettermbyidResponse } from '#shared/types/routes/termdb.termbyid'
```
xzhou82, line 1: please define the request and response types at `server/shared/types/routes/` it's easy, refer to existing examples

client/termsetting/termsetting.ts
airenzp: The functions receives different parameters so I thought it was ok to name it similar, so it is clear what it does. When making fillTermWrappers I tried to minimize changes to avoid breaking changes. It was not so straightforward as calling fillTermWrapper inside a loop.

airenzp: @xzhou82 the function returns a list of term wrappers

siosonel: @airenzp 
- A one letter difference in long function names is confusing, makes IDE auto-completion and find-replace prone to error. Please change.

- having two server routes that do almost the same thing is not ideal, it increases maintenance/testing/support.  Maybe the `fillTermWrapper()` should call the new function and supply it with an array of its own argument? It would be bad if the "filler" logic and expectations start to diverge between these two functions, having to duplicate logic in two place (like create isAtomic) is not good. Maybe you can extract what's common between them?

airenzp: I think I have addressed now all the issue @siosonel. Please check.

airenzp: About unifying endpoints, getterm is used in many places. I am looking at it. This is a very central function, so I am cautious about introducing changes.

airenzp: @siosonel now I have unified the endpoints. Also all the unit and integration tests pass. Please check.

client/termsetting/termsetting.ts
```diff
@@ -874,7 +878,19 @@ export async function fillTermWrapper(tw: TermWrapper, vocabApi: VocabApi, defau
...
 
 	mayValidateQmode(tw)
-	return tw as TermWrapper
+}
+
+export async function fillTwLst(twlst: TermWrapper[], vocabApi: VocabApi, defaultQByTsHandler?: DefaultQByTsHandler) {
+	const ids: Array<string | undefined> = []
+	for (const tw of twlst) ids.push(tw.id)
+	const terms = await vocabApi.getTerms(ids)
+
```
siosonel, line 887: @airenzp In the single `fillTermWrapper()`, there is the code below. Why is that check not in the `fillTwLst()`? Just making sure that necessary validation steps are not removed in the new function.
```js
if (!tw.term) {
		if (tw.id == undefined || tw.id === '') throw 'missing both .id and .term'
		// has .id but no .term, must be a dictionary term
		// as non-dict term must have tw.term{}
		tw.term = await vocabApi.getterm(tw.id)
	}
```

client/termsetting/termsetting.ts
```diff
@@ -874,7 +878,19 @@ export async function fillTermWrapper(tw: TermWrapper, vocabApi: VocabApi, defau
...
+	for (const tw of twlst) {
+		tw.term = terms[tw.id || tw.term?.id]
+		tw.isAtomic = true
+		if (!tw.$id) tw.$id = get$id()
+		await initTermWrapper(tw, vocabApi, defaultQByTsHandler)
```
siosonel, line 892: @airenzp should `Promise.all()` be used here?

client/termsetting/termsetting.ts
airenzp: Yeah, fixed now.

client/termsetting/termsetting.ts
airenzp: Done

client/termdb/TermdbVocab.js
```diff
@@ -477,33 +477,40 @@ export class TermdbVocab extends Vocab {
...
 	}
 
-	async getterm(termid, dslabel = null, genome = null) {
-		if (!termid) throw 'getterm: termid missing'
+	async getTerms(ids, dslabel = null, genome = null) {
+		if (!ids) throw 'getTerms: ids missing'
 		if (this && this.state && this.state.vocab) {
 			if (this.state.vocab.dslabel) dslabel = this.state.vocab.dslabel
 			if (this.state.vocab.genome) genome = this.state.vocab.genome
 		}
```
siosonel, line 485: @airenzp The logic above cancels the overrides if supplied as arguments. This should be
```js
async getTerms(ids: string[], _dslabel = null, _genome = null) {
   if (!ids) throw 'getTerms: ids missing'
  const dslabel = _dslabel || this.state.vocab?.dslabel // this.vocab is guaranteed to exist?
  const genome = _genome || this.state.vocab?.genome // this.vocab is guaranteed to exist?
```

client/termsetting/termsetting.ts
```diff
@@ -874,7 +878,27 @@ export async function fillTermWrapper(tw: TermWrapper, vocabApi: VocabApi, defau
...
+export async function fillTwLst(twlst: TermWrapper[], vocabApi: VocabApi, defaultQByTsHandler?: DefaultQByTsHandler) {
+	const ids: Array<string | undefined> = []
+	const id2term = {}
+	for (const tw of twlst) {
+		if (!tw.term) {
+			if (tw.id == undefined || tw.id === '') throw 'missing both .id and .term'
+			ids.push(tw.id)
+		} else id2term[tw.term.id] = tw.term
+	}
+	const missingTerms = await vocabApi.getTerms(ids)
```
siosonel, line 892: @airenzp should only make a getTerms request if ids is not empty

siosonel:
@airenzp I refactored to clarify the shared code between `fillTwLst()` and `fillTermWrapper()`. I also had to fix tests, but an hic integration test is failing (may be unrelated). I had to add similar methods to FrontendVocab to fix tests. Since `fillTermWrapper()` is called in many places, the extra care in this review is required to make sure nothing breaks later on.

client/termsetting/termsetting.ts
siosonel: Please see my [comment in the commit](https://github.com/stjude/proteinpaint/commit/1542863e44b464489824a5e4b7070f3cdeda651e#comments). I've fixed it in my commit.

client/termsetting/termsetting.ts
```diff
@@ -847,18 +847,46 @@ type DefaultQByTsHandler = {
...
+		if (!terms[id]) throw `missing dictionary term for id=${tw.id}`
+	}
+	return terms
+}
+
+export async function fillTermWrapper(
+	tw: TermWrapper,
+	vocabApi: VocabApi,
+	defaultQByTsHandler?: DefaultQByTsHandler
+): TermWrapper {
```
airenzp, line 881: Here it should return a Promise<TermWrapper>

client/termsetting/termsetting.ts
```diff
@@ -847,18 +847,46 @@ type DefaultQByTsHandler = {
...
 	tw.isAtomic = true
 	if (!tw.$id) tw.$id = get$id()
-
 	if (!tw.term) {
-		if (tw.id == undefined || tw.id === '') throw 'missing both .id and .term'
-		// has .id but no .term, must be a dictionary term
-		// as non-dict term must have tw.term{}
-		tw.term = await vocabApi.getterm(tw.id)
+		const terms = await getDictTerms([tw], vocabApi)
+		tw.term = terms[tw.id]
```
airenzp, line 886: tw.id might be undefined

airenzp (commented):
Thanks for the commits and for the careful review @siosonel. I think we'll benefit of all these changes. I added some minor comments as I was not sure I should commit on top of your changes.

client/termsetting/termsetting.ts
```diff
@@ -847,18 +847,46 @@ type DefaultQByTsHandler = {
 	snplst?: SnpsQ
 }
 
-export async function fillTermWrapper(tw: TermWrapper, vocabApi: VocabApi, defaultQByTsHandler?: DefaultQByTsHandler) {
+export async function fillTwLst(twlst: TwLst, vocabApi: VocabApi, defaultQByTsHandler?: DefaultQByTsHandler): void {
+	const dictTerms = await getDictTerms(twlst, vocabApi)
+	const promises: Promise[] = []
```
airenzp, line 852: Should be `Promise<TermWrapper>`

client/termsetting/termsetting.ts
```diff
@@ -847,18 +847,46 @@ type DefaultQByTsHandler = {
...
+		// non-dictionary tw would always have a tw.term, so only dictionary tw will be tracked here
+		if (!tw.term) {
+			if (tw.id == undefined || tw.id === '') throw 'missing both .id and .term'
+			ids.push(tw.id)
+		}
+	}
+	// ids only have dictionary terms
+	const terms = ids.length ? await vocabApi.getTerms(ids) : {}
+	for (const id of ids) {
+		if (!terms[id]) throw `missing dictionary term for id=${tw.id}`
```
airenzp, line 872: tw does not exist in this line, should be `id=${tw.id}`.

airenzp:
I do not see the changes you mention in the tests, did you forgot to push your last changes?

client/termsetting/termsetting.ts
siosonel: @airenzp could you help fix these type errors and commit? my code editor is not as helpful in finding these

client/termsetting/termsetting.ts
siosonel: @airenzp could you help fix these type errors? my code editor does highlight these for me

siosonel:
> I do not see the changes you mention in the tests @siosonel , did you forgot to push your last changes?

The unrelated failing test that I saw last night passes now, so can ignore my comments related to that.

client/termsetting/termsetting.ts
siosonel: @airenzp could you help fix these type errors? my code editor does highlight these for me

client/termsetting/termsetting.ts
siosonel: @airenzp could you help fix these type errors? my code editor does highlight these for me

client/termsetting/termsetting.ts
airenzp: Done

airenzp:
I fixed the typing and a broken test. Now they all pass. Do you think we could merge this PR now?

xzhou82 (approved):
works thanks

PR #1011, opened 2023-12-18
Passing  filter to getAllSamplesByName to make a single request

airenzp:
## Checklist

### DIFF
--- a/server/src/termdb.js
+++ b/server/src/termdb.js
@@ -17,6 +17,7 @@ import { get_lines_bigfile, mayCopyFromCookie } from './utils'
 import { authApi } from './auth'
 import { getResult as geneSearch } from './gene'
 import { searchSNP } from './app'
+import { get_samples } from './termdb.sql'
 
 /*
 ********************** EXPORTED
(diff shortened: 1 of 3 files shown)


server/src/termdb.js
```diff
@@ -494,15 +493,23 @@ async function get_AllSamplesByName(q, req, res, ds) {
...
-			// {name}
 			result[s.sample] = ds.cohort.termdb.q.sampleName2id(s.sample)
 		}
 		res.send(result)
 		return
-	}
+	} else {
+		let sampleName2Id = ds.sampleName2Id
+		if (q.filter) {
+			q.ds = ds
```
xzhou82, line 503: no need to attach ds to q?

airenzp: it is actually needed, if you comment out that line it fails. It comes from the line 76, 
`// needed by some helper functions
q.ds = ds`

xzhou82 (approved):
fyi by creating a global filter, it doesn't affect the list of samples available in an already-opened sample view. no need to rush a fix though. can think about if to replace preloaded list of samples with search box instead

PR #1012, opened 2023-12-18
getdescsrstats route and sayerror clean up

karishma-gangwani:
## Checklist


client/termdb/TermdbVocab.js
```diff
@@ -474,7 +474,7 @@ export class TermdbVocab extends Vocab {
 		if (filter) {
 			body.filter = getNormalRoot(filter)
 		}
-		return await dofetch3('/termdb', { body })
+		return await dofetch3('/termdb/getdescrstats', { body })
```
xzhou82, line 477: wrong route name. no `get`

creilly8:
@karishma-gangwani Do you have a test link? Or something specific to test?

client/termdb/TermdbVocab.js
karishma-gangwani: okay thanks for pointing. It's fixed now.

siosonel:
@karishma-gangwani @creilly8 I converted to draft because:
- we need the next release this week to GDC to be stable
- there are a lot of changes in this PR that affect many files including `app.js`
- none of the checklists are checked, not clear if unit/integration tests still pass
-

siosonel:
@creilly8 in `dom/sayerror.ts`, as part of the clean up, the try-catch inside `throwMsgWithFilePathAndFnName()` is not needed.

```js
export function throwMsgWithFilePathAndFnName(message: string) {
	try {
		throw new Error() // since there is no other code, should not bother to wrap in a try-catch
		                               // the catch will always be called
	} catch (error: any) {.    // if the catch was needed, then the argument type is for sure `error: Error`, not `any`
		const stackLines = error.stack.split('\n')
		const callerLine = stackLines[2].trim()
```

creilly8:
> @creilly8 in `dom/sayerror.ts`, as part of the clean up, the try-catch inside `throwMsgWithFilePathAndFnName()` is not needed.
> 
> Also, why would parsing the file path/name be needed, when the error stack/trace has that information and someone could just look at that? What's the benefit in using this function?
> 
> ```js
> export function throwMsgWithFilePathAndFnName(message: string) {
> 	try {
> 		throw new Error() // since there is no other code, should not bother to wrap in a try-catch, the catch will always be called
> 	} catch (error: any) {   // if the catch was needed, then the argument type here, as coded, is for sure `error: Error`, not `any`
> 		const stackLines = error.stack.split('\n')
> 		const callerLine = stackLines[2].trim()
> 		...
>        }      
> ```

@siosonel I use this in validation steps when there is no call stack, no processing error to throw. I made this because throughout our code, especially when a value is missing or in the wrong format, the code throws a non-descript statement like "Missing value." That can be really difficult to debug. I use this to throw the call stack in those instances.

siosonel:
> @siosonel I use this in validation steps when there is no call stack, no processing error to throw. I made this because throughout our code, especially when a value is missing or in the wrong format, the code throws a non-descript statement like "Missing value." That can be really difficult to debug. I use this to throw the call stack in those instances.

@creilly8  couldn't it be written this way:
```js
export function throwMsgWithFilePathAndFnName(message: string) {
	throw new Error(message) 
       // or
       const e =  new Error(message)
       console.log(e)
       throw e
```
The function name and code file should not be emitted in the user-viewable message, it seems like the current code will do that. Code/dev details should only be visible in the console or response, not as part of any message that is displayed for the end user.

creilly8:
> > @siosonel I use this in validation steps when there is no call stack, no processing error to throw. I made this because throughout our code, especially when a value is missing or in the wrong format, the code throws a non-descript statement like "Missing value." That can be really difficult to debug. I use this to throw the call stack in those instances.
> 
> @creilly8 couldn't it be written this way:
> 
> ```js
> export function throwMsgWithFilePathAndFnName(message: string) {
> 	throw new Error(message) 
>        // or
>        const e =  new Error(message)
>        console.log(e)
>        throw e
> ```
> 
> The function name and code file should not be emitted in the user-viewable message, it seems like the current code will do that. Code/dev details should only be visible in the console or response, not as part of any message that is displayed for the end user.

@siosonel I'm not sure I quite understand. I use this to enhance throw statements in the browser console log, not as an error message to the user.

creilly8 (approved):

siosonel:
> @siosonel I'm not sure I quite understand. I use this to enhance throw statements in the browser console log, not as an error message to the user.

@creilly8 If it's only in the console, then wouldn't the error trace itself be sufficient? In the following screenshot, the error trace says `at TermdVocab.getCategories (TermdVocab.js:576:49)`, which has both the function name, code filename, and line+column numbers. Why is `[client/termdb/vocabulary-js TermdbVocab. getCategories ()]` needed in the message itself if it's already in the trace? That's why I thought the message is meant for error message display, because in the console you can already see the trace. And genuinely curious, where did the `[client/termdb/vocabulary-js` come from, it's not in the trace?

Screenshot 2023-12-19 at 12 06 28 PM

creilly8:
> > @siosonel I'm not sure I quite understand. I use this to enhance throw statements in the browser console log, not as an error message to the user.
> 
> @creilly8 If it's only in the console, then wouldn't the error trace itself be sufficient? In the following screenshot, the error trace says `at TermdVocab.getCategories (TermdVocab.js:576:49)`, which has both the function name, code filename, and line+column numbers. Why is `[client/termdb/vocabulary-js TermdbVocab. getCategories ()]` needed in the message itself if it's already in the trace? That's why I thought the message is meant for error message display, because in the console you can already see the trace. And genuinely curious, where did the `[client/termdb/vocabulary-js` come from, it's not in the trace? (EDIT: I see it's from `callerLines` which is from the bundled code file, which I guess may be helpful if the code source map is not used.)
> 
> Screenshot 2023-12-19 at 12 06 28 PM

@siosonel Hmmm. That's not what's supposed to happen. The call trace started where the function was called, not within it, and `throwMsg...` shouldn't be included. That's new. Do you mind taking this conversation out of github and on to Slack so I can see what you did better?

PR #1015, opened 2023-12-19
R json output

robinpaul85:
## Description

Output from hclust.R is returned as JSON. fastclust.R has been renamed to hclust.R. 
Please test this branch with high-sample and high-gene GDC example to ensure that the jsonlite package in R is able to stream data back to nodejs for such cases. 

## Checklist


server/src/mds3.load.js
```diff
@@ -479,77 +479,41 @@ async function geneExpressionClustering(data, q, ds) {
...
 	const Rinputfile = path.join(serverconfig.cachedir, Math.random().toString() + '.json')
 	await write_file(Rinputfile, JSON.stringify(inputData))
-	const Routput = await lines2R(path.join(serverconfig.binpath, 'utils/fastclust.R'), [], [Rinputfile])
+	const Routput = JSON.parse(await lines2R(path.join(serverconfig.binpath, 'utils/hclust.R'), [], [Rinputfile]))
 	fs.unlink(Rinputfile, () => {})
-	//const r_output_lines = Routput.trim().split('\n')
-	//console.log('r_output_lines:', r_output_lines)
 
 	let row_coordinates = []
+	for (const item of JSON.parse(Routput['RowNodeJson'])) {
```
xzhou82, line 486: should only JSON.parse() once at line 482. e.g. to get rid of this json.parse, delete `toJSON()` in line 105 hclust.r

server/utils/hclust.R
```diff
@@ -123,11 +142,35 @@ SortedColumnNames <- colnames(normalized_matrix)[ColumnDend$order]
...
+})}
+output_df$SortedColumnNames <- {lapply(1:length(SortedColumnNames), function(y){
+    list(sample=SortedColumnNames[y])
+})}
+output_df$OutputMatrix <- {lapply(1:length(normalized_matrix), function(y){
+    list(elem=normalized_matrix[y])
+})}
+#print ("output_json")
+output_json <- toJSON(output_df)
+print (output_json)
```
gavrielm, line 167: `toJSON()` will output to standard out. No need for the `print()` statement.

josephpowi (commented):
it seems that the 'toJSON' function won't work for streaming data. It might be better to use 'streaamR' package for streaming data to node js if we deal with the big size data.

gavrielm:
What data can be used for testing?

xzhou82:
use non-gdc links in `clustering` section

server/src/mds3.load.js
```diff
@@ -479,77 +479,43 @@ async function geneExpressionClustering(data, q, ds) {
 
 	const Rinputfile = path.join(serverconfig.cachedir, Math.random().toString() + '.json')
 	await write_file(Rinputfile, JSON.stringify(inputData))
-	const Routput = await lines2R(path.join(serverconfig.binpath, 'utils/fastclust.R'), [], [Rinputfile])
+	const Routput = JSON.parse(await lines2R(path.join(serverconfig.binpath, 'utils/hclust.R'), [], [Rinputfile]))
 	fs.unlink(Rinputfile, () => {})
-	//const r_output_lines = Routput.trim().split('\n')
-	//console.log('r_output_lines:', r_output_lines)
 
-	let row_coordinates = []
```
xzhou82, line 487: should not use `let`, use const. the variable is not mutable

robinpaul85:
> it seems that the 'toJSON' function won't work for streaming data. It might be better to use 'streamR' package for streaming data to nodejs if we deal with the big size data.

Yes, that is what I am also a bit afraid. That is why until now, I was refraining from using toJSON()

server/utils/hclust.R
josephpowi: I think you need to remove output_json as well when you comment out the print statement. Just "toJSON(output_df)" will work.

xzhou82 (approved):

PR #1016, opened 2023-12-19
Fixed tvs.samplelst editor

airenzp:
## Description

The tvs.samplelst editor works now, the label reflects the group info and the filter supports negation. I used for label 
`samples in ${group} n=${n}` after discussing it with @siosonel. Otherwise the filter would not make sense.
## Checklist

### DIFF
--- a/client/filter/filter.js
+++ b/client/filter/filter.js
@@ -725,7 +725,7 @@ function setRenderers(self) {
 				const itemCopy = JSON.parse(JSON.stringify(item))
 				const keys = 'keys' in values[index] ? values[index].keys : [values[index].key]
 				itemCopy.tvs.values = keys.map(key => {
-					return { key, label: key }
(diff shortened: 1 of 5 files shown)


xzhou82:
samplelst tvs pill still looks wrong. it should be like this 
image

xzhou82 (approved):
i will merge it now but the tvs pill should be fixed next

airenzp:
Thanks, this PR fixes several issues. About the format, I have to see about how to show it in this format with is. I still think the groups should be specified though, you may have several filters using group variables and if not shown they would look the same, maybe with a different number of samples. Could it be something like 
_SAMPLES IN GROUPX IS N=9_ or 
_GROUPX IS N=9_?

xzhou82:
no worry Edgar is pushing a fix, including broken tvs test

airenzp:
Yes, he just told me. Great

PR #1018, opened 2023-12-19
Better JSON nomenclature about hiercluster dendrogram data from R

robinpaul85:
## Description
Better JSON nomenclature output from hclust.R. Except for the matrix (gene counts), all other variables are first converted into dataframes. In case of gene counts, the older implementation is used because R dataframes require unique column names. 

Sidenote:

The EMT77 example was taking around 24sec. 

## Checklist

### DIFF
--- a/server/src/mds3.load.js
+++ b/server/src/mds3.load.js
@@ -482,30 +482,30 @@ async function geneExpressionClustering(data, q, ds) {
 
 	const row_coordinates = []
 	for (const item of Routput.RowNodeJson) {
-		row_coordinates.push({ x: item['x'], y: item['y'] })
+		row_coordinates.push({ x: item.x, y: item.y })
 	}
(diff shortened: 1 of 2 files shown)


server/src/mds3.load.js
```diff
@@ -486,42 +486,43 @@ async function geneExpressionClustering(data, q, ds) {
 
 	const row_coordinates = []
 	for (const item of Routput.RowNodeJson) {
-		row_coordinates.push({ x: item[0].x[0], y: item[1].y[0] })
+		row_coordinates.push({ x: item['x'], y: item['y'] })
```
xzhou82, line 489: `item.x` is the simpler way to write

robinpaul85: Added a new commit

xzhou82 (approved):

PR #1034, opened 2023-12-21
Hic.4

creilly8:
## Description
- Additional reorganization
- Fix for rendering as 'NONE' by default. Test by comparing https://proteinpaint.stjude.org/?genome=hg38&hicfile=proteinpaint_demo/hg38/hic/hic_demo_v9.hic&enzyme=MboI to http://localhost:3000/?genome=hg38&hicfile=proteinpaint_demo/hg38/hic/hic_demo_v9.hic&enzyme=MboI. Change the normalization method to anything and then change it back to 'INTER_SCALE'. Will add test in future PRs. 

## Checklist

### DIFF
--- a/client/tracks/hic/controls.whole.genome.ts
+++ b/client/tracks/hic/controls.whole.genome.ts
@@ -53,12 +53,12 @@ export function initWholeGenomeControls(hic: any, self: any) {
 		.style('width', '80px')
 		.style('margin-left', '0px')
 		.attr('type', 'number')
-		.property('value', hic.wholegenome.bpmaxv)
+		.property('value', self.wholegenome.bpmaxv)
 		.on('keyup', (event: KeyboardEvent) => {
 			if (event.code != 'Enter') return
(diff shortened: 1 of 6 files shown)


xzhou82 (approved):
neat. thanks
horizontal view should open "in place" and not in floating pane which we should no longer use

creilly8:
> neat. thanks horizontal view should open "in place" and not in floating pane which we should no longer use

@xzhou82 Just so we're clear - The horizontal view should open as a new sandbox above the detailed view? If that's true, should the hicstat app also open in a new sandbox by default? That will give the user the opportunity to collapse and expand both rather than one. 

Also our main user wants to be able to zoom in and out of both sides of the 2chr view independently. The global zoom in/out doesn't work very well in this particular instance. Maybe we can chat about it in the new year?

xzhou82:
to open `horizontal view`, let's just hide the `detailed XY view`. in edit
ui can add a new `XY view` button in that case for user to toggle back

PR #1066, opened 2023-12-30
fix: display total number of mutations on disco plot

aacic:
## Description

Displays the total number snv mutations in the disco plot filter check box label. 

Filters both nonexonic and exonic snv mutations by census genes if only show mutations clicked. 

Deleted MutationType enum as suggested by @siosonel. Now mutation types are imported from "#shared/common" script. 

## Checklist

### DIFF
--- a/client/plots/disco/DiscoRenderer.ts
+++ b/client/plots/disco/DiscoRenderer.ts
@@ -37,8 +37,11 @@ export class DiscoRenderer {
 			controlsDiv,
 			viewModel.settings.label.prioritizeGeneLabelsByGeneSets,
 			viewModel.settings.label.showPrioritizeGeneLabelsByGeneSets,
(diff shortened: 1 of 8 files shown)


xzhou82:
just realized that `mlst[]` has cnv and other data type in addition to snv, but that number count at the CGC gene set filter label should only show snv, since the gene set filter only applies to snv but not cnv and other
thus please change to `mlst.filter(i=>i.dt==common.dtsnvindel).length`

test with http://localhost:3000/?genome=hg38&dslabel=GDC&disco=1&sample=HCM-BROD-0048-C71-01A, should show 61 mutations by excluding cnv

also please change to `import {dtsnvindel} from client/shared/common.js` and do not hardcode dt values

xzhou82 (approved):

PR #1086, opened 2024-01-03
refactor adding termsPerRequest to getAnnotatedSampleData

airenzp:
## Description

Removed getAnnotatedSampleDataSimple and used instead getAnnotatedSampleData that now receives a termsPerRequestParameter. Please check.

## Checklist

### DIFF
--- a/client/termdb/TermdbVocab.js
+++ b/client/termdb/TermdbVocab.js
@@ -649,7 +649,7 @@ export class TermdbVocab extends Vocab {
     .terms[{}]     an array of termWrapper objects
         tw.$id       id to disambugate when multiple terms
                      with the same ID are in terms[], such as in matrix plot
-    	
+    .termsPerRequest When provided, a request includes the number of terms specified, improving the response times
     Returns 
     {
         lst: [{
@@ -735,7 +735,7 @@ export class TermdbVocab extends Vocab {
 			const tws = termsToUpdate.splice(0, termsPerRequest)
 			// request data for one term each time, empty list and break while loop
 			// possible to change to pop two or more each time
-			const copies = tws.map(tw => this.getTwMinCopy(tw))
+			const copies = tws.map(this.getTwMinCopy)
 			const init = {
 				headers,
 				credentials: 'include',
--- a/client/termdb/TermdbVocab.js
+++ b/client/termdb/TermdbVocab.js
@@ -729,12 +729,10 @@ export class TermdbVocab extends Vocab {
 
 		let numResponses = 0
 		if (opts.loadingDiv) opts.loadingDiv.html('Updating data ...')
-		let termsPerRequest = opts.termsPerRequest || 10
+		let termsPerRequest = opts.termsPerRequest || 1
 		// fetch the annotated sample for each term
 		let index = 0
 		while (index < termsToUpdate.length) {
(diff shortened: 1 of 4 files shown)


client/termdb/TermdbVocab.js
```diff
@@ -729,12 +729,14 @@ export class TermdbVocab extends Vocab {
 
 		let numResponses = 0
 		if (opts.loadingDiv) opts.loadingDiv.html('Updating data ...')
+		let termsPerRequest = opts.termsPerRequest || 1
 		// fetch the annotated sample for each term
-		while (termsToUpdate.length) {
+		let index = 0
+		while (index < termsToUpdate.length) {
+			const tws = termsToUpdate.slice(index, index + termsPerRequest)
```
siosonel, line 736: May improve to not require index tracking: 
```js
while (termsToUpdate.length) {
  const tws = termsToUpdate.spiice(0, termsPerRequest)
  ...
}
```

client/termdb/TermdbVocab.js
```diff
@@ -729,12 +729,14 @@ export class TermdbVocab extends Vocab {
...
 			// request data for one term each time, empty list and break while loop
 			// possible to change to pop two or more each time
-			const tw = termsToUpdate.pop()
-			const copy = this.getTwMinCopy(tw)
+			const copies = tws.map(tw => this.getTwMinCopy(tw))
```
siosonel, line 739: can be simplified to below, since `getTwMinCopy()` does not use the `this` keyword in it so can directly use that function as a callback
```js
const copies = tws.map(this.getTwMinCopy)
```

client/termdb/TermdbVocab.js
```diff
@@ -743,64 +745,64 @@ export class TermdbVocab extends Vocab {
...
+								const s = {
+									sample: sampleId,
+									// must reserve _ref -> should not be used as a term.id or tw.$id
+									_ref_
+								}
+								samples[sampleId] = s
 							}
-							samples[sampleId] = s
+							const row = samples[sampleId]
+							if (idn in sample) row[tw.$id] = sample[idn]
```
siosonel, line 788: @airenzp The `idn` may not be unique within the tws/copies, for example if the same term like `agedx` is submitted twice, once as discrete and another as continuous. The `tws/copies` must not submit the same term more than once in the same request, in order to avoid this issue.

siosonel: @airenzp In older implementation of `getAnnotatedSampleData()`, the code detected if the term id or name already exists in the `copies`, and if so, move that term to the end of the tws array so that it will be processed in a subsequent request batch. This way, each request will only have one copy of the same term.

client/termdb/TermdbVocab.js
```diff
@@ -731,7 +731,10 @@ export class TermdbVocab extends Vocab {
 		if (opts.loadingDiv) opts.loadingDiv.html('Updating data ...')
 		let termsPerRequest = opts.termsPerRequest || 1
 		// fetch the annotated sample for each term
+		const pendingTerms = []
 		while (termsToUpdate.length) {
+			const duplicates = termsToUpdate.filter((tw, index) => termsToUpdate.indexOf(tw) !== index)
```
siosonel, line 736: @airenzp the callback will always return false?

xzhou82 (approved):

Issue #1088, opened 2024-01-03
matrix defect: when filtering samples on a discrete numeric term, change the term to continuous breaks


xzhou82:
[test link](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22PNET%22,%22genome%22:%22hg19%22,%22plots%22:[{%22chartType%22:%22matrix%22,%22termgroups%22:[{%22lst%22:[{%22term%22:{%22name%22:%22TP53%22,%22type%22:%22geneVariant%22}},{%22term%22:{%22name%22:%22BCOR%22,%22type%22:%22geneVariant%22}},{%22id%22:%22Gender%22},{%22id%22:%22Age%22},{%22id%22:%22TSNE%20Category%22}]}]}]}), Age term is discrete
first, hide a Age category, then change Age to continuous breaks 

TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at Matrix.getLegendData (matrix.legend.js:97:33)
    at Matrix.getSerieses (matrix.serieses.js:123:25)

siosonel:
Open session file to test: [issue.1088.txt](https://github.com/stjude/proteinpaint/files/13823012/issue.1088.txt)

xzhou82:
yes it fixed the breakage. more issues remain:
1. Hide "<2" from age, turn age to continuous, Age row no longer shows in legend, making it a invisible filter
2. Hide "<2", edit age binning, breaks

PR #1089, opened 2024-01-03
detect if bins exists for a numeric term, before processing its legend data

siosonel:
## Description

Fixes https://github.com/stjude/proteinpaint/issues/1088

## Checklist

### DIFF
--- a/client/plots/matrix.cells.js
+++ b/client/plots/matrix.cells.js
@@ -36,6 +36,8 @@ function setNumericCellProps(cell, tw, anno, value, s, t, self, width, height, d
 		}
 
 		// TODO: may use color scale instead of bars
+		// for bars, use a hardcoded color; TODO: allow a user to customize the bar color?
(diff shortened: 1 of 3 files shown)


siosonel:
Numeric bars are colored, should not be

xzhou82:
(comment moved here from the issue, didn't notice it)
yes it fixed the breakage. more issues remain:

Hide "<2" from age, turn age to continuous, Age row no longer shows in legend, making it a invisible filter
Hide "<2", edit age binning, breaks

xzhou82 (dismissed):
thanks works well
this is like `what you see is what you get` and straightforward. if a numeric term is changed (continuous, changed bin config, or deleted), it's previously hidden bin range will not apply anymore

gdc works too (except a bug found with custom-regular switching)

later on such occasion, can show a message box to inform user about this since the number of map samples are changed and user will wonder why

xzhou82 (approved):

PR #1092, opened 2024-01-04
load premade matrix map by name

xzhou82:
## Description

did this fix to load a premade map by [direct link](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22PNET%22,%22genome%22:%22hg19%22,%22plots%22:[{%22chartType%22:%22matrix%22,%22name%22:%22Matrix%20plot%22}]}), just like scatter plot
to make it easier to debug the actual issue

err handling bug fixed so it will report wrong plot name or lack of support on backend
we will keep the GDC hardcode in this PR and try to tackle it in next branch

## Checklist

### DIFF
--- a/client/plots/matrix.config.js
+++ b/client/plots/matrix.config.js
@@ -120,11 +120,6 @@ export async function getPlotConfig(opts = {}, app) {
 		}
 	}
 
-	if (opts.name) {
-		// should be identifier of premade plot; try to load
-		config = await app.vocabApi.getMatrixByName(opts.name)
-	}
-
 	const s = config.settings
 	const fontsize = Math.max(s.matrix.rowh + s.matrix.rowspace - 3 * s.matrix.rowlabelpad, 12)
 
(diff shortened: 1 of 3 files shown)


xzhou82:
right now it still breaks. not familiar with matrix.config.js could you take a look?

TypeError: Cannot read properties of undefined (reading 'lst')
    at applyLegendValueFilter (matrix.data.js:158:37)
    at Matrix.main (matrix.js:172:27)

airenzp:
I think this may be related to the changes in getAnnotatedSampleData
getData on line 154 of the  termdb.scatter returns this. 
{ samples: {}, refs: { byTermId: {}, bySampleId: {} } }
Maybe the matrix data returned is not right.

airenzp:
Looking a bit deeper, this function is not called, the function involved is `getSampleData`, from the termdb.matrix.js file, that returns an empty list of samples if no term is provided.

client/plots/matrix.config.js
```diff
@@ -120,6 +120,11 @@ export async function getPlotConfig(opts = {}, app) {
 		}
 	}
 
+	if (opts.name) {
```
airenzp, line 123: You may need to do this instead to keep the other configurations

```
 const config2 = await app.vocabApi.getMatrixByName(opts.name)
config = copyMerge(config, config2)
```

airenzp: I tested it and it works

airenzp:
These changes also fixed the PNET matrix. Now the error coming up yesterday is gone.

airenzp (approved):

PR #1096, opened 2024-01-04
detect if bins exists for a numeric term, before processing its legend data (#1089)

akanksha-r:
* detect if bins exists for a numeric term, before processing its legend data

* force a hardcoded bar color for continuous mode term values

* remove any legendValueFilter.lst entries that correspond to an edited tw, to avoid stale/hidden filters being used

## Checklist

### DIFF
--- a/client/tracks/hic/controls.whole.genome.ts
+++ b/client/tracks/hic/controls.whole.genome.ts
@@ -65,9 +65,28 @@ export function initWholeGenomeControls(hic: any, self: any) {
 	addLabel(resolutionRow, 'RESOLUTION')
 	self.dom.controlsDiv.resolution = resolutionRow.append('td').append('span')
 
+	// const matrixTypeRow = menuTable.append('tr')
+	// addLabel(matrixTypeRow, 'matrix type') //Display option is another name? Data type? No label?
+	// self.dom.controlsDiv.matrixType = matrixTypeRow.append('td').text('Observed')
+
+	// Drop down
 	const matrixTypeRow = menuTable.append('tr')
-	addLabel(matrixTypeRow, 'matrix type') //Display option is another name? Data type? No label?
-	self.dom.controlsDiv.matrixType = matrixTypeRow.append('td').text('Observed')
+	addLabel(matrixTypeRow, 'MATRIX TYPE')
+	const matrixTypeDropdownContainer = matrixTypeRow.append('td')
+	const matrixTypeDropdown = matrixTypeDropdownContainer.append('select').on('change', () => {
+		const selectedOption = matrixTypeDropdown.property('value')
+		// Handle the selected option as needed
+		console.log('Selected Matrix Type:', selectedOption)
+	})
+
+	// Options for the dropdown
+	const matrixTypeOptions = ['Observed', 'Expected', 'Observed/Expected']
+
+	// Populate dropdown with options
+	matrixTypeOptions.forEach(option => {
+		matrixTypeDropdown.append('option').attr('value', option.toLowerCase()).text(option)
+	})
+	self.dom.controlsDiv.matrixType = matrixTypeDropdownContainer // Update the reference to the dropdown container
 
 	const viewRow = menuTable.append('tr')
 	addLabel(viewRow, 'VIEW')
--- a/server/src/hicstat.ts
+++ b/server/src/hicstat.ts
@@ -112,7 +112,126 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
 		})
 	//console.log('Reading matrix ...')
 	out_data.normalization = normalization
+
+	// Fetch observed and expected values for all resolutions
+	const resolutions = out_data['Base pair-delimited resolutions'].concat(out_data['Fragment-delimited resolutions'])
+
+	for (const resolution of resolutions) {
+		const observedExpectedData: Record<string, ResolutionData> = {}
+
+		for (const chr1 of out_data.chrorder) {
+			for (const chr2 of out_data.chrorder) {
+				const observed = await getObservedValues(file, resolution, chr1, chr2)
+				const expected = await getExpectedValues(file, resolution, chr1, chr2)
+				const observedOverExpected = calculateObservedOverExpected(observed, expected)
+
+				// Store the data in resolutionsData
+				observedExpectedData[`${chr1}-${chr2}`] = {
+					observed,
+					expected,
+					observedOverExpected
+				}
+			}
+		}
+
+		out_data.resolutionsData[resolution] = observedExpectedData
+	}
+
 	return out_data
+
+	// Functions for o,e, o/e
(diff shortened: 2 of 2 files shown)


server/src/hicstat.ts
```diff
@@ -112,8 +112,127 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
...
+		// Replace the logic below with the actual logic based on your Hi-C file format
+		// The goal is to extract the expected values for the given chromosomes and resolution
+
+		const expectedValues = []
+
+		// This is a placeholder. Replace it with the actual logic to read expected values.
+		for (let i = 0; i < 100; i++) {
+			// Assuming the expected values are stored as 4-byte floats (adjust based on your format)
+			const expected = view.getFloat32(i * 4, true)
+			expectedValues.push(expected)
```
creilly8, line 210: The `Argument of type 'number' is not assignable to parameter of type 'never'.ts(2345)` error on `expected` is because `i` starts at `0`. Check the console log for `const expected = view.getFloat32(i * 4, true)` and see if you always get a result.

server/src/hicstat.ts
```diff
@@ -112,8 +112,127 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
...
+			for (const chr2 of out_data.chrorder) {
+				const observed = await getObservedValues(file, resolution, chr1, chr2)
+				const expected = await getExpectedValues(file, resolution, chr1, chr2)
+				const observedOverExpected = calculateObservedOverExpected(observed, expected)
+
+				// Store the data in resolutionsData
+				observedExpectedData[`${chr1}-${chr2}`] = {
+					observed,
+					expected,
+					observedOverExpected
```
creilly8, line 132: We may need to investigate this approach further. I'm curious if the calculations for the matrix type can be done in the client rather than loading them all at once. If we intend to add more functionality - like all the matrix types listed [the aiden's hic code](https://github.com/aidenlab/Juicebox/blob/master/src/juicebox/windowui/MatrixType.java#L109) - then we may load them asynchronously to avoid expensive processes. This is probably something we ought to discuss to Xin in case it's a moot point.

server/src/hicstat.ts
```diff
@@ -112,8 +112,127 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
...
+		return observedValues
+	}
+
+	async function getMatrixPosition(file: string, resolution: number, chr1: string, chr2: string) {
+		const view = new DataView(isurl ? await readHicUrl(file, 0, 32000) : await readHicFile(file, 0, 32000))
+
+		let position = 0
+		let length = 0
+
+		// Replace the logic below with the actual logic based on your Hi-C file format
```
creilly8, line 174: What does this comment mean? It seems like a placeholder for additional logic.

server/src/hicstat.ts
```diff
@@ -112,8 +112,127 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
...
+		// The goal is to find the position and length of the matrix block for the given chromosomes and resolution
+
+		const magic = getString()
```
creilly8, line 177: Lines 177-185 are already performed at the beginning of `do_hicstat()`. Why is the logic repeated here?

client/tracks/hic/controls.whole.genome.ts
```diff
@@ -65,9 +65,28 @@ export function initWholeGenomeControls(hic: any, self: any) {
...
+	// addLabel(matrixTypeRow, 'matrix type') //Display option is another name? Data type? No label?
+	// self.dom.controlsDiv.matrixType = matrixTypeRow.append('td').text('Observed')
+
+	// Drop down
 	const matrixTypeRow = menuTable.append('tr')
-	addLabel(matrixTypeRow, 'matrix type') //Display option is another name? Data type? No label?
-	self.dom.controlsDiv.matrixType = matrixTypeRow.append('td').text('Observed')
+	addLabel(matrixTypeRow, 'MATRIX TYPE')
+	const matrixTypeDropdownContainer = matrixTypeRow.append('td')
+	const matrixTypeDropdown = matrixTypeDropdownContainer.append('select').on('change', () => {
```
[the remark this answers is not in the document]
creilly8: Also use `self.dom.controlsDiv.matrixType` and not `matrixTypeDropdownContainer`. It's important we use `self ` to access the value of this dropdown later.

server/src/hicstat.ts
```diff
@@ -112,8 +112,127 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
...
+		}
+
+		const version = getInt()
+		if (version !== 7 && version !== 8 && version != 9) {
+			throw Error('Unsupported hic version: ' + version)
+		}
+
+		// Find the position and length of the matrix block for the given chromosomes and resolution
+		// This is a placeholder, and you need to adapt it based on your actual file format
+		position = 1000 // Replace with the actual position based on your format
```
creilly8, line 189: `Replace with the actual position based on your format`.... so are the values for `position` and `length` not correct or universal? Why are they set to 0 then changed to 1000 and 500 respectively without any logic to calculate those numbers?

server/src/hicstat.ts
```diff
@@ -112,8 +112,127 @@ export async function do_hicstat(file: string, isurl: boolean): Promise<HicstatR
...
+		// Read the observed values from the identified matrix block
+		const observedData = isurl
+			? await readHicUrl(file, matrixPosition.position, matrixPosition.length)
+			: await readHicFile(file, matrixPosition.position, matrixPosition.length)
+
+		// Adjust the logic below based on your actual file format
+		const view = new DataView(observedData)
+		let position = 0
+
+		// Example: Reading observed values assuming they are stored as 4-byte floats
```
creilly8, line 158: So if this an example, then it's not applicable to all files. This code will need to account for all file versions, data shapes, etc.

creilly8 (changes requested):
@akanksha-r, I left several comments as a guide on how to finish this PR. The goal for this round is to make sure the returned `outdata` from the server contains the right information needed so we can update the rendering code in the client. Right now, when the user changes the dropdown, nothing happens. There's nothing in `tracks/hic/hic.straw.ts` to rerender or update the svgs. Until that happens, there's no need to show the dropdown. 

It's important to complete and check off all the TODOs in the PR description before marking this PR as ready to review. 

Also, Github always names the PR after the first commit. In this case, the commit message doesn't describe the work that was done and the PR name doesn't describe the purpose. As a result, it will not be easy to search for this work. Please change the name of the PR and be careful to always describe the appropriately in the commit message.  Here's documentation on [how to change the commit message ](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message).

xzhou82:
the PR title is name of this commit
https://github.com/stjude/proteinpaint/commit/ebb4bbe8679a899fd5c532d7a380fa203f164ad2
just start over, copy out modified scripts and create a new branch

PR #1121, opened 2024-01-09
Hic.oe1: Server side code for o,e,o/e options

akanksha-r:
## Description
New variable named oevalues added to the server/routes/hicdata.ts, which can take options of 'observed', 'expected', and 'oe' for running the 'straw' command.  

- It helps to calculate the observed, expected, and oe values for the whole genome as well as for chromosomes.
- By default the option in the straw is 'observed'.

Test it by: (http://localhost:3000/?genome=hg38&hicfile=proteinpaint_demo/hg38/hic/hic_demo_v9.hic&enzyme=MboI) and other demo files.
 

## Checklist

### DIFF
--- a/server/shared/types/routes/hicdata.ts
+++ b/server/shared/types/routes/hicdata.ts
@@ -1,5 +1,5 @@
 export type HicdataRequest = {
-	/** oe values  */
+	/** specify whether observed, expected, or observed/expected(o/e) values.  */
 	oevalues: string
 
 	/** HiC file path from tp/ */
--- a/server/shared/types/routes/hicdata.ts
+++ b/server/shared/types/routes/hicdata.ts
@@ -1,4 +1,7 @@
 export type HicdataRequest = {
+	/** oe values  */
+	oevalues: string
+
 	/** HiC file path from tp/ */
 	file?: string
 	/** Remote HiC file URL */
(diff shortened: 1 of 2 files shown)


server/shared/types/routes/hicdata.ts
```diff
@@ -1,4 +1,7 @@
 export type HicdataRequest = {
+	/** specify whether observed, expected, or observed/expected(o/e) values.  */
+	oevalues: string
```
xzhou82, line 3: please change string into union of allowed values

creilly8 (approved):

server/shared/types/routes/hicdata.ts
```diff
@@ -1,4 +1,7 @@
 export type HicdataRequest = {
+	/** specify whether observed, expected, or observed/expected(o/e) values.  */
```
xzhou82, line 2: use this new description:
Value is the 1st parameter of straw tool, and can only use these specific strings but not anything else

server/shared/types/routes/hicdata.ts
```diff
@@ -1,4 +1,7 @@
 export type HicdataRequest = {
+	/** Value is the 1st parameter of straw tool, and can only use these specific strings but not anything else. */
+	oevalues: 'observed' | 'expected' | 'oe'
```
xzhou82, line 3: here the value is required but client doesn't send it yet. does it throw any err?

creilly8: @xzhou82 I've tested it on a few files and am not seeing an error. @akanksha-r Did you test and see any error?

xzhou82:
will the client fix be added soon? if so fine

creilly8:
I've been working with her on this all morning. I haven't been able to get back to what was doing yesterday. I'm not sure if you got my text this morning about losing power.

I'll get my changes done as fast as possible. I'd really rather not fix a dirty commit history in my branch.
________________________________

xzhou82 (approved):

PR #1135, opened 2024-01-11
hiercluster holderTitle should be determined by dataType of state

xzhou82:
## Description

continue from https://github.com/stjude/proteinpaint/pull/1129
but my changes doesn't work. see if u have good idea

## Checklist


congyu-lu:
the header is set in the init of matrix, so have to set it again in hierCluster after "await super.init(appState)"

congyu-lu (approved):

PR #1160, opened 2024-01-13
detect undefined x1, x2 hits before trying to render a bam variant

siosonel:
## Description
Fixes https://gdc-ctds.atlassian.net/browse/SV-2353.

Instead of a code/read error, the follwing message will show up instead for empty hits for x1/x2.
Screenshot 2024-01-13 at 12 41 34 PM

## Checklist


xzhou82 (approved):
thanks this is a legit fix. forgot that seekcoord() always return an array and if empty, means the coord cannot be mapped to onscreen position. still this shouldn't happen with gdc bam. this can happen if variant either has negative position, falls off the end of chr, or on a different chr entirely. next we need to inspect what the variant data parameter is from a broken view in GFF

PR #1186, opened 2024-01-18
add tests for gene exp clustering

karishma-gangwani:
## Description

I noticed something while writing up the test for gene exp clust.  'divideby' seems to be working on clustering when specified in the config. It adds the title/label 'Case groups by' in the left corner of the variable rows. Would this need to be disabled for hierCluster?

## Checklist


xzhou82:
divideBy is indeed disabled on hiercluster UI. let's don't use it in test for now

xzhou82 (approved):
thanks. let's keep adding to it

karishma-gangwani:
> thanks. let's keep adding to it

sounds good.

PR #1208, opened 2024-01-22
Show value in violin plot

airenzp:
## Description

Allow to open the violin plot from the sample view, displaying a term value. Removes color categories by default.

## Checklist


karishma-gangwani (approved):

xzhou82 (approved):
this works. can you also implement the logic to reduce violin height when number of groups times violin height should not exceed 800px

also termdbtest which uses subcohort cannot launch sample view, you can use it to fix the termdb.filter.js issue and no need to use sjlife

PR #1210, opened 2024-01-22
fix filter creation and group creation error

congyu-lu:
## Description
geneVariant and samplelst TVS don't have term id, so this termjsonByOneid checking will result in filter creation and group creation error

## Checklist

### DIFF
--- a/server/src/termdb.filter.js
+++ b/server/src/termdb.filter.js
@@ -47,7 +47,8 @@ export async function getFilterCTEs(filter, ds, CTEname = 'f') {
 			// .CTEname: str
 		} else if (!item.tvs) {
 			throw `filter item should have a 'tvs' or 'lst' property`
-		} else if (!ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) {
+		} else if (item.tvs.term.id && !ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) {
+			// some tvs such as geneVariant and samplelst don't have term id
 			throw 'invalid term id in tvs'
 		} else if (item.tvs.term.type == 'categorical') {
 			f = get_categorical(item.tvs, CTEname_i)
--- a/client/plots/sampleView.js
+++ b/client/plots/sampleView.js
@@ -1,6 +1,7 @@
 import { getCompInit, copyMerge } from '#rx'
 import { select } from 'd3-selection'
 import { controlsInit } from './controls'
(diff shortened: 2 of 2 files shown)


xzhou82:
Airen spotted some other issue with this code, please help assess

server/src/termdb.filter.js
```diff
@@ -47,7 +47,8 @@ export async function getFilterCTEs(filter, ds, CTEname = 'f') {
 			// .CTEname: str
 		} else if (!item.tvs) {
 			throw `filter item should have a 'tvs' or 'lst' property`
-		} else if (!ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) {
+		} else if (item.tvs.term.id && !ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) {
```
siosonel, line 50: should be
```js
else if (item.tvs.term.id) {
   // some tvs such as geneVariant and samplelst don't have term id
  if !ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) throw 'invalid term id in tvs'			
}
```

congyu-lu: need to put "!ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)" in the condition so that the "else if" below it won't be skipped

congyu-lu:
> Airen spotted some other issue with this code at [http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38-test%22,%22dslabel%22:%22TermdbTest%22}](http://localhost:3000/?noheader=1&mass=%7B%22genome%22:%22hg38-test%22,%22dslabel%22:%22TermdbTest%22%7D), click Sample View button and this code breaks. see if this is related
> 
> ```
> 	thislevelCTEnames.push(f.CTEname)
> 	                         ^
> ```
> 
> TypeError: Cannot read properties of undefined (reading 'CTEname')

This error is not related, it occurred before adding this validation of term id

airenzp:
Yeah I also think it is not related. We can look at it in another PR. For some reason the filtering done to read samples from a subcohort(sjlife or ccss) when accessing the SampleView fails.

congyu-lu:
The error in TermdbTest is caused by the empty tvslst in the filter passed to getAllSampleByName. Should be fixed now, please test again

airenzp (approved):
It works. Thanks!

PR #1211, opened 2024-01-23
Violin fixes

airenzp:
## Description

Restored list samples function. Added ticks/bins input parameter. Added `Use common thickness` parameter.

## Checklist

### DIFF
--- a/client/plots/violin.interactivity.js
+++ b/client/plots/violin.interactivity.js
@@ -58,8 +58,7 @@ export function setInteractivity(self) {
 				}
 			}
 		]
-
-		if (self.config.settings.violin.displaySampleIds && self.state.hasVerifiedToken) {
+		if (self.state.displaySampleIds && self.state.hasVerifiedToken) {
 			options.push({
 				label: `List samples`,
 				callback: async () => {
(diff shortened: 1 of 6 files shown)


airenzp:
I am wondering if this plot is correct. If so I can fix the tests. Please check.

karishma-gangwani:
> I am wondering if this plot is correct. If so I can fix the tests. Please check.

That doesn't look correct Airen.

karishma-gangwani:
> > I am wondering if this plot is correct. If so I can fix the tests. Please check.
> 
> That doesn't look correct Airen. I need to find an example that Xin is talking about and test it in R once.

karishma-gangwani:
> > > I am wondering if this plot is correct. If so I can fix the tests. Please check.
> > 
> > 
> > That doesn't look correct Airen. I need to find an example that Xin is talking about and test it in R once.

In this check that you're performing, I think we shouldn't check the bins but the values themselves before they go into the bins. That's where we can exclude the values from passing into the binbuilder if they're the same in 'plot.values'.

airenzp:
Yeah I noticed the bin object is not an array of numbers, and moved the check to the plot values. It looks like this now

Screenshot 2024-01-23 at 10 23 47 AM

karishma-gangwani:
thanks @airenzp that works but do you think we should move that logic to violin.bins.js because that is a shared file and bins are going to be created there and not on the client side for different plot types specifically?

airenzp:
> thanks @airenzp that works and looks more correct but do you think we should move that logic to violin.bins.js because that is a shared file and bins are going to be created there and not on the client side for different plot types specifically?

Good idea. Done

karishma-gangwani (approved):
thank you! it works.

xzhou82 (approved):

PR #1212, opened 2024-01-23
detect and handle race conditions in the matrix and hiercluster 

siosonel:
## Description

Fixes https://gdc-ctds.atlassian.net/browse/SV-2373.

The issue is due to late-arriving server data requests that could overwrite more recent data requests from the latest state update. The solution is to track the action sequence ID that triggered a state update + server request, and if that ID has changed by the time the server response is complete, then do not process that stale data.

To test:
- http://localhost:3000/testrun.html?dir=plots&name=hierCluster.integration
- http://localhost:3000/testrun.html?dir=mass&name=matrix.integration, may change the last test to `tape.only`

To cause the test to FAIL, comment out line 238 of `rx/index.js`, to verify that the implemented solution works to prevent a simulated race condition. It would not be readily apparent that the matrix has a wrong row label, until the matrix `cell.__data__` is inspected showing the data.value `key` does not match the gene label. This sneaky error is quite bad because nothing looks out of the ordinary.

## Checklist


client/test/testdata/fake-hcdata.js
```diff
@@ -0,0 +1,1974 @@
+// TODO: use typescript to ensure that the hardcoded fake data conforms
```
xzhou82, line 1: we should replace it by including a fpkm gz file to termdbtest, so that `query.geneExpress` will work for termdbtest

xzhou82 (approved):

PR #1259, opened 2024-01-30
fix hierCluster group label being cut off

congyu-lu:
## Checklist


client/plots/matrix.layout.js
```diff
@@ -54,6 +57,24 @@ export function setAutoDimensions(xOffset) {
 	copyMerge(this.settings.matrix, this.computedSettings)
 }
 
+export function getMaxGrpLabelWidth() {
```
siosonel, line 60: @congyu-lu could you add comments on what is being done in this function, like why the labels are attached and removed. Also change the opacity to almost invisible (0.01), we don't want the labels to flash.

siosonel (approved):

PR #1265, opened 2024-01-31
disable unsuitable controls for hierCluster

congyu-lu:
## Checklist

### DIFF
--- a/client/plots/matrix.controls.js
+++ b/client/plots/matrix.controls.js
@@ -399,7 +399,10 @@ export class MatrixControls {
 								width: 50,
 								align: 'center',
 								chartType: 'matrix',
-								inputs: [{ label: 'N/A' }, { settingsKey: 'rowh', min: 8, max: 30, step: 1 }]
+								inputs: [{ label: 'N/A' }, { settingsKey: 'rowh', min: 8, max: 30, step: 1 }],
+								getDisplayStyle(plot) {
+									return plot.chartType == 'hierCluster' ? 'none' : 'table-row'
+								}
 							},
 							{
 								label: 'Min Col. Width',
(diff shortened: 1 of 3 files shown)


client/plots/matrix.controls.js
```diff
@@ -399,7 +399,10 @@ export class MatrixControls {
 								width: 50,
 								align: 'center',
 								chartType: 'matrix',
-								inputs: [{ label: 'N/A' }, { settingsKey: 'rowh', min: 8, max: 30, step: 1 }]
+								inputs: [{ label: 'N/A' }, { settingsKey: 'rowh', min: 8, max: 30, step: 1 }],
+								getDisplayStyle(plot) {
```
siosonel, line 403: @congyu-lu User's may prefer a short row height to be able to view more rows without having to scroll.

client/plots/matrix.controls.js
```diff
@@ -429,7 +432,10 @@ export class MatrixControls {
 								inputs: [
 									{ settingsKey: 'colspace', min: 0, max: 20, step: 1 },
 									{ settingsKey: 'rowspace', min: 0, max: 20, step: 1 }
-								]
+								],
```
siosonel, line 435: @congyu-lu Users may prefer wider gap between heatmap columns, especially in small cohorts. If a heatmap should not have column spacing, then the default value should be 0.

client/plots/matrix.controls.js
congyu-lu: changing row height now will make the left dendrogram misaligned. should we disable it for now before it's fixed?

client/plots/matrix.controls.js
congyu-lu: similar to rowh changing these will make the dendrograms misaligned. should we disable it for now before it's fixed?

xzhou82 (approved):

client/plots/matrix.controls.js
siosonel: @congyu-lu we should try to fix this in a new branch.

PR #1313, opened 2024-02-09
make additional server middleware assumptions on URL param encoding

siosonel:
## Description

This update detects `"true"`, `"false"`, `"undefined"`, and numeric values for json-decoding, even though these URL param values are `string` by default when processed as `req.query` by expressjs. This fixes the mismatch that the server processing has not been fully updated to match the `init.body` conversion to URL params, and also catches typical encoding errors that don't get processes as init.body in `mayAdjustRequest()` in `dofetch.js`.

To test
- http://localhost:3000/testrun.html?dir=shared&name=urljson.unit
- run all unit and integration tests
- all other auto and manual tests that you are familiar with

NOTE: This PR requires lots of testing since it impacts all server routes/handlers.

## Checklist

### DIFF
--- a/server/shared/test/urljson.unit.spec.ts
+++ b/server/shared/test/urljson.unit.spec.ts
@@ -54,7 +54,7 @@ tape('array values', test => {
 	const params: UrlJsonRaw = { a: ['xyz', 1, 0, true, false, null, 'null', undefined, { b: 5 }] }
 	const encodedParams = encode(structuredClone(params))
 	// Special handling of undefined value in an array, which JSON.stringify() converts to null
-	params.a[params.a.indexOf(undefined)] = null
+	if (params.a && Array.isArray(params.a)) params.a[params.a.indexOf(undefined)] = null
 	test.deepEqual(params, decode(preprocess(encodedParams)), 'should correctly encode and decode array values')
 	test.end()
 })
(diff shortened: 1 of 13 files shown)


airenzp:
Thanks for working on this. It works for me the boolean case, I tested it [here](https://github.com/stjude/proteinpaint/commit/16cd5d234e80e27f4cabdbb7ee53a53cf943b920) passing a boolean to change the method in the violin

server/shared/test/urljson.unit.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+const structuredClone = /*window?.structuredClone ||*/ p => JSON.parse(JSON.stringify(p))
+
+/**************
+ test sections
+***************/
+
+// Test that the original values are recovered with the expected type
+
+tape('string values', test => {
+	const params: UrlJsonRaw = { a: 'abc', b: '123', c: '{}', d: '[]', e: 'false', f: 'true', g: 'null', h: 'undefined' }
```
xzhou82, line 22: can add string value bound by pair of quotes e.g. `"xx"` or `'xx'`

xzhou82 (approved):

xzhou82:
thank you so much! in the following we will fix below as we see them
- client side manual urlparam encoding
- backend parameter type check, json.parse and Number() casting, all those are unnecessary now

PR #1369, opened 2024-02-24
prototype fetch2() that wraps native fetch to replace got

xzhou82:
## Description

fetch2 is used in two places. test [this](http://localhost:3000/?genome=hg38&dslabel=GDC&disco=1&sample=TCGA-85-7710-01A), and [this](http://localhost:3000/?genome=hg38&gene=ENST00000407796&mds3=GDC) by clicking any dot. both works
if looks good can proceed to replace all in this file first

## Checklist

### DIFF
--- a/server/src/mds3.gdc.js
+++ b/server/src/mds3.gdc.js
@@ -62,6 +62,25 @@ export function convertSampleId_addGetter(tdb, ds) {
 	}
 }
 
+export function fetch2(url, opts) {
+	// wrap fetch common steps; throw gdc-tailored message on http error status
+	return fetch(url, opts)
+		.then(response => {
+			if (response.status > 399) {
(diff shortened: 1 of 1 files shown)


siosonel:
@xzhou82 Instead of us coding a wrapper for native fetch, we should see if there are popular nodejs http client packages, I can look into it. For example, https://github.com/sindresorhus/ky. It's from the unmaintained `got` developers and it uses native fetch, much smaller, from that link:

It's just a tiny file with no dependencies.

Benefits over plain fetch
Simpler API
Method shortcuts (ky.post())
Treats non-2xx status codes as errors (after redirects)
Retries failed requests
JSON option
Timeout support
URL prefix option
Instances with custom defaults
Hooks

PR #1392, opened 2024-02-29
tentative fix on bin computing; rendering is broken

xzhou82:
## Description

need help to evaluate if updated bin computing is okay, also any fix on rendering based on revised bin data

## Checklist


xzhou82:
kde uses a default bandwidth of 5 and causes issue for data with tight range (see Gav's example). how about using a dynamic bandwidth (max-min)/20

airenzp:
The bandwidth is not related with the data values. See here https://observablehq.com/@d3/kernel-density-estimation

airenzp:
The number of bins, which is what you mean, is being calculated that way

airenzp:
When looking at this [plot](http://localhost:3000/?noheader=1&mass=%7B%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22id%22:%22PGS000001_ceu_maf%22,%22q%22:%7B%22mode%22:%22continuous%22%7D%7D%7D%5D,%22nav%22:%7B%22activeTab%22:1%7D%7D) suggested by  @gavrielm I found out that the KDE can have negative probabilities, so I had to fix the code accordingly. Should be fixed now.

airenzp (approved):

PR #1401, opened 2024-03-01
fix issue 1367

### LINKED ISSUE #1367: matrix cell tooltip behavior change
moving cursor across multiple matrix cells has a nasty effect of fuzzy tooltip, because it's busy fading in and out

do not show the tip on mouseover cell and hide as soon as mouseout a cell
instead, show the tooltip when mouseover entire matrix body, and only hide when mouseout the body; when moving across cells, reposition tip and update content, do not show/hide! that will get rid of nasty effect

congyu-lu:
## Description
fix #1367

## Checklist


xzhou82 (approved):
works nicely

PR #1435, opened 2024-03-13
Add blat server image

aacic:
## Checklist


container/blat/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM debian:bullseye-slim
```
siosonel, line 1: I think we should use `bookworm-slim`, just like we do in `container/deps/Dockerfile`, we may be able to reuse a cached image layer. Also, `bullseye` is old stable, `bookworm` is current stable.

container/blat/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+RUN wget https://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/blat/gfServer \
+    && chmod +x gfServer
+
+RUN mkdir -p ./genomes
+
+EXPOSE $PORT
+
+CMD ["sh", "-c", "$COMMAND"]
```
siosonel, line 25: the variable "$COMMAND" here seems too permissive, may want to restrict this a bit. For example, 
```dockerfile
CMD ["sh", "-c", "./gfServer start $HOST $PORT -stepSize=$STEPSIZE /home/root/blat/genomes/$HGNUM.2bit"]
```

container/blat/Dockerfile
aacic: Changed.

container/blat/Dockerfile
aacic: I did like this:

`CMD ["sh", "-c", "./gfServer start $HOST $PORT -stepSize=$STEPSIZE /home/root/blat/genomes/$HGNUM.2bit"]`

siosonel (approved):

container/blat/Dockerfile
siosonel: @aacic I think `$HOST` will always be `127.0.0.1` within the container? If yes, should hard code this instead of being an argument. But I approved the PR either way.

aacic: Ok. It's hardcoded and merged.

PR #1445, opened 2024-03-18
Sample parent

airenzp:
## Description

This PR allows to represent relation between samples. It displays related samples in the SampleView. It can be tested with this [db](https://drive.google.com/file/d/14wSk6fdJl2pq2xUsa0QPISNKai512FWu/view?usp=sharing) from pnet. Only the databases with ancestry information should be regenerated, using  `node ~/dev/sjpp/utils/termdb/buildTermdb.bundle.js terms=terms.json annotation3Col=annotations.txt sampleAncestry=sample_ancestry.tsv dbfile=db
`. PNET was already regenerated using this file copied to hpc `hpc:tp/files/hg19/pnet/clinical/sample_ancestry.tsv`

## Checklist


siosonel:
@airenzp I think this should require a corresponding PR branch in the `sjpp` repo? For example, in `sjpp/termdb/create.sql`, to show how the change to the table schema.

airenzp:
> @airenzp I think this should require a corresponding PR branch in the `sjpp` repo? For example, in `sjpp/termdb/create.sql`, to show the change to table schema(s).

Yes, Edgar you are right. I just made a PR. I am also going to work on some changes discussed with Xin, so I will mark this PR as a draft. We have to consider the possibility of child samples that are not related between each other, but have the same parent.

utils/termdb/buildTermdb.js
```diff
@@ -245,10 +245,7 @@ function loadTermsFile(scriptArg) {
 	 */
 	const terms = []
 	const parent_ids = new Set() //To store parent_id for each term in terms
-	for (const line of fs
```
xzhou82, line 248: please delete this file and use sjpp/utils/termdb/buildTermdb.js instead. this one was accidentally left here before repo switch. sjpp version has far more features

airenzp: There is only this pp file Xin, in sjpp a bundle version is generated, but I could not find the duplicate you mentioned.

airenzp: Then I am trying to move it to sjpp/utils/termdb and compile it to test the changes done but I need to update the import paths. I am having compilation errors.

airenzp: What I see now is this error `	initBinConfig is not a function` when running `node ~/dev/sjpp/utils/termdb/buildTermdb.bundle.js terms=terms.json annotation3Col=annotations.txt dbfile=db
sampleAncestry=sample_ancestry.tsv`

siosonel: @airenzp Could you console.log `initBinConfig`? 

Also, may be unrelated but good to update since it's needed anyway: in `sjpp/utils/termdb/webpack.config.js` line 33, set the `targets.node: 20` instead of `12`.

Screenshot 2024-03-19 at 9 20 43 PM

airenzp: I could fix the initBinConfig error, and updated now the node version, thanks.

xzhou82 (approved):

server/src/termdb.server.init.js
```diff
@@ -61,6 +61,8 @@ export function server_init_db_queries(ds) {
 		'anno_categorical',
 		'buildDate'
 	]
+	if (tables.has('sample_ancestry') && cn.prepare('SELECT * FROM sample_ancestry').all().length > 0)
```
siosonel, line 64: @airenzp Should eventually improve this to `SELECT count(*) as num_rows FROM sample_ancestry`, which would return only one row versus returning potentially thousands of rows.

PR #1477, opened 2024-03-29
Handle custom matrix geneset input option

siosonel:
## Description

This features addresses https://gdc-ctds.atlassian.net/browse/FEAT-766. 
```text
Upload User custom gene sets:

Similar to v1, GDC 2.0 allows users to create gene sets using the Manage Sets feature. 

See https://docs.gdc.cancer.gov/Data_Portal/Users_Guide/getting_started/#manage-sets 

In OncoMatrix and Gene Expression Clustering tools, under Genes button, 

Under Edit Genes & Create Genes, add “Load Gene Sets” option.

If the user has already created a gene set using the “Manage Sets” feature, then similar to MSigDB, display the gene sets
```

To test with `example.gdc.*.html`:
- pull sjpp/master
- open http://localhost:3000/example.gdc.matrix.html?maxGenes=3&cohort=CDDP_EAGLE-1
- open `Genes -> Gene Sets: Edit Group -> Load Gene Set` 
- Enter `KRAS, MYCN` or other gene list in the popup modal and enter
- the gene text input div should be filled-in, click Submit
- the matrix should update with the new gene set
- can repeast the test using `example.gdc.exp.html`

To test in GFF (@congyu-lu )
- open OncoMatrix or Gene Expression Clustering tool
- open `Genes -> Gene Sets: Edit Group -> Load Gene Set` 
- click on the first tab to enter comma-separated gene list then submit, or click on second tab to `USE A PREVIOUSLY SAVED GENE SET`
- the OncoMatrix or GEC should update directly. Pre-filling the gene input div will require the geneset edit UI to be reactive (updatable by external code), will require more work in a separate PR

## Checklist

### DIFF
--- a/client/dom/GeneSetEdit/GeneSetEditUI.ts
+++ b/client/dom/GeneSetEdit/GeneSetEditUI.ts
@@ -54,6 +54,11 @@ export type CallbackArg = {
 	geneList: Gene[]
 }
 
+type CustomInputs = {
+	label: string
+	showInput: (arg: any) => void
+}[]
+
 export type GeneSetEditArg = {
 	holder: Elem
 	genome: ClientCopyGenome
@@ -71,6 +76,7 @@ export type GeneSetEditArg = {
 	geneList?: Gene[]
 	/** Title appearing above the UI */
 	titleText?: string
+	customInputs?: CustomInputs
 }
 
 export class GeneSetEditUI {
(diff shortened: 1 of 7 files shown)


xzhou82:
thanks a lot. this new button `Load Gene sets` should be put into geneset edit UI instead, rather than next to two existign buttons (edit current group, create new group). this will allow directing the newly loaded geneset into a specific group, rather than unable to direct it as it is now

xzhou82:
not urgent. can wait after apr release

creilly8 (commented):
I typed `123` in to the modal and it appeared as a gene in the geneHolder. I see on submit the `unknown gene` error pops up and greys out the matrix. Wouldn't be better to confirm if it's a real gene before leaving the UI? Or is that the standard error message behavior for the matrix?

creilly8:
> > I typed `123` in to the modal and it appeared as a gene in the geneHolder. I see on submit the `unknown gene` error pops up and greys out the matrix. Wouldn't be better to confirm if it's a real gene before leaving the UI? Or is that the standard error message behavior for the matrix?
> 
> Ideally, the gene entry validation will be done in the geneset edit UI, either as a helper function that can be called within the customInputs menu.callback, or as part of `this.render()`. The customInputs is expected to return a `geneList`. In the GDC it's assumed to be from a saved user gene set and so it would have been validated when the user entered/saved the gene set, the list is guaranteed to have validated gene symbols. For other portals/example html, it will require a case-by-case assessment of whether gene symbol validation is required, where the helper function/validation in genesetEditUI.render() will be used.

I see. Since it's for the GDC, I don't think we need to worry about it.

creilly8 (dismissed):

xzhou82 (approved):

PR #1512, opened 2024-04-09
Term search

airenzp:
## Description

This PR adds the TermTypeSearch component. Now everytime the tree is loaded the tabs may be shown if needed. Once a tab is selected a corresponding view is shown to select a term. The search input was customized to only search for items of a type. On each use case the component should allow the expected tabs. I worked in this together with @siosonel. Please check.

## Checklist


gavrielm:
I know this is still a draft PR, but I noticed that `SNP List` and `SNP Locus` both error out in the [sjlife portal dictionary](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22SJLife%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22dictionary%22}]}) when I try to submit an input.

Also, the UIs for SNP List and SNP Locus are written for regression analysis. They will likely need to be modified for rendering in data dictionary.

xzhou82:
can keep snplst and snplocus as-is and just make them to work for regression and download

karishma-gangwani:
Will the omni search box have specific default terms based on the data type in the future? Now it says 'search term' for both.

gavrielm:
> I just pushed a fix for the dictionary use case, please check again.

Great thanks. In regression, this intermediate menu should be removed:

image

karishma-gangwani (approved):
It does look nice with the terms segregated according to data types. Should we just label 'Data Variables' as 'Variables' since the button does say `Data dictionary' ? Just a thought.

gavrielm (dismissed):
Tested in sjlife, mbmeta, and pnet and all work well

creilly8:
This is minor. If I search for something in one tab, do not hit enter, then switch to another tab, the search results remain: 
Screenshot 2024-04-11 at 2 19 12 PM

Also minor: I noticed the input must be at least 3 characters before a search result appears. That works well for the dictionary terms because they're long, but what about the mutations, genes, etc? They're really short.

creilly8:
@airenzp After looking at your PR, it occurs to me we never include a `No Result` message for non-matches for any of our inputs. We probably do that going forward. What do you think?

airenzp:
> It does look nice with the terms segregated according to data types. Should we just label 'Data Variables' as 'Variables' since the button does say `Data dictionary' ? Just a thought.

@karishma-gangwani  this component is shown everywhere there is a tree with more than one choice, not just under the mass UI dictionary.

airenzp:
> I have two questions.
> 
> 1. Will the omni search box have specific default terms based on the data type in the future? Now it says 'search term' for both.
> 2. Would this be also supported for overlay and divide by later?

Thanks for the review.  Now I show a different placeholder when changing tabs. And it is customized now to show geneVariant on the overlay and divide by, if you find something that before was possible missing let me know!

airenzp:
> This is minor. If I search for something in one tab, do not hit enter, then switch to another tab, the search results remain: Screenshot 2024-04-11 at 2 19 12 PM
> 
> Also minor: I noticed the input must be at least 3 characters before a search result appears. That works well for the dictionary terms because they're long, but what about the mutations, genes, etc? They're really short.

Thanks @creilly8 . now when changing tabs the search is cleared. Thanks for pointing this out. I also noticed we need to type 3 characters to find a gene. @xzhou82 should we ask for less? like 2 chars instead?

karishma-gangwani:
> > I have two questions.
> > 
> > 1. Will the omni search box have specific default terms based on the data type in the future? Now it says 'search term' for both.
> > 2. Would this be also supported for overlay and divide by later?
> 
> Thanks for the review. Now I show a different placeholder when changing tabs. And it is customized now to show geneVariant on the overlay and divide by, if you find something that before was possible missing let me know!

works great! Thanks Airen

xzhou82:
there's no `omnisearch`. dictionary search will only turn up dict terms

xzhou82:
gene search should allow single letter (gene `T`). Airen can search.js accomondate that?

xzhou82:
for searching genevariant/geneexp/etc, best to just use `dom/genesearch.js`

server/src/searchSNP.js
```diff
@@ -14,6 +14,9 @@
 .lst[ str ]
 	used for byName
 */
+
+import { query_bigbed_by_coord } from './utils.js'
+
 export async function searchSNP(q, genome) {
```
xzhou82, line 20: looks like this func is now exported by routes/snp.ts. may dedup

airenzp: Done

xzhou82:
snp search works after dedup. thanks. please fix minor issues:
- at [violin](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22hrtavg%22,%22q%22:{%22mode%22:%22continuous%22}}}],%22nav%22:{%22activeTab%22:1}}), open Overlay breaks
- snplst and snplocus tabs are still shown, e.g. [sjlife cuminc overlay](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22cuminc%22,%22settings%22:{%22cuminc%22:{}},%22term%22:{%22id%22:%22Cardiovascular%20System%22,%22q%22:{%22breaks%22:[1]}}}],%22nav%22:{%22activeTab%22:1}}), and sjlife download (select Dictionary variables first then see the tabs). please hardcode in termTypeSearch to hide them for now
-

airenzp:
> snp search works after dedup. thanks. please fix minor breaking issues:
> 
> * at [violin](http://localhost:3000/?noheader=1&mass=%7B%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22id%22:%22hrtavg%22,%22q%22:%7B%22mode%22:%22continuous%22%7D%7D%7D%5D,%22nav%22:%7B%22activeTab%22:1%7D%7D), open Overlay breaks
> 
> i will create a new issue with non-breaking things for you to fix in next pr. thanks

Thanks. I added the violin use case now.

xzhou82 (approved):

PR #1528, opened 2024-04-15
Added overlay/divide case

airenzp:
## Description

Overlay and divide can allow dictionary terms or geneVariant terms as they have categories. By default this use case can be specified in the termsetting. When a use case is not set, this case is set.

## Checklist


client/termsetting/termsetting.ts
```diff
@@ -550,15 +550,14 @@ function setInteractivity(self) {
 				holder instanceof Element ? holder : this instanceof Element ? this : self.dom.holder.node()
 			)
 		else self.dom.tip.show(event!.clientX, event!.clientY)
-
+		if (!self.usecase) self.usecase = { target: 'overlayOrDivide' }
```
xzhou82, line 553: may just call it `default`

xzhou82 (approved):

PR #1533, opened 2024-04-17
replace got with ky in topmgene route

xzhou82:
## Description

[local works](http://localhost:3000/example.gdc.matrix.html?cohort=Gliomas). please test with localGFF+logging in+oncomatrix which uses the wrong api url, to see if it can catch the error?

## Checklist


xzhou82:
we found that ky (and got) is able to throw `404` error at requesting topMgenes (caused by /v0/v0/graphql bug from ds file). but client/plots/geneset.js only console log this err and does not indicate it on ui, thus it was super confusing
@siosonel

congyu-lu (approved):

PR #1549, opened 2024-04-23
fix: GDC bam slicing ui show actual number of available bam files whi…

xzhou82:
…ch can be lower than 1000, fix scrollbar appearance

## Description

please pull sjpp master first
at http://localhost:3000/example.gdc.bam.html, due to cohort filter use, it shows `68 available bams` instead of 1000 as before which is wrong
please help me test it in localGFF to see when applying cohort filter (has to be a very restrictive filter), the number of bams can go down below 1000

nothing else should break 

## Checklist


congyu-lu (approved):
348 Available BAM Files with APOLLO

PR #1593, opened 2024-05-09
fix conflicting type def for getCategories() request, simplify def

xzhou82:
## Checklist

### DIFF
--- a/server/routes/termdb.categories.ts
+++ b/server/routes/termdb.categories.ts
@@ -82,16 +82,9 @@ function init({ genomes }) {
 }
 
 async function trigger_getcategories(
-	q: {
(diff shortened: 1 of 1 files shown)


server/routes/termdb.categories.ts
```diff
@@ -82,16 +82,9 @@ function init({ genomes }) {
 }
 
 async function trigger_getcategories(
-	q: {
```
xzhou82, line 85: line 65 `const q = req.query as getcategoriesRequest` already defines q type thus no need redefine here

karishma-gangwani (approved):

airenzp (approved):

PR #1602, opened 2024-05-10
Gene exp filter

airenzp:
## Description

This PR adds to support to filter  by gene expression.

## Checklist

### DIFF
--- a/server/src/termdb.filter.js
+++ b/server/src/termdb.filter.js
@@ -48,11 +48,9 @@ export async function getFilterCTEs(filter, ds, CTEname = 'f') {
 			// .CTEname: str
 		} else if (!item.tvs) {
 			throw `filter item should have a 'tvs' or 'lst' property`
-		} else if (
-			item.tvs.term.id &&
-			(!item.tvs.term.type || !isNonDictionaryType(item.tvs.term.type)) &&
-			!ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)
-		) {
+		} else if (isNonDictionaryType(item.tvs.term.type)) {
+			f = { CTEname: CTEname_i, CTEs: [`${CTEname_i} AS (select id as sample from sampleidmap)`], values: [] }
+		} else if (item.tvs.term.id && !ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) {
 			throw 'invalid term id in tvs'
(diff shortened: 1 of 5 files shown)


server/src/termdb.filter.js
```diff
@@ -48,11 +48,9 @@ export async function getFilterCTEs(filter, ds, CTEname = 'f') {
...
 		} else if (!item.tvs) {
 			throw `filter item should have a 'tvs' or 'lst' property`
-		} else if (
-			item.tvs.term.id &&
-			(!item.tvs.term.type || !isNonDictionaryType(item.tvs.term.type)) &&
-			!ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)
-		) {
+		} else if (isNonDictionaryType(item.tvs.term.type)) {
+			f = { CTEname: CTEname_i, CTEs: [`${CTEname_i} AS (select id as sample from sampleidmap)`], values: [] }
+		} else if (item.tvs.term.id && !ds.cohort.termdb.q.termjsonByOneid(item.tvs.term.id)) {
```
xzhou82, line 53: i suppose it should be extended after line 69 to plug in geneExp type

airenzp: This is just a fake filter handling so it does not break this code. Filtering by gene expression does not query the db. The filtering is being done in the getData.

server/src/termdb.matrix.js
```diff
@@ -178,7 +186,14 @@ async function getSampleData(q) {
 				const values = data.gene2sample2value.get(tw.term.gene)
 				const value = Number(values[sampleId])
 				let key = value
-				if (tw.q?.mode == 'discrete') key = getBinLabel(tw.q, lst, value)
+				if (ranges.length) {
+					const filterBin = getBin(ranges, value)
+					if (filterBin == -1) continue
```
airenzp, line 191: See here, if the value does not fall in the filter range the sample it is not added

airenzp:
Moving forward with this logic would mean that other methods that query the db, like get_samples, would use getData instead so they get the filtered samples for non dict terms. It is another major change that would affect a lot of code.

xzhou82 (approved):
thanks works, next:
1. cover geneExp in filter integration test
2. at filter UI, should indicate term type under a geneExp/geneVariant pill

PR #1609, opened 2024-05-14
Moved term constants and helper functions to shared/terms

airenzp:
## Description

Moved term type enumerators, mappings and helper functions to shared/terms. Updated Imports. Fixed handler imports using #shared. Used constants in typeGroups definition. I also added added the groupTypes definition in case we want to add it, if not we can remove it. Please check.

## Checklist


gavrielm:
Thanks for the refactoring. All changes seem to work well.

But I do not think the `groupTypes` definition is worth adding right now. If we need to add a new term type, we would need to add it to `TermTypes`, `typeGroup`, `groupTypes`, and `nonDictTypes` in `client/shared/terms.js`. That can be difficult to maintain. Ideally, we could have a single object to store all term types and then work off of that object. Not sure if that is feasible right now, but something to possibly work towards.

xzhou82 (approved):
all CI pass
in termdb.matrix.js, `sampleFilterSet` is still used by geneExp but it should not

airenzp:
> all CI pass in termdb.matrix.js, `sampleFilterSet` is still used by geneExp but it should not

fixed now

server/src/termdb.matrix.js
```diff
@@ -110,7 +110,8 @@ async function getSampleData(q) {
 		return { samples, refs: { byTermId, bySampleId: {} } }
 	}
 
-	const sampleFilterSet = await mayGetSampleFilterSet(q, nonDictTerms) // conditionally returns a set of sample ids
+	const sampleFilterSet = await mayGetSampleFilterSet4snplst(q, nonDictTerms) // conditionally returns a set of sample ids, FIXME *only* for snplst and snplocus data download in supported ds, not for anything else. TODO remove this bad quick fix
```
airenzp, line 113: can we move this line to the if case handling snp cases then?

xzhou82:
then geneExp query breaks

airenzp:
please send the example

airenzp:
Moved mayGetSampleFilterSet4snplst to snp cases

gavrielm (approved):

PR #1619, opened 2024-05-20
DE filter

robinpaul85:
## Description

In this PR, now min_count and min_total_count can now be controlled through the UI. min_count is used to calculate the cpm cutoff and min_total_count is an absolute cutoff. For more details, see the filterByExpr [source code ](https://rdrr.io/bioc/edgeR/src/R/filterByExpr.R)in edgeR.

For testing, may use this sessions file
[sjmb12_DE.txt](https://github.com/stjude/proteinpaint/files/15381858/sjmb12_DE.txt)

## Checklist

### DIFF
--- a/client/plots/DEanalysis.js
+++ b/client/plots/DEanalysis.js
@@ -54,6 +54,24 @@ class DEanalysis {
 		this.dom.controlsDiv.selectAll('*').remove()
 		const inputs = [
 			{
+				label: 'Min count',
+				type: 'number',
+				chartType: 'DEanalysis',
+				settingsKey: 'min_count',
+				title: 'Min count',
+				min: 0,
+				max: 100
+			},
+			{
+				label: 'Min total count',
+				type: 'number',
+				chartType: 'DEanalysis',
+				settingsKey: 'min_total_count',
+				title: 'Min total count',
+				min: 0,
+				max: 100
+			},
+			{
 				label: 'P-value significance (linear scale)',
 				type: 'number',
 				chartType: 'DEanalysis',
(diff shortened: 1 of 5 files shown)


client/plots/DEanalysis.js
```diff
@@ -53,6 +53,24 @@ class DEanalysis {
 	async setControls(output) {
 		this.dom.controlsDiv.selectAll('*').remove()
 		const inputs = [
+			{
+				label: 'Min count',
+				type: 'number',
+				chartType: 'DEanalysis',
+				settingsKey: 'min_count',
+				title: 'Min count',
```
xzhou82, line 61: at title, fully specify the purpose "Genes with xxx will be excluded"

robinpaul85: I added a title to both min_count and min_total_count.

client/plots/DEanalysis.js
```diff
@@ -53,6 +53,25 @@ class DEanalysis {
 	async setControls(output) {
 		this.dom.controlsDiv.selectAll('*').remove()
 		const inputs = [
+			{
+				label: 'Min count',
```
xzhou82, line 57: for these two new params, will it be better to call them `Min read count per sample` and `Min total read count`
at a minimum, if the cutoff is about read count, then label must indicate so

xzhou82:
works well.
later should investigate that on UI, after changing cutoff parameter, the app should show a cloak and indicate to user it's loading which takes seconds

client/plots/DEanalysis.js
robinpaul85: Discussed privately in slack

xzhou82 (approved):
you should actually make `label` more intuitive. at minimum mention `read count` otherwise it's not user friendly

xzhou82 (approved):

robinpaul85:
> you should actually make `label` more intuitive. at minimum mention `read count` otherwise it's not user friendly

Added `read count` to label

PR #1621, opened 2024-05-21
Enable new type cases

airenzp:
## Description

This PR adds support to geneExpression and metaboliteIntensity in all the use cases allowed. Now we list the use cases excluded in TermTypeSearch, instead of included cases. Also we disable terms in the dictionary passing the term and not the id, so we can disable non dict terms. As there are many use cases and files affected, please check if the tree looks as expected in the different use cases and check selecting terms. 

## Checklist

### DIFF
--- a/client/termdb/TermTypeSearch.ts
+++ b/client/termdb/TermTypeSearch.ts
@@ -30,15 +30,34 @@ const useCases = {
 		TermTypeGroups.GENE_EXPRESSION,
 		TermTypeGroups.METABOLITE_INTENSITY
 	],
-	summary: [TermTypeGroups.DICTIONARY_VARIABLES, TermTypeGroups.GENE_EXPRESSION],
-	barchart: [TermTypeGroups.DICTIONARY_VARIABLES, TermTypeGroups.MUTATION_CNV_FUSION, TermTypeGroups.GENE_EXPRESSION],
-	violin: [TermTypeGroups.DICTIONARY_VARIABLES, TermTypeGroups.MUTATION_CNV_FUSION, TermTypeGroups.GENE_EXPRESSION],
-	sampleScatter: [TermTypeGroups.DICTIONARY_VARIABLES, TermTypeGroups.GENE_EXPRESSION], //This case covers scaleBy and dynamic scatters with coordinates from numeric terms
+	summary: [TermTypeGroups.DICTIONARY_VARIABLES, TermTypeGroups.GENE_EXPRESSION, TermTypeGroups.METABOLITE_INTENSITY],
+	barchart: [
+		TermTypeGroups.DICTIONARY_VARIABLES,
+		TermTypeGroups.MUTATION_CNV_FUSION,
+		TermTypeGroups.GENE_EXPRESSION,
+		TermTypeGroups.METABOLITE_INTENSITY
+	],
+	violin: [
+		TermTypeGroups.DICTIONARY_VARIABLES,
+		TermTypeGroups.MUTATION_CNV_FUSION,
+		TermTypeGroups.GENE_EXPRESSION,
+		TermTypeGroups.METABOLITE_INTENSITY
+	],
+	sampleScatter: [
+		TermTypeGroups.DICTIONARY_VARIABLES,
+		TermTypeGroups.GENE_EXPRESSION,
+		TermTypeGroups.METABOLITE_INTENSITY
+	], //This case covers scaleBy and dynamic scatters with coordinates from numeric terms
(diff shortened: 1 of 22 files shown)


server/src/termdb.js
```diff
@@ -474,10 +475,8 @@ async function trigger_getDefaultBins(q, ds, res) {
 	const lst = []
 	let min = Infinity
 	let max = -Infinity
-	if (tw.term.type == TermTypes.GENE_EXPRESSION) {
-		if (ds.queries.geneExpression.gene2bins[tw.term.gene])
```
xzhou82, line 478: should keep this cache, so that multiple users requesting bin for the same gene can be quickly returned without recomputing

airenzp: when testing it, the request is not happening, the second time, probably because of the cache in the browser.

client/termdb/TermTypeSearch.ts
```diff
@@ -13,7 +12,12 @@ The tree target is used to determine the allowed term types.
  */
 
 const useCases = {
-	matrix: [TermTypeGroups.DICTIONARY_VARIABLES, TermTypeGroups.MUTATION_CNV_FUSION],
```
xzhou82, line 16: here worth to add a comment and state that this design is deliberate so when introducing a new term, it can be gradually added to each usecase and get enabled in the plot/target code. looks redundant but necessary

airenzp: I think we can start listing the cases excluded instead. Working on it.

xzhou82: yes good idea

karishma-gangwani:
I see some dictionary terms when clicking 'Data dictionary' on redomics UI. but when selecting x and y coordinates in scatter plot I don't see the tree but only the search box is shown. Is that intentional?

airenzp:
> I see some dictionary terms when clicking 'Data dictionary' and 'Dictionary variables' on redomics UI. but when selecting x and y coordinates in scatter plot I don't see the tree but only the search box is shown. Is that intentional?

This dataset does not have dictionary terms that are numeric, that is why.

gavrielm:
Only condition dictionary terms should be allowed for cumulative incidence. The dictionary is filtered properly here, but non-dictionary terms should not be allowed:
image

Note that non-dictionary terms should still be allowed for overlay and divideBy in cumulative incidence.

xzhou82:
cuminc term1 can only be `condition` term. overlay/divideby can be any dict term, which can expand to non-dict terms later

gavrielm:
Survival should be fixed in the same way as cumulative incidence:
image

gavrielm:
For linear regression outcome, gene expression term could be used because it is continuous variable, but entering `TP53` here will break:
image

Similar case with logistic regression.

For cox regression, only dictionary term should be allowed.

Could disable non-dictionary term for any regression use case for now.

xzhou82:
regression: can only enable dict term/snplst/snplocus for now. will enable more in that code later
download: all term types should be supported

airenzp:
> Survival should be fixed in the same way as cumulative incidence: image

Thanks for reporting this. Survival and cuminc should be fixed now

airenzp:
Thanks for the input provided. The cases reported should be fixed now.

xzhou82:
seems[ gene exp barplot](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22term%22:{%22type%22:%22geneExpression%22,%22gene%22:%22TP53%22,%22name%22:%22TP53%22},%22q%22:{%22mode%22:%22continuous%22}}}]}) no longer shows

airenzp:
> seems[ gene exp barplot](http://localhost:3000/?noheader=1&mass=%7B%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:%7B%22activeTab%22:1%7D,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22term%22:%7B%22type%22:%22geneExpression%22,%22gene%22:%22TP53%22,%22name%22:%22TP53%22%7D,%22q%22:%7B%22mode%22:%22continuous%22%7D%7D%7D%5D%7D) no longer shows

if you restart the server and try the link it works. I am looking into this issue. I am not sure what happens that causes this to break, but I think it is related with the cache.

xzhou82:
thanks it works now! probably ran into same issue with our tooling as before

xzhou82 (approved):

PR #1622, opened 2024-05-21
build ds.queries.metaboliteIntensity.find

josephpowi:
## Description
ds.queries.metaboliteIntensity.find:
   read the metabolite intensity file only once and save the metabolites into cache (q.metabolites). When you do metabolite search, the code will first check if q.metabolites is available. If not, reading through the metabolite file is needed. 

## Checklist

### DIFF
--- a/server/routes/termdb.cluster.ts
+++ b/server/routes/termdb.cluster.ts
@@ -287,6 +287,30 @@ async function validateMetaboliteIntensityNative(q: MetaboliteIntensityQueryNati
 		}
 	}
 
+	q.find = async (param: TermdbClusterRequest) => {
+		// if !q.metabolites, read all metabolites from file
+		if (!q.metabolites) {
+			const metabolites = [] as string[]
+			await utils.get_lines_txtfile({
+				args: [q.file],
+				callback: line => {
+					const l = line.split('\t')
+					if (l[0].startsWith('#Metabolites')) return
+					metabolites.push(l[0])
+				}
+			} as any)
+			q.metabolites = metabolites
+		}
+		const matches = []
+		for (const m of param.metabolites!) {
+			if (!m) continue
+			for (const metabolite of q.metabolites) {
+				if (metabolite.toLowerCase().includes(m.toLowerCase())) matches.push(metabolite)
+			}
+		}
+		return { matches }
+	}
+
 	q.get = async (param: TermdbClusterRequest) => {
 		const limitSamples = await mayLimitSamples(param, q.samples, ds)
 		if (limitSamples?.size == 0) {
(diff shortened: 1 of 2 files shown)


server/routes/termdb.cluster.ts
```diff
@@ -287,6 +287,30 @@ async function validateMetaboliteIntensityNative(q: MetaboliteIntensityQueryNati
 		}
 	}
 
+	q.find = async (param: TermdbClusterRequest) => {
+		// if !q.metabolites, read all metabolites from file
+		if (!q.metabolites) {
```
xzhou82, line 292: customary to name it `q.__metabolites` to indicate this is temp variable

server/routes/termdb.cluster.ts
```diff
@@ -287,6 +287,30 @@ async function validateMetaboliteIntensityNative(q: MetaboliteIntensityQueryNati
 		}
 	}
 
+	q.find = async (param: TermdbClusterRequest) => {
```
xzhou82, line 290: it doesn't make sense to use TermdbClusterRequest type of parameter. try to move this initiator function out of `routes/termdb.cluster.ts` and into `mds3.init.js`

josephpowi: the hierachical clustering will be implemented to metabolite data. Do you think  we should not use TermdbClusterRequest as the parameter for metabolites data?

airenzp: I think he means that the find method is not going to be called from the client so the input parameter does not have to be a request, the get does have to be a request as it is called from the client to do the hierarchical clustering.

airenzp: we copy pasted from the get method as the input is very similar, that is why we used the request param, we can improve it.

josephpowi:
1. move metaboliteIntensity initiator function from termdb.cluster.ts to mds3.init.js
2. resolved 'unknown gene' error when gene can not be found from db. Now printing warning message instead of throwing an error

server/routes/termdb.cluster.ts
```diff
@@ -224,7 +218,10 @@ async function validateNative(q: GeneExpressionQueryNative, ds: any, genome: any
 			if (!g.chr) {
 				// quick fix: newly added gene from client will lack chr/start/stop
 				const re = getResultGene(genome, { input: g.gene, deep: 1 })
-				if (!re.gmlst || re.gmlst.length == 0) throw 'unknown gene'
+				if (!re.gmlst || re.gmlst.length == 0) {
+					console.warn('unknown gene:' + g.gene) //throw 'unknown gene'
```
xzhou82, line 222: thanks for the fix; please comment a todo that unknown genes should be notified to client

server/shared/types/dataset.ts
```diff
@@ -308,7 +308,9 @@ export type MetaboliteIntensityQueryNative = {
 	src: 'native' | string
 	file: string
 	samples?: number[]
+	_metabolites?: string[]
```
xzhou82, line 311: in type def, always add comment on every property. here indicate this is a dynamically built cache of string names to speed up search

xzhou82:
thanks looks good overall

xzhou82 (approved):
works well thanks

PR #1637, opened 2024-05-31
fix: Do not trim text returned by gdc gene expression /values API to …

xzhou82:
…handle the case of missing value cases at the end

## Description

local gene exp views work
awaiting for Edgar to test 

## Checklist

### DIFF
--- a/release.txt
+++ b/release.txt
@@ -1,1 +1,2 @@
-
+Fixes:
+- Do not trim text returned by gdc gene expression /values API to handle the case of missing value cases at the end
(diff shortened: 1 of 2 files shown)


siosonel:
@xzhou82 I'll wait a bit to see whether Phil will fix on their side, now that they are able to replicate the issue of missing data for the last two columns. We can create another release on Monday for late changes, and other fixes before the feature freeze on Friday 6/7.

siosonel:
may not need this change, [UC is working on a fix](https://gdc-ctds.atlassian.net/browse/SV-2468): 
"We are able to reproduce the issue when requesting with that payload.  We further isolated the issue to case_id ada19f65-5256-4c79-b3b9-7b9da69be437.  The /availability endpoint states that data is available for this case.  However, the /values endpoint returns no data for this case.

The UChicago team will continue to investigate."

xzhou82:
sure will close this now. if later turns out the pp cache (set of cases with gene exp data) can indeed mismatch with api/data release, then will need this fix again

PR #1640, opened 2024-05-31
Fixed violin bug

### LINKED ISSUE #1639: wrong violin plot in showing distribution skewed to max
[test here](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22MB_meta_analysis%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22Confidence%20Score%22,%22q%22:{%22mode%22:%22continuous%22}}}]})

airenzp:
## Description

Closes #1639
This PR solves the pending issues in the violin found when looking at the confidence score in mbmeta

## Checklist


xzhou82:
subtype 5 violin still looks wrong.
also found a separate issue, on dragging on violin, it does not properly calculate a numeric range under selection and seems to hardcode to `toFixed(1)`

airenzp:
> subtype 5 violin still looks wrong. also found a separate issue, on dragging on violin, it does not properly calculate a numeric range under selection and seems to hardcode to `toFixed(1)`

I am not sure what you mean with the separate issue. As about the violin, One think we can look at is into finding a better way to tune the bandwidth. If you remember, this was a pending issue, the silvermanBandwidth is fast to calculate but not always accurate. There were other methods that were more expensive but more trustworthy. Lets talk about this on monday. Here it is the plot from `ggplot(df, aes(x = x, y = y)) + geom_violin()` for the G3/G4-5, n=67

Screenshot 2024-05-31 at 5 39 35 PM

xzhou82:
yes that's the issue with subtype-5 violin. it's important we use a reliable method to render correct violin

i created a new issue about the separate thing

xzhou82 (approved):
the problem is resolved now. thank you very much
if bin number is an important param to control plot, we may consider using a bigger value by default, from 20 to e.g. 50, that seems to give a more realistic distribution

PR #1657, opened 2024-06-07
Add plotBrainImaging.py to server package and python dependencies to deps image 

aacic:
## Description

Phyton virtual environment added to deps image and phyton dependencies required by plotBrainImaging.py

@airenzp can you verify this: 

```
RUN /opt/venv/bin/pip install --upgrade pip && \
    /opt/venv/bin/pip install nibabel matplotlib numpy
```

## Checklist


siosonel:
@aacic I don't think we need to worry about virtual environment inside the container, it's only one use case where we know for sure there will not be any conflict by other python code

aacic:
> @aacic I don't think we need to worry about virtual environment inside the container, it's only one use case where we know for sure there will not be any conflict with other python code

It works when I install it like:
`
RUN pip3 install nibabel matplotlib numpy --break-system-packages --user`

```
Or using apt-get:

RUN apt-get update && apt-get install -y \
 python3-matplotlib \
 python3-nibabel \
 python3-numpy \
```

@airenzp do you know what's the better option?

siosonel (approved):

PR #1674, opened 2024-06-12
Enable survival use cases

### LINKED ISSUE #1651: Term type search in survival plot
For survival plot, term type search correctly restricts to survival dictionary terms for term1:

image

But term type search for term2 and term0 should not restrict to dictionary terms:

image

Any type of term should be allowed for term2 and term0.

@siosonel, feel free to include your suggestions here as well.

airenzp:
## Description

This PR enables the survival cases missing. Closes #1651. It also changes the default mode used in geneExpression to discrete. I felt at the barchart/summary and the scatter, we always start using numeric terms in discrete mode when doing overlays, so it made more sense to use it as the default mode. If so, we may want to do this for all numeric non dict terms when filling the term wrapper.

## Checklist

### DIFF
--- a/client/termsetting/handlers/geneExpression.ts
+++ b/client/termsetting/handlers/geneExpression.ts
@@ -29,7 +29,7 @@ export async function fillTW(tw: GeneExpressionTW, vocabApi: VocabApi, defaultQ:
 	if (typeof tw.term.gene != 'string' || !tw.term.gene) throw 'geneExpression tw.term.gene must be non-empty string'
 	if (!tw.term.name) tw.term.name = tw.term.gene // auto fill if .name is missing
 
-	if (!tw.q?.mode) tw.q = { mode: 'discrete' } // supply default q if missing
+	if (!tw.q?.mode) tw.q = { mode: 'continuous' } // supply default q if missing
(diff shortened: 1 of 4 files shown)


client/termsetting/handlers/geneExpression.ts
```diff
@@ -29,7 +29,7 @@ export async function fillTW(tw: GeneExpressionTW, vocabApi: VocabApi, defaultQ:
 	if (typeof tw.term.gene != 'string' || !tw.term.gene) throw 'geneExpression tw.term.gene must be non-empty string'
 	if (!tw.term.name) tw.term.name = tw.term.gene // auto fill if .name is missing
 
-	if (!tw.q?.mode) tw.q = { mode: 'continuous' } // supply default q if missing
```
xzhou82, line 32: any reason to default to discrete?

airenzp: See in the PR description

xzhou82: default q=discrete should be set in controls.divide.js rather than here. so to maintain behavior of showing violin for gene exp. i will change it

xzhou82 (approved):

PR #1686, opened 2024-06-13
Use automatic rounding based on the number

airenzp:
## Description

This draft PR is to help to address the range issues discussed.
Some of the issues with the violin ranges are related with the rounding used. A previous fix was deciding the number of digits to round based on the term type, if the type was integer would use 0 digits, that needs to be fixed as it was clipping the brush ranges.  I removed that and used roundValueAuto as it helps to decide the number of digits to keep based on the number. For a number 0.0003 for example, you want to keep 4 digits. but for 35.4890 is ok to keep just two. 

Feel free to disregard, I just wanted to give you my feedback in case it helps. I think now the next thing to check is the shifting of the brush painted.

## Checklist


creilly8 (approved):

xzhou82:
thank you both. i noticed two separate issues about violin/density and will fix next

PR #1701, opened 2024-06-14
optimize to require tape bundling only in certain contexts 

siosonel:
## Description

no tape bundling in dev, and only rebundle for CLI tests if the bundle is missing

To test:
- stop `npm run dev`
- `rm client/test/tape.bundle.js*`
- rerun `npm run dev`
- `http://localhost:3000/testrun.html?name=*.unit` should pass, so it works without the webpack-bundled `tape` code
- run from the client dir `npm run test:unit` or `./test.sh *tvs.unit.spec.*`, should pass
- run the same command again, the webpack bundling should not be triggered

## Checklist

### DIFF
--- a/client/dev.sh
+++ b/client/dev.sh
@@ -4,10 +4,6 @@ set -euxo pipefail
 
 rm -rf ./dist
 
-# TODO: 
-# use a good esbuild node polyfill plugin to avoid having to use webpack,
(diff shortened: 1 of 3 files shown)


xzhou82 (approved):
works thanks a lot

PR #1704, opened 2024-06-14
Implementing GSEA

robinpaul85:
## Description
Implementing GSEA within ProteinPaint using [blitzgsea](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9004650/). 

Need to install blitzgsea python package. This can be done using `pip3 install blitzgsea`

For testing please use this sessions file [sjmb12_GSEA.txt](https://github.com/user-attachments/files/15892252/sjmb12_GSEA.txt)

TODO:
1) Need to test blitzgsea as a dependency in the docker image.

## Checklist


server/routes/genesetEnrichment.ts
```diff
@@ -0,0 +1,122 @@
...
+		await fs.promises.stat(path)
+	} catch (e) {
+		throw `${path} does not exist`
+	}
+	return new Promise((resolve, reject) => {
+		const _stdout: any[] = []
+		const _stderr: any[] = []
+		// spawn python process
+		let python_path = 'python3'
+		if (serverconfig.py_path) python_path = serverconfig.py_path
```
siosonel, line 82: The default python binary path should be set in `src/serverconfig.js`, like other tools. It's centralized there to minimize confusing overrides that occur randomly in the server codebase. And prefer to name this as `serverconfig.python` to be consistent with name casing of the other tools (they use camelCase instead of underscore/snake_case.)

robinpaul85: Changed `serverconfig.py_path` to `serverconfig.python`.

server/routes/genesetEnrichment.ts
```diff
@@ -79,7 +79,7 @@ async function run_gsea(path, data) {
 		const _stderr: any[] = []
 		// spawn python process
 		let python_path = 'python3'
-		if (serverconfig.py_path) python_path = serverconfig.py_path
+		if (serverconfig.python) python_path = serverconfig.python
```
siosonel, line 82: @robinpaul85 What I meant in my previous comment is this:
- in `server/src/serverconfig.js`, add the last line
```js
if (!serverconfig.Rscript) serverconfig.Rscript = 'Rscript'
if (!serverconfig.gfServer) serverconfig.gfServer = 'gfServer'
if (!serverconfig.gfClient) serverconfig.gfClient = 'gfClient'
if (!serverconfig.python) serverconfig.python = 'python3'
```

- in `genesetEnrichment.ts`, do not use `python_path`, use `serverconfig.python` directly. This will guarantee that all nodejs script use the same tool binary, instead of each code file setting its own default binary path, we're trying to avoid that
```ts
const sp = spawn(serverconfig.python, [path])
```

PR #1748, opened 2024-06-24
Backwards compatibility fix for old saved sessions missing .isOrdered…

creilly8:
… and/or .disabled

## Description
PR as a suggested fix for issue found an old session file by @karishma-gangwani. Without this fix, `Error: Cannot read properties of undefined (reading 'isOrdered')` appears. 

## Checklist

### DIFF
--- a/client/mass/sessionBtn.js
+++ b/client/mass/sessionBtn.js
@@ -86,6 +86,7 @@ class MassSessionBtn {
 				if (loc.includes('browser')) {
 					this.sessionName = id
 					const state = this.savedSessions[id]
+					await preprocessState(state, this.app)
 					const targetWindow = this.dom.tip.d.node().querySelector(`[name="${radioName}"]:checked`).value
 					if (targetWindow == 'current') {
 						this.app.dispatch({ type: 'app_refresh', state })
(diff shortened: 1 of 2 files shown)


karishma-gangwani:
I think the callbacks for the 'sortByMutation' and 'sortByCNV' would also need to be checked? because the sorting functionality on the UI won't work with that session file.

creilly8:
@karishma-gangwani and I figured out that the sorting still doesn't work with this change because there is only one `tiebreaker` entry under `sortPriority` (see `matrix.controls.js`, line 22).

siosonel:
I thought I had fixed this issue before, see the `reshapeSortPriority()` function in `matrix.sort.js`. Some logic would need to be added to this function to consider other saved sessions that are not handled properly. Making the change in` matrix.config.js` is likely not sufficient because other parts of the code now expect the new sorter data shape.

siosonel:
@creilly8 I fixed by editing the recovered session state using each plot's applicable `getPlotConfig()`. However, there are still cases that would not be recovered/rehydrated correctly, such as when the saved session uses a dictionary term that no longer exists on the server.

siosonel (approved):

PR #1775, opened 2024-06-28
fix: Speed up top variably expressed genes query from gdc api, direct…

xzhou82:
…ly submit case filter and do not first retrieve list of cases

## Description

querying top variably expressed genes from gdc sees a speed gain. there had been a step to first query list of cases passing filter (takes 10 seconds, especially bad when there's no cohort), then pass these cases to /gene_selection/ api to get top genes. now the case querying step is eliminated. the cohort is directly passed to api (didn't know that before)

for exp clustering, will still retrieve list of cases from cohort as before, since the app must restrict to 1K cases otherwise it breaks. no speed gain there

all tested to work: adhoc and integrated views, non-ci test, adding gene exp row to matrix

## Checklist

### DIFF
--- a/release.txt
+++ b/release.txt
@@ -1,1 +1,2 @@
-
+Fixes:
+- Speed up top variably expressed genes query from gdc api, directly submit case filter and do not first retrieve list of cases
--- a/server/routes/termdb.topVariablyExpressedGenes.ts
+++ b/server/routes/termdb.topVariablyExpressedGenes.ts
@@ -169,7 +169,8 @@ function gdcValidateQuery(ds: any, genome: any) {
 	}
 
 	function getGeneSelectionArg(q: any) {
-		const arg = {
+		const arg: any = {
+			// add any to avoid tsc err
 			case_filters: makeFilter(q),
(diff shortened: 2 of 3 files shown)


karishma-gangwani (approved):

congyu-lu (commented):
can no longer add gene expression terms to GDC oncoMatrix any more, from server log:
_Start gdc gene exp query...
1 out of 1 genes selected for exp: 16 ms
case submitter id unknown for a uuid_

If on master, can still add gene expression terms to GDC oncoMatrix

xzhou82:
thanks for testing. i haven't run into the case id issue, and adding gene exp to oncomatrix works. could you delete the cache at `~/data/cache/extApiResponse/` and recache to test again?

congyu-lu:
> thanks for testing. i haven't run into the case id issue, and adding gene exp to oncomatrix works. could you delete the cache at `~/data/cache/extApiResponse/` and recache to test again?

Thanks, no error now and did see speed gain

congyu-lu (approved):

PR #1787, opened 2024-07-01
Improve termsbyids route

karishma-gangwani:
## Checklist

### DIFF
--- a/server/routes/termdb.termsbyids.ts
+++ b/server/routes/termdb.termsbyids.ts
@@ -20,9 +20,15 @@ export const api: any = {
 	}
 }
 
+/**
+ * Initializes the function with the given genomes object and handles the request and response asynchronously.
+ *
+ * @param {Object} genomes - The object containing genome data.
+ * @returns {Promise<void>} A promise representing the asynchronous operation.
+ */
 function init({ genomes }) {
 	return async (req: any, res: any): Promise<void> => {
-		const q = req.query // as getcategoriesRequest
+		const q = req.query as gettermsbyidsRequest
 		try {
 			const g = genomes[req.query.genome]
 			if (!g) throw 'invalid genome name'
@@ -31,7 +37,7 @@ function init({ genomes }) {
 			const tdb = ds.cohort.termdb
 			if (!tdb) throw 'invalid termdb object'
 
-			await trigger_gettermsbyid(q, res, tdb) // as getcategoriesResponse
+			await trigger_gettermsbyid(q, res, tdb)
 		} catch (e) {
 			// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 			// @ts-ignore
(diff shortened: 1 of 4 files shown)


xzhou82:
i see getTerms() is only used in termsetting.ts but don't know how to invoke it. how do you test getTerms()?

server/routes/termdb.termsbyids.ts
```diff
@@ -20,9 +20,15 @@ export const api: any = {
...
 }
 
+/**
+ * Initializes the function with the given genomes object and handles the request and response asynchronously.
+ *
+ * @param {Object} genomes - The object containing genome data.
+ * @returns {Promise<void>} A promise representing the asynchronous operation.
+ */
 function init({ genomes }) {
 	return async (req: any, res: any): Promise<void> => {
```
creilly8, line 30: It's better in this case to define the Request type above instead of using the `as` keyword like this `req: { query: gettermsbyidsRequest }`. Please see the explanation in [our guidelines](https://docs.google.com/document/d/1SO7pu8-cVA3W5FWUQi7NMFKjXiUaUh40eDQ7s86atOI/edit#heading=h.pjpvzb8lhvju).

server/routes/termdb.termsbyids.ts
```diff
@@ -59,5 +72,5 @@ async function trigger_gettermsbyid(
 	}
 	res.send({
 		terms
-	})
+	} as gettermsbyidsResponse)
```
creilly8, line 75: As a general rule, all typescript types, interfaces, and classes are in Pascal Case.

xzhou82 (approved):
thanks

server/routes/termdb.termsbyids.ts
```diff
@@ -41,6 +47,13 @@ function init({ genomes }) {
...
 
+/**
+ * Retrieves terms by their IDs and sends the terms as a response.
+ *
+ * @param {Object} q - An object containing the IDs of the terms to retrieve.
+ * @param {Object} res - An object with a 'send' function to send the response.
+ * @param {Object} tdb - An object with a 'termjsonByOneid' function to retrieve term by ID.
+ */
 async function trigger_gettermsbyid(
 	q: { ids: any },
```
creilly8, line 58: Defining the q as `{ids: any}` is a little misleading. The entire `q: gettermsbyidsRequest` is passed to this function. Yes, only the ids are used now, but let's say in the future there's genome or dataset specific logic to add. `{ids: any}` implies that information isn't available.

creilly8 (changes requested):

server/routes/termdb.termsbyids.ts
karishma-gangwani: yeah but ids could be anything from strings or numbers or some alphanumeric chars. for now I think this should be ok. can revise as needed later. I will address your first comment in a separate pr.

PR #1808, opened 2024-07-08
Hier cluster missing term type

congyu-lu:
## Checklist

### DIFF
--- a/client/plots/hierCluster.config.js
+++ b/client/plots/hierCluster.config.js
@@ -57,8 +57,13 @@ export async function getPlotConfig(opts = {}, app) {
 			const tw = i.term ? i : { term: i }
 
 			if (!tw.term.type) {
-				// must provide term type for hierCluster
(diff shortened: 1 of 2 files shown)


client/plots/hierCluster.config.js
```diff
@@ -57,8 +57,13 @@ export async function getPlotConfig(opts = {}, app) {
 			const tw = i.term ? i : { term: i }
 
 			if (!tw.term.type) {
-				// must provide term type for hierCluster
-				throw `term type is missing`
+				if (config.dataType) {
```
xzhou82, line 60: config.dataType should be required and this should be verified elsewhere. also please define supported values for dataType. in case of geneExpression and metabolite dataType can equal to term type. however it's not the case for dictionary terms. thus here will need some if(dataType=='geneExpression') then auto assign

xzhou82 (approved):

PR #1870, opened 2024-07-16
may pass unfrozen term, q to term1 pill to allow setting additional t…

siosonel:
…erm/q props on non-dictionary terms when edit menu is rendered

Tested with http://localhost:3000/testrun.html?name=*.unit and `cd client; npm run test:integration`. Also with geneExpression term summary plot, opens as violin plot, pill edit menu/toggle to `Discrete` mode should work without error.

NOTE: Won't fix here unrelated geneExpression pill UI issue that may be missing defaultQ, such as sampleScatter shape input.

## Checklist

### DIFF
--- a/client/plots/controls.term1.js
+++ b/client/plots/controls.term1.js
@@ -1,7 +1,7 @@
 import { getCompInit } from '../rx'
 import { termsettingInit } from '#termsetting'
 import { getNormalRoot } from '#filter'
-import { TermTypes } from '#shared/terms'
+import { TermTypes, isDictionaryType } from '#shared/terms'
 
 /*
 for configuring term1; wraps termsetting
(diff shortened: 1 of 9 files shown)


client/termdb/TermdbVocab.js
```diff
@@ -991,15 +991,13 @@ export class TermdbVocab extends Vocab {
 		//TODO use the PresetNumericBins type for defaultBins
 		const defaultBins = await this.getDefaultBins({ tw })
 		if ('error' in defaultBins) throw defaultBins.error
-
-		const termIsFrozen = Object.isFrozen(tw.term)
-		if (termIsFrozen) tw.term = structuredClone(tw.term)
-
```
siosonel, line 997: @congyu-lu Please note the changes and comments here, the previous code was not allowing term.bins to be set on the original `term` copy.

client/plots/controls.term1.js
```diff
@@ -146,9 +146,15 @@ function setRenderers(self) {
 				throw 'unknown term type'
 		}
 		if (!self.pill) self.setPill()
+		const isDictTerm = isDictionaryType(plot.term.term.type)
+		// assume that dictionary tw are complete after fillTermWrapper(),
+		// whereas non-dictionary terms may require additional term/q properties
+		// to be added when the pill edit UI is opened, so term/q must not be frozen
```
siosonel, line 152: @xzhou82 An assumption is made here that dictionary terms will not need new properties to be added when the pill edit menu is opened, whereas non-dict terms like geneExpression may require new term key-values to be set when rendering the edit menu, so `term`/`q` must not be frozen only for non-dict terms.

congyu-lu (approved):

PR #1879, opened 2024-07-18
put notes and instructions at testrun.html page bottom

xzhou82:
## Description

need Edgar's help to complete

## Checklist

### DIFF
--- a/public/testrun.html
+++ b/public/testrun.html
@@ -78,7 +78,7 @@
 
 const time = Date.now() // why need to add time to href?
 d3.selectAll('.allTestsBtn').each(function() {
-  this.href += `&time=${time}`
+  //this.href += `&time=${time}`
 })
 
 const backArrows = d3.select('#testrun-back-arrows')
@@ -102,8 +102,7 @@ function updateNavCrumbs(display, params={}) {
   currTest.style('display', display).text(pattern)
   backArrows.style('display', display)
   if (display == 'none') loadingDiv.html(
-    `Click on a directory or file name to trigger a test. ` +
-    `Tests in <span style='color: ${activeColor}'>red</span> are currently imported in 'client/test/internals.js'.` // don't see any red color. is this outdated?
+    `Click on a directory or file name to trigger a test.`
   )
 }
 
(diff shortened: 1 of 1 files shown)


siosonel (dismissed):

public/testrun.html
```diff
@@ -65,11 +63,24 @@ <h1>
...
+  <li>If tests are missing and console shows error with /specs?exclude= request, check your <code>sjpp/serverconfig.json</code> to make sure <span class='code-snippet'>"debugmode":true</span> is set at root level.</li>
+  <li>To run tests on commandline, do <span class='code-snippet'>npm run test:integration --workspace=client</span> at /sjpp/proteinpaint/ directory.
+    Or, from the client dir, <span class='code-snippet'>npm run test:integration</span> or <span class='code-snippet'>npm run test:unit</span>
+  </li>
+<ul>
+
 <script>
-const time = Date.now()
+
+const time = Date.now() // why need to add time to href?
```
xzhou82, line 79: comment on why time is needed?

public/testrun.html
```diff
@@ -92,7 +103,7 @@ <h1>
   backArrows.style('display', display)
   if (display == 'none') loadingDiv.html(
     `Click on a directory or file name to trigger a test. ` +
-    `Tests in <span style='color: ${activeColor}'>red</span> are currently imported in 'client/test/internals.js'.`
+    `Tests in <span style='color: ${activeColor}'>red</span> are currently imported in 'client/test/internals.js'.` // don't see any red color. is this outdated?
```
xzhou82, line 106: check out my question?

public/testrun.html
siosonel: I put the `time` URL param to force the browser to always reload a test glob pattern, I think early in the switch to esbuild, the test would not refresh to use a new client bundle. I tested just now without the `time` URL param, and edited client code and spec files, and I cannot replicate the previous issue, so I removed it for now as it doesn't seem to be needed anymore.

public/testrun.html
siosonel: This red text comment is outdated, from the previous webpack approach. All the spec files are always bundled now every time, but there is an if-else logic to only run specs that matchs the glob pattern in the URL params. 

Previously, selecting a test would trigger a request to the PP server, which triggers a rewrite of `server/test/internals.js`, and then that triggers a webpack rebundling. The current approach is much simpler, no need for a server request to trigger import rewrites and rebundling.

siosonel (dismissed):

PR #1908, opened 2024-07-23
matrix Continuous term zscore

congyu-lu:
## Checklist

### DIFF
--- a/client/plots/matrix.layout.js
+++ b/client/plots/matrix.layout.js
@@ -248,6 +248,15 @@ export function setLabelsAndScales() {
 				t.counts.maxval *= vc.scaleFactor
 			}
 
+			/*
+				compute mean and std on the client side using the renderedValues, instead of asking the server to 
+				transform the values when the data source does not actually change. That means re-reading 
+				db/files/external API for the same data on the server, it would impact the server less 
+				(by allowing caching and offloading computation) if the value transformation is done on 
+				the client. Also, in case some of the data/rendered filters are done on the client-side plot, 
+				then the server may not know which data are actually rendered, meaning the mean and stdev 
+				may not be computable on the server side.
+			*/
 			if (renderedContinuousVs.length && t.tw.q.convert2ZScore) {
 				const mean = renderedContinuousVs.reduce((acc, val) => acc + val, 0) / renderedContinuousVs.length
 				const std = Math.sqrt(
--- a/client/plots/matrix.layout.js
+++ b/client/plots/matrix.layout.js
@@ -101,6 +101,7 @@ export function setLabelsAndScales() {
 	for (const t of this.termOrder) {
 		const countedSamples = new Set()
 		t.counts = { samples: 0, hits: 0 }
+		const renderedContinuousVs = []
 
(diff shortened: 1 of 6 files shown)


client/termsetting/handlers/numeric.continuous.ts
```diff
@@ -55,6 +55,19 @@ export function getHandler(self) {
 
 			vr.render()
 
+			const convert2ZScoreDiv = div.append('div').style('display', 'inline-block').style('padding', '3px 10px')
+
+			convert2ZScoreDiv
```
xzhou82, line 60: use `dom/checkbox.js` to create this

xzhou82: i agree to have the zscore option in termsetting edit menu. when the zscore box is checked, it shouldn't show the `Scale values` option as it cannot be coapplied to zscore

client/plots/matrix.layout.js
```diff
@@ -233,6 +241,24 @@ export function setLabelsAndScales() {
...
 		}
 		if (t.grp.type !== 'hierCluster' && t.tw.q?.mode == 'continuous') {
+			const vc = t.tw.term.valueConversion
+			if (vc) {
+				// convert values
+				t.counts.minval *= vc.scaleFactor
+				t.counts.maxval *= vc.scaleFactor
+			}
+
+			if (renderedContinuousVs.length && t.tw.q.convert2ZScore) {
```
xzhou82, line 251: how about you look into carrying out zscore computing in getData() so that no need to do all these on client. also that will make zscore to work for violin

congyu-lu: mean and std were calculated using the renderedValues, which are decided on client side (classifyValues). So cannot calculate zscore in getData() since at that time we don't know which values will be rendered yet.

siosonel: @xzhou82 my opinion: I think it's better to compute on the client side, instead of asking the server to transform the values when the data source does not actually change. That means re-reading db/files/external API for the same data on the server, it would impact the server less (by allowing caching and offloading computation) if the value transformation is done on the client.  Also, in case some of the data/rendered filters are done on the client-side plot, then the server may not know which data are actually rendered, meaning the mean and stdev may not be computable ton the server.

xzhou82: ok to merge. since the new boolean flag is added to tw.q{}, it becomes part of request payload and will incur more requests since payload are now different

xzhou82: Congyu should add these notes as comments otherwise i will be wondering about it again

xzhou82:
the zscore option now shows whenever qmode is continuous, e.g. for violin, but is nonfunctional. is this okay to merge?

congyu-lu:
> the zscore option now shows whenever qmode is continuous, e.g. for violin, but is nonfunctional. is this okay to merge?

Thanks changed the zscore option to show only in matrix continuous terms Edit

xzhou82 (approved):
thanks
in the separate branch to redo term typing, need to document the `convert2ZScore` property

PR #1912, opened 2024-07-23
fix for bug in label clicking for violin on overlaying variable

karishma-gangwani:
## Description

when clicking labels on violin plot breaks on master. small bug fix.

## Checklist

### DIFF
--- a/client/plots/violin.interactivity.js
+++ b/client/plots/violin.interactivity.js
@@ -101,9 +101,7 @@ export function setInteractivity(self) {
 
 		if (self.displayBrushMenu.called == true) {
 			const [niceStart, niceEnd] =
-				self.config.term.term.type == 'integer' || self.config.term.term.type == 'float'
-					? [Math.round(start), Math.round(end)]
-					: niceNumLabels([start, end])
+				self.config.term.term.type == 'integer' ? [Math.round(start), Math.round(end)] : niceNumLabels([start, end])
 
 			self.app.tip.d.append('div').text(`From ${niceStart} to ${niceEnd}`)
 		}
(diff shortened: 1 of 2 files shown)


client/plots/violin.interactivity.js
```diff
@@ -99,10 +99,14 @@ export function setInteractivity(self) {
 		//For testing and debugging
 		self.app.tip.d.classed('sjpp-violin-brush-tip', true)
 
-		const [niceStart, niceEnd] =
-			self.config.term.term.type == 'integer' ? [Math.round(start), Math.round(end)] : niceNumLabels([start, end])
+		if (self.displayBrushMenu.called == true) {
+			const [niceStart, niceEnd] =
+				self.config.term.term.type == 'integer' || self.config.term.term.type == 'float'
```
xzhou82, line 104: only when type is integer should it round.

client/plots/violin.interactivity.js
```diff
@@ -99,10 +99,12 @@ export function setInteractivity(self) {
 		//For testing and debugging
 		self.app.tip.d.classed('sjpp-violin-brush-tip', true)
 
-		const [niceStart, niceEnd] =
-			self.config.term.term.type == 'integer' ? [Math.round(start), Math.round(end)] : niceNumLabels([start, end])
+		if (self.displayBrushMenu.called == true) {
```
xzhou82, line 102: use of this flag is unreliable. you can first brush, then click on label and it breaks again
also it's bad the displayMenu function 3rd arg can be either an object, or number. suggest to change to `displayMenu({event, options, plot})` and `display({event, options, start, stop})`

karishma-gangwani: okay. can you check now? I modified the arguments

xzhou82 (approved):
thanks for fixing the issue. ok to merge now but the use of the arbitrary boolean flag `called` on the function and create filter logic should be cleaned up and improved

karishma-gangwani:
> thanks for fixing the issue. ok to merge now but the use of the arbitrary boolean flag `called` on the function and create filter logic should be cleaned up and improved

okay sure I will look into cleaning that up. Thanks

PR #1922, opened 2024-07-24
Groupsetting type separated into TermGroupsetting and QGroupsetting

creilly8:
## Description
First of many PRs to address type changes, errors, etc.

Changes: 
1. Separated Groupsetting type into TermGroupsetting and QGroupsetting. 
2. Fixed type errors in categorical and single cell handlers created by separating the types. 
3. Removed commented out unused `getQlst` code
4. Cleaned up comments and term names types, when appropriate

Test: 
- Run `npx tsc`

## Checklist


xzhou82 (approved):
thanks for the efforts. this is gradual cleaning up of all term related type def and checking. we want to do small pr if possible. the next fix for the groupsetting type def is planned

server/shared/types/terms/term.ts
```diff
@@ -59,29 +57,28 @@ export type GroupSetEntry = BaseGroupSet & {
...
 
-export type CustomGroupSetting = {
+export type CustomQGroupSetting = {
 	/** When “predefined_groupset_idx” is undefined, will use this set of groups.
 	This is a custom set of groups either copied from predefined set, or created with UI.
 	Custom set definition is the same as a predefined set. */
 	customset: BaseGroupSet
-	disabled?: boolean
+	/** if false, not applied */
 	inuse?: boolean
```
gavrielm, line 66: Should `inuse` be required?

server/shared/types/terms/term.ts
```diff
@@ -59,29 +57,28 @@ export type GroupSetEntry = BaseGroupSet & {
...
-	inuse?: boolean
-	disabled?: boolean
-	useIndex?: number
-	/**Value is array index of term.groupsetting.lst[] */
+export type PredefinedQGroupSetting = {
+	/** If .inuse true, apply and will require predefined_groupset_idx */
+	/** Value is array index of term.groupsetting.lst[] */
 	predefined_groupset_idx: number
-	lst: GroupSetEntry[]
+	inuse: true
```
gavrielm, line 73: Should this be `inuse: boolean`?

server/shared/types/terms/term.ts
```diff
@@ -59,29 +57,28 @@ export type GroupSetEntry = BaseGroupSet & {
...
 }
 
-export type EmptyGroupSetting = {
-	inuse?: false
-	disabled?: true
+export type QGroupSetting = PredefinedQGroupSetting | CustomQGroupSetting
+
+export type TermGroupSetting = {
+	/** if there are only two values, means groupsetting definition is not applicable for the term */
+	disabled?: boolean
```
gavrielm, line 80: I think `disabled` should be required

server/shared/types/terms/geneVariant.ts
```diff
@@ -40,9 +39,14 @@ export type GeneVariantGeneTerm = BaseTerm & {
 	stop?: number
 }
 
-export type GeneVariantTerm = GeneVariantCoordTerm | GeneVariantGeneTerm
+export type GeneVariantTerm =
+	| GeneVariantCoordTerm
+	| (GeneVariantGeneTerm & {
+			groupsetting: TermGroupSetting
+	  })
```
gavrielm, line 46: Does this only add `groupsetting` property onto `GeneVariantGeneTerm`? If so, it should also be added onto `GeneVariantCoordTerm`.

server/shared/types/terms/term.ts
creilly8: `inuse` isn't always present in groupsetting. For example in the categorical and single cell handlers use this code `if (!('inuse' in tw.q.groupsetting)) tw.q.groupsetting.inuse = false` because the it's potentially an empty object from the lines above.

server/shared/types/terms/term.ts
creilly8: Yes, it should. Thanks for point that out.

server/shared/types/terms/term.ts
creilly8: We chatted about about this property this morning. I don't think always present but I'll double check.

server/shared/types/terms/geneVariant.ts
creilly8: Yes, you are correct. I will be changing this to a super type in another PR.

PR #2027, opened 2024-08-15
fix: At GDC lollipop subtrack, disable survival term when building th…

### LINKED ISSUE #1948: in gdc mds3 tk, when selecting term for subtk filtering, hide survival term from tree
gdc cohort filter doesn't support survival term

xzhou82:
…e filter

## Description

closes #1948 

please pull sjpp master to test

test at http://localhost:3000/?genome=hg38&gene=ENST00000407796&mds3=GDC. build a subtk. click on the filter handle on left, and +ADD, at the dict tree, Overall survival is no longer present

for a non-gdc case, the survival term can still be added to subtk filter

all CI passed

## Checklist

### DIFF
--- a/server/shared/termdb.usecase.js
+++ b/server/shared/termdb.usecase.js
@@ -17,13 +17,15 @@ export const graphableTypes = new Set([
 	TermTypes.SNP
 ])
 /*
-	isUsableTerm() will
+isUsableTerm() will
+
 	- centralize the "allowed term" logic
 	which can be intricate or dataset-specific 
 	for certain terms or contexts
 	- make it easy to handle new term types
 
-	Arguments:
+Arguments:
+
 	term {}
 		.type: 'categorical', etc.
 		.child_types: []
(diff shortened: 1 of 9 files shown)


server/shared/termdb.usecase.js
```diff
@@ -179,6 +179,18 @@ export function isUsableTerm(term, _usecase, ds) {
...
+		case 'filter':
+			if (graphableTypes.has(term.type)) {
+				if (usecase.detail == 'GDC') {
+					// is gdc ds. api doesn't support filtering by survival, do not show survival
+					if (term.type != 'survival') uses.add('plot')
+				} else {
+					uses.add('plot')
+				}
+			}
+			if (!term.isleaf) uses.add('branch')
```
siosonel, line 191: Should use `if (hasNonSurvivalTermChild(child_types)) uses.add('branch')` here.? The way it's written above, the survival branch is assumed to never have a non-survival term as a descendant, but it's better to be flexible now instead of potentially having to revisit this logic later.

server/shared/termdb.usecase.js
```diff
@@ -179,6 +179,18 @@ export function isUsableTerm(term, _usecase, ds) {
 				return uses
 			}
 
+		case 'filter':
+			if (graphableTypes.has(term.type)) {
+				if (usecase.detail == 'GDC') {
```
siosonel, line 184: There is `ds` argument, maybe `GDC` should not be hardcoded. It should be `if (!ds.cohort?.excludedTermTypesByUseCase?.filter?.includes == 'gdc') uses.add('plot')`

xzhou82: this can be hard. it intends ds to be server-side obj. here the relevant logic calls isUsable on client, so it's only able to pass in termdbconfig. and must change a dozen places isUsable is called on client

siosonel (approved):

PR #2057, opened 2024-08-26
Remove plot errors when no plots render

creilly8:
## Description
@airenzp, Please let me know if this is what you had in mind. 

## Checklist

### DIFF
--- a/client/plots/plot.disco.js
+++ b/client/plots/plot.disco.js
@@ -24,8 +24,11 @@ genomeObj={}
 
 _overrides={}
 	optional override parameters to pass to disco
+
+throwError
+	optional, instead of showing error to the user, throw
 */
-export default async function (termdbConfig, dslabel, sample, holder, genomeObj, _overrides = {}) {
+export default async function (termdbConfig, dslabel, sample, holder, genomeObj, _overrides = {}, throwError) {
 	const overrides = computeOverrides(_overrides, termdbConfig, genomeObj, sample)
 
 	const loadingDiv = holder.append('div').style('margin', '20px').text('Loading...')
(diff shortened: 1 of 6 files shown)


client/plots/sampleView.js
```diff
@@ -567,8 +567,15 @@ class SampleView {
 			for (const sample of samples) {
 				const cellDiv = div.append('div').style('display', 'inline-block')
 				const wsiViewer = await import('./wsiviewer/plot.wsi.js')
-				wsiViewer.default(state.vocab.dslabel, cellDiv, this.app.opts.genome, sample.sampleName)
+				await wsiViewer.default(state.vocab.dslabel, cellDiv, this.app.opts.genome, sample.sampleName)
```
airenzp, line 570: I would throw the error and make the try catch here, if there is an error continue, nothing would be added to cellDiv

client/plots/sampleView.js
```diff
@@ -567,8 +567,15 @@ class SampleView {
...
 			}
+
```
airenzp, line 572: This block would not be needed

client/plots/plot.disco.js
```diff
@@ -45,7 +48,12 @@ export default async function (termdbConfig, dslabel, sample, holder, genomeObj,
 			sample: sample[termdbConfig.queries.singleSampleMutation.sample_id_key]
 		}
 		const data = await dofetch3('termdb/singleSampleMutation', { body })
-		if (data.error) throw data.error
+		if (data.error) {
```
airenzp, line 51: I would keep the throw here and in the catch would throw if throwError is true

PR #2068, opened 2024-08-27
Handling survival use cases after moving survival annotation to patients

airenzp:
## Description

To test it please use termdbTest with the XYZ cohort, that has 28 patients and 40 samples.
Everything seems to work as expected, but when changing overlay from a patient term to a sample term there is an error. 
`Error: R process exited with non-zero status code=1 R stderr: Error in solve.default(vv, temp2) : Lapack routine dgesv: system is exactly singular: U[1,1] = 0 Execution halted
`
Possibly is expecting the same number of samples?
See corresponding PR in [sjpp](https://github.com/stjude/sjpp/pull/450). Please check. We can also huddle to explain the PR. 

## Checklist


server/src/termdb.survival.js
```diff
@@ -44,8 +44,13 @@ export async function get_survival(q, ds) {
...
 		// ot: overlay term, the series term
 		const ot = q[`term${survTermIndex == 1 ? 2 : 1}`]
-
-		const data = await getData({ terms: twLst, filter: q.filter }, ds, q.genome, true)
+		const types = new Set()
+		const ids = [st.id]
+		if (ot) ids.push(ot.id)
+		if (q.term0) ids.push(q.term0.id)
+		for (const id of ids) types.add(ds.cohort.termdb.term2SampleType.get(id))
+		const onlyChildren = types.size > 1
```
xzhou82, line 52: i would like to see more comments on these logic. especially where an assumption is made (e.g. survival data is always on "root" samples). so that we can be aware of those. thanks a lot

gavrielm:
Cannot replicate error with overlay term. Which term triggers an error?

server/src/termdb.survival.js
airenzp: there is no assumption that survival is a root term

karishma-gangwani:
> Cannot replicate error with overlay term. Which term triggers an error?

in XYZ cohort, under survival analysis with OS and overlay with 'Diagnosis group' breaks for me with the error that Airen reported

gavrielm (approved):

PR #2072, opened 2024-08-28
process q.hidden_values in categorical router

siosonel:
## Description

This bug was not released to any environment, so not adding a `release.txt` entry. And the bug may only affect edge cases, it's not visible when opening a few examples URL that would  have required `categoricalRouter.fill()`. So, this is mostly a proactive fix.

Tested with  http://localhost:3000/testrun.html?name=*.unit and `cd client; npm run test:integration`.
 
## Checklist


xzhou82:
what is the plot that can test tw.q.hiddenValues?

siosonel:
> what is the plot that can test tw.q.hiddenValues?

barchart, cuminc, violin. For example, uncomputable values should be hidden by default for numeric terms when opening a barchart example URL, it looks like it already works in master even before this fix.

Issue #2090, opened 2024-09-05
new mass nav tab `About`


xzhou82:
dataset should have optional "About" contents in `ds.cohort.termdb.about={html:'xxx'}` we can decide on details

if mass nav code finds it in termdbConfig, it should generate the `About` tab as the first one. clicking the tab should hide the chart button tray, and display the about contents in a new sandbox

expect about contents to be arbitrary, even images

PR #2095, opened 2024-09-06
Prototype About tab in mass app

creilly8:
## Description
Closes issue #2090. 

Changes: 
1. An About tab appears in the mass ui if `ds.cohort.termdb.about` is present and `ds.cohort.termdb.selectCohort` is not present.
2. Types for mass app and rx app api for typescript files. 

Test: 
1. Comment out `cohort.termdb.selectCohort` and uncomment `cohort.termdb.about` in `server/dataset/termdb.test.ts` -> open [TermdbTest example](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38-test%22,%22dslabel%22:%22TermdbTest%22}) -> should see the About tab in the last position. 
2. Does not appear in [other datasets, such as sjmb12](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22sjmb12%22}). To see in other datasets, add the following to `cohort.termdb`: 
```
about: {
    tab: {
	    order: 1,
	    topLabel: 'test',
	    midLabel: 'test about',
	    btmLabel: 'test'
    },
    html: 'Test'
}
```
## Checklist

### DIFF
--- a/client/mass/about.ts
+++ b/client/mass/about.ts
@@ -1,8 +1,8 @@
 import { getCompInit } from '../rx'
 import { Elem } from '../types/d3'
 import { MassAppApi } from './types/mass'
-import { newSandboxDiv } from '#dom/sandbox'
-import { NewSandbox } from '#dom/types/sandbox'
+import { newSandboxDiv } from '../dom/sandbox' //Do not use #dom. Will look for .js file
+import { NewSandbox } from '../dom/types/sandbox' //Do not use #dom. Will look for .js file
 
 /** --In development--
  * Only .html property enabled from ds.cohort.termdb.about
--- a/client/mass/about.ts
+++ b/client/mass/about.ts
@@ -1,13 +1,7 @@
 import { getCompInit } from '../rx'
 import { Elem } from '../types/d3'
 import { MassAppApi } from './types/mass'
-import { newSandboxDiv } from '../dom/sandbox' //Do not use #dom. Will look for .js file
-import { NewSandbox } from '../dom/types/sandbox' //Do not use #dom. Will look for .js file
 
-/** --In development--
- * Only .html property enabled from ds.cohort.termdb.about
- * Later may add other properties
- */
 type AboutObj = {
 	html: string
 }
@@ -23,31 +17,22 @@ class MassAbout {
 	app: MassAppApi
 	holder: Elem
 	features: AboutObj
-	sandbox: NewSandbox
-	readonly id = `${Math.random() * (1 - 0) + 1}`
 
 	constructor(opts: MassAboutOpts) {
 		this.type = 'about'
 		this.app = opts.app
 		this.holder = opts.holder
 		this.features = opts.features
-		this.sandbox = newSandboxDiv(this.holder, { plotId: this.id })
 	}
 
 	init() {
-		this.sandbox.app_div.style('display', 'none')
-		this.sandbox.header.append('div').text('About').style('color', 'rgb(85, 85, 85)')
-		this.sandbox.body.append('div').style('padding', '10px').html(this.features.html)
-	}
-
-	getState(appState) {
-		return {
-			nav: appState.nav
-		}
+		this.holder.append('div').style('padding', '10px').html(this.features.html)
 	}
 
 	main() {
-		this.sandbox.app_div.style('display', this.app.Inner.state.nav.activeTab == 0 ? '' : 'none')
+		const aboutIdx = this.app.Inner.components.nav.Inner.tabs.findIndex(tab => tab.subheader == 'about')
+		const isActive = this.app.Inner.state.nav.activeTab == aboutIdx
+		this.holder.style('display', isActive ? '' : 'none')
 	}
 }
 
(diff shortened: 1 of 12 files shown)


client/mass/nav.js
```diff
@@ -31,6 +32,7 @@ headtip.d.style('z-index', 5555)
 // headtip must get a crazy high z-index so it can stay on top of all, no matter if server config has base_zindex or not
 
 // data elements for navigation header tabs
+const aboutTab = { top: 'ABOUT', mid: '', btm: '', subheader: 'about' }
```
xzhou82, line 35: show word ABOUT in mid instead

airenzp: Maybe Home would be a better title for this tab if it is going to be the first one

airenzp (approved):

xzhou82:
i prefer ABOUT, because we want to display readonly info here. home can
mean to reset the ui

siosonel:
Another opinion: I think the Cohort tab in TermdbTest is really the About tab, at least the original intent for the cohort tab is to be a landing "tab" or "page". So, the labels and other info for this tab should be customizable for a given dataset, as well as the content un the main section. I'm not suggesting to rewrite the cohort tab completely; I'm only suggesting to make the Cohort tab an SJLife-specific implementation of the "about" tab. If it's difficult to implement dataset-specific tab labels/info, then I'm fine with the current PR.

xzhou82:
yes i agree the Cohort tab totally shows "About" info. i haven't thought about it. Colleen could you evaluate if it's feasible to make existing "cohortTab" to show if: 1) ds uses subcohort then the tab will be called COHORT 2) ds has "about" and the tab should be called ABOUT

siosonel:
I think make the label part of the option, so maybe like this:
```js
navTabs: [{
  label: 'Cohort' // or "About' or something else, depending on the dataset, shown in the middle of the tab
  navTop?: string | ()=>string // function to get what to display at the nav top
  navBtm?: string | ()=>string // function to get what to display at the nav top
  subheader?: string[] // list of subheader labels that can be toggled
  content: string/html | (subheaderLabel)=>string|html // function to get the contents based on the subheader label
}]
```

creilly8:
> ubcohort then the tab w

@xzhou82. Sure this shouldn't be an issue.

creilly8:
> I think make the label part of the option, so maybe like this:
> 
> ```ts
> extraNavTabs: [{
>   order: (tabArray) => number // flexible to allow an index number position, or to look for a certain tab before or after this tab, maybe this function should already do the array.splice() to mutate tabArray???
>   label: 'About' // or "Cohort' or something else, depending on the dataset, shown in the middle of the tab
>   navTop?: string | appState => string // function to get what to display at the nav top, will be hidden if option is not supplied
>   navBtm?: string | appState => string // function to get what to display at the nav bottom, will be hidden if option is not supplied
>   subheader?: string[] // list of subheader labels that can be toggled; the subheader, will be hidden if option is not supplied
>   content: string/html | (subheaderLabel)=>string|html // function to get the contents based on the subheader label
> }]
> ```
> 
> @creilly8 Also one other comment: the content should not use a sandbox, it's a bit unusual to allow closing/hiding the content. The holder should be the "parent" div of the sandbox.

@siosonel, the issue requested the content appear in a sandbox. @xzhou82 I agree with Edgar. Do you agree to show the content not in a sandbox? 

To clarify for termdbtest, should the COHORT tab appear or ABOUT? 

Tangential to this, maybe we should think about a second termdbtest dataset file. What is currently in use is specialized for SJLife rendering but realistically can't test other rendering scenarios, like the above. The second dataset file could use the same db and files but allow for other configurations for testing. That's a broader discussion beyond the scope of this PR but worth considering.

siosonel:
"To clarify for termdbtest, should the COHORT tab appear or ABOUT?" The cohort tab should appear like before, but underneath it's really an about tab with a custom label (Cohort)

airenzp (dismissed):
After checking with Colleen the changes, I see one can add an about section to a dataset to display the tab in the UI, thanks. We should be aware that this section should not be present for datasets with multiple cohorts, as it prevents the app from selecting a cohort and it breaks the UI, besides is not expected to happen, so lets reinforce it with some validation, can be done in a later PR. The datase to test should be one without cohorts, like PNET, for example.

airenzp:
After checking with Colleen the changes, I see one can add an about section to a dataset to display the tab in the UI, thanks. We should be aware that this section should not be present for datasets with multiple cohorts, as it prevents the app from selecting a cohort and it breaks the UI, besides is not expected to happen, so lets reinforce it with some validation, can be done in a later PR. One can test adding this section in a dataset without cohorts, like PNET, for example.

xzhou82 (approved):
works well, thanks

client/mass/about.ts
```diff
@@ -0,0 +1,39 @@
...
+		this.holder = opts.holder
+		this.features = opts.features
+	}
+
+	init() {
+		this.holder.append('div').style('padding', '10px').html(this.features.html)
+	}
+
+	main() {
+		const aboutIdx = this.app.Inner.components.nav.Inner.tabs.findIndex(tab => tab.subheader == 'about')
```
xzhou82, line 33: @creilly8 and all, `Inner` is only present when debugmode flag is set which should only be on dev. it breaks prod now.
as a rule, Inner must not be used in portal code

PR #2098, opened 2024-09-09
Tw refactor.3

siosonel:
## Description

Updated [tw/README.md](https://github.com/stjude/proteinpaint/blob/tw-refactor.3/client/tw/README.md). Now recommends the `addons` option  to add custom methods to a tw subclass, because:
- extending every relevant subclass with custom plot/app methods may be unnecessarily repetitive 
- extending a custom composite base class will lose inheritance of specialized subclass methods that are specific to a term type + q.type/mode

Tested with
- http://localhost:3000/testrun.html?name=*.unit
- http://localhost:3000/testrun.html?name=*.integration
- `cd client && npm run tsc`

## Checklist

### DIFF
--- a/client/tw/categorical.ts
+++ b/client/tw/categorical.ts
@@ -1,6 +1,5 @@
 import {
 	CategoricalTerm,
-	CategoricalQ,
 	ValuesQ,
 	PredefinedGroupSettingQ,
 	CustomGroupSettingQ,
@@ -21,10 +20,12 @@ import { set_hiddenvalues } from '#termsetting'
 export type CatInstance = CatValues | CatPredefinedGS | CatCustomGS
 export type CatTypes = typeof CatValues | typeof CatPredefinedGS | typeof CatCustomGS
 
-export class CategoricalBase {
+export class CategoricalBase extends TwBase {
+	// type, isAtomic, $id are set in ancestor base classes
 	term: CategoricalTerm
 
-	constructor(tw: CatTWTypes, opts) {
+	constructor(tw: CatTWTypes, opts: TwOpts) {
+		super(tw, opts)
 		this.term = tw.term
 	}
 
@@ -114,6 +115,7 @@ export class CategoricalBase {
 }
 
 export class CatValues extends CategoricalBase {
+	// term, type, isAtomic, $id are set in ancestor base classes
 	q: ValuesQ
 	#tw: CatTWValues
 	#opts: TwOpts
@@ -127,8 +129,12 @@ export class CatValues extends CategoricalBase {
 		this.#opts = opts
 	}
 
+	getTw() {
+		return this.#tw
+	}
+
 	// See the relevant comments in the CategoricalBase.fill() function above
-	static fill(tw: RawCatTWValues, opts: TwOpts = {}): CatTWValues {
+	static fill(tw: RawCatTWValues): CatTWValues {
 		if (!tw.type) tw.type = 'CatTWValues'
 		else if (tw.type != 'CatTWValues') throw `expecting tw.type='CatTWValues', got '${tw.type}'`
 
@@ -139,8 +145,8 @@ export class CatValues extends CategoricalBase {
 
 		// GDC or other dataset may allow missing term.values
 		if (!term.values) term.values = {}
-		const numVals = Object.keys(tw.term.values).length
-		// GDC or other dataset may allow empty term.values
+		//const numVals = Object.keys(tw.term.values).length
+		//GDC or other dataset may allow empty term.values
 		//if (!numVals) throw `empty term.values`
 		if (q.mode == 'binary') {
 			if (Object.keys(tw.term.values).length != 2) throw 'term.values must have exactly two keys'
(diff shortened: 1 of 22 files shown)


airenzp (dismissed):
Looks great thanks

airenzp (approved):

client/tw/categorical.ts
```diff
@@ -139,8 +123,8 @@ export class CatValues extends CategoricalBase {
 
 		// GDC or other dataset may allow missing term.values
 		if (!term.values) term.values = {}
-		const numVals = Object.keys(tw.term.values).length
-		// GDC or other dataset may allow empty term.values
+		//const numVals = Object.keys(tw.term.values).length
+		//GDC or other dataset may allow empty term.values
 		//if (!numVals) throw `empty term.values`
 		if (q.mode == 'binary') {
```
gavrielm, line 129: Should not allow `mode='binary'` here. Should only be allowed for custom groupsetting (for categorical term).

When verifying `mode='binary'`, should check `tw.q.customset`, not `tw.term.values`.

Can address in a later PR if you prefer.

gavrielm:
Maybe related to comment above, not sure why the `mode` property is necessary in these term definitions:

```
export type ValuesQ = MinBaseQ & {
	type: 'values'
	mode?: 'binary'
}

export type PredefinedGroupSettingQ = MinBaseQ & {
	type: 'predefined-groupset'
	predefined_groupset_idx: number
	mode?: 'binary'
}

export type CustomGroupSettingQ = MinBaseQ & {
	type: 'custom-groupset'
	customset: BaseGroupSet
	mode?: 'binary'
}
```

client/tw/categorical.ts
```diff
@@ -162,7 +146,7 @@ export class CatValues extends CategoricalBase {
 }
 
 export class CatPredefinedGS extends CategoricalBase {
-	//term: CategoricalTerm
+	// term, type, isAtomic, $id are set in ancestor base classes
 	q: PredefinedGroupSettingQ
 	#groupset: BaseGroupSet
```
gavrielm, line 151: Why is `#groupset: BaseGroupSet` necessary? Term groupsetting should already be in `this.term`.

gavrielm:
In `tw/numeric.ts`:

```
if (preferredBins) {
	const q = opts.defaultQ
	if (tw.term.bins[preferredBins]) {
		tw.q = structuredClone(tw.term.bins[preferredBins])
		delete q.preferredBins
	} else if (preferredBins == 'median') {
		if (q.type != 'custom-bin') throw '.type must be custom-bin when .preferredBins=median'
		Object.assign(tw.q, q)
	} else {
		throw `unrecognized defaultQ.preferredBins='${preferredBins}'`
	}
}
```

It looks like if `tw.term.bins[preferredBins]`, then `defaultQ` will not be merged into `q`

gavrielm:
Hard to follow this code block in `tw/numeric.ts`:

```
if (tw.term.bins) {
	// detect if tw.q still needs to use term.bins
	//
	// TODO: may move some of these logic in the applicable switch-case code block,
	//       which may require not allowwing "mixed-modes", like some code expecting
	//       q.type='regular-bin' to possibly have q.mode='binary' | 'continuous',
	//       should clean or clarify such usecases
	//
	...
```
Will be good to do the TODO mentioned in the comment to help clarify the logic.

client/tw/categorical.ts
siosonel: setting a `groupset` instance property would allow abstraction of what groupset refers to, between predefined-groupset and custom-groupset. In this way, the same addon method can refer to just `this.groupset` without needing to test if it's a CatPredefinedGS or CatCustomGS. However, in order for an addon method to access the `groupset` property, it cannot be private (with a hash prefix), I'll need to change to a non-enumerable `groupset` property (I'll do it in this PR).

siosonel:
> In `tw/numeric.ts`:
> 
> ```
> if (preferredBins) {
> 	const q = opts.defaultQ
> 	if (tw.term.bins[preferredBins]) {
> 		tw.q = structuredClone(tw.term.bins[preferredBins])
> 		delete q.preferredBins
> 	} else if (preferredBins == 'median') {
> 		if (q.type != 'custom-bin') throw '.type must be custom-bin when .preferredBins=median'
> 		Object.assign(tw.q, q)
> 	} else {
> 		throw `unrecognized defaultQ.preferredBins='${preferredBins}'`
> 	}
> }
> ```
> 
> It looks like if `tw.term.bins[preferredBins]`, then `defaultQ` will not be merged into `q`

Yes, the code assumes that `defaultQ.preferredBins` indicates to only consider `tw.term.bins[preferredBins]`, that `defaultQ` has no other relevant properties to merge into `tw.q`. But if that is a wrong assumption, should `opts.defaultQ` **always** be merged to `tw.q`, before doing anything else? But that would mean that the `tw.term.bins.default | less` that's merged later would replace matching property keys in "preferred bins". Or maybe, set tw.q to the preferred bins first, then apply the defaultQ overrides? So, would need to resolve the order in which to apply the overrides versus preferred bins.

gavrielm (approved):

PR #2109, opened 2024-09-12
Added profile buttons to charts if profile dataset loaded

airenzp:
## Checklist


client/mass/charts.js
```diff
@@ -254,6 +255,10 @@ function getChartTypeList(self, state) {
 			}
 		}
 	]
+	if (state.vocab.dslabel.includes('Profile')) {
```
xzhou82, line 258: this is too arbitrary, shouldn't do it. just add all plot types in `buttons[]` and it's fine

airenzp: Should be solved now, please check in last commit if this is what you meant

PR #2110, opened 2024-09-12
Tentative implementation  to query single cell data from HDF5 files

robinpaul85:
## Description

This branch includes a new implementation for querying single cell from HDF5 files using Rust. 

Following points before compiling the rust code.

1) Needs a recent version of rust compiler 1.80. To check current cargo version run `cargo --version`. If the result is less than 1.80, it needs to be updated. Run `rustup update`.

2) The rust HDF5 [crate](https://lib.rs/crates/hdf5-metno) wraps around hdf5 library. https://formulae.brew.sh/formula/hdf5. I had to install HDF5 on Mac using brew using the command `brew install HDF5`. For Debian linux, may need to install https://packages.debian.org/bullseye/libhdf5-dev .

3) rust/Cargo.lock - This file is currently git ignored in master. However, I was forced to commit this file because the rust HDF5 crate optionally supports two different version of the `ndarray` crate. As a result different version of ndarray were getting installed in different devices and was causing downstream compilation errors. Commiting a particular version of `ndarray` in Cargo.lock solved the problem. I had to force commit this file. Please suggest if this is the best way to go about it. 

## Checklist


server/shared/descriptive.stats.js
```diff
@@ -55,3 +54,9 @@ export default function summaryStats(array) {
 		]
 	}
 }
+
+function computePercentile(values, percentile) {
+	const index = Math.abs((percentile / 100) * values.length - 1)
+	const value = Number.isInteger(index) ? (values[index] + values[index + 1]) / 2 : values[Math.ceil(index)]
+	return value
+}
```
gavrielm, line 62: It would be better to import `computePercentile()` from `server/shared/compute.percentile.js` instead of duplicating it here. To prevent repeatedly sorting the array, you can add a `isSorted` parameter to the function. For example:
```
export default function computePercentile(values, percentile, isSorted) {
	if (!isSorted) values.sort((a, b) => a - b)
	const index = Math.abs((percentile / 100) * values.length - 1)
	const value = Number.isInteger(index) ? (values[index] + values[index + 1]) / 2 : values[Math.ceil(index)]
	return value
}
```

If you know that the input array is already sorted then you can set `isSorted=true`.

server/shared/descriptive.stats.js
```diff
@@ -20,24 +21,22 @@ export default function summaryStats(array) {
 	}
 
 	//compute median
-	const median = computePercentile(arr, 50)
-
-	const squareDiffs = arr.map(x => (x - mean(arr)) ** 2).reduce((a, b) => a + b, 0)
+	const median = computePercentile(sorted_arr, 50)
+	const mean_arr = mean(sorted_arr)
```
gavrielm, line 25: Can use `mean_arr` in line 48 instead of recomputing it

PR #2111, opened 2024-09-12
Fixed tooltip issues

airenzp:
## Description

Fixes tooltip size and adds fade out effect.

## Checklist

### DIFF
--- a/client/src/style.css
+++ b/client/src/style.css
@@ -1275,10 +1275,20 @@ sandbox header
     transition: opacity 0.5s 0.25s; /* 👈 Change the time to meet your requirements. */;
     z-index: 1000;
     width: max-content;
-    max-width: 300px;
+    max-width: 200px;
     white-space: wrap;
 }
 
   .sja_root_holder [aria-label]:hover:after, .sja_menu_div [aria-label]:hover:after, .sja_pane [aria-label]:hover:after {
-    opacity: 1;
+    animation-name: fadeOut;
+    animation-duration: 5s;
+  }
+
+  @keyframes fadeOut {
+    0% {
+      opacity: 1;
+    }
+    100% {
+      opacity: 0;
+    }
   }
--- a/client/src/style.css
+++ b/client/src/style.css
@@ -1289,6 +1289,9 @@ sandbox header
     0% {
       opacity: 1;
     }
+    70% {
+      opacity: 1;
+    }
     100% {
       opacity: 0;
     }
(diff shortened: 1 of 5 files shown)


siosonel:
the tooltip automatically fades out even though I may not have finished reading it, I think it should only disappear after moving the mouse out of the described/aria-labeled element

client/src/style.css
```diff
@@ -1274,8 +1274,21 @@ sandbox header
     opacity: 0;
     transition: opacity 0.5s 0.25s; /* 👈 Change the time to meet your requirements. */;
     z-index: 1000;
+    width: max-content;
+    max-width: 200px;
+    white-space: wrap;
```
siosonel, line 1279: suggest adding  `pointer-events: none;` here, so that the tooltip does not block mouse clicks on termsetting pills or other inputs that may be blocked by the tooltip text.

airenzp:
> the tooltip automatically fades out even though I may not have finished reading it, I think it should only disappear after moving the mouse out of the described/aria-labeled element

We could increase the time to fade out a bit more. Keeping the tooltip on hover causes the tooltip to cover options in the controls, when we are trying to fill a value, like here:

[
Screenshot 2024-09-12 at 10 18 57 PM
](url)

client/src/style.css
airenzp: Added but it does not make it disappear when hovering over nearby elements

siosonel: The tooltip disappears for me, but I also had to restore the css style to not use the fadeout:

https://github.com/user-attachments/assets/4277678b-d5cb-46ce-9687-b5fa450b9e21`

```css
.sja_root_holder [aria-label]:after, .sja_menu_div [aria-label]:after, .sja_pane [aria-label]:after {
    content: attr(aria-label);
    position: absolute;
    left: 1em;
    top: 1.25em; /* put it on the top */
    /*bottom: 50%;*/
    background-color: #aaa;
    border-radius: 6px ;
    font-size: 15px;
    color: #fff;
    padding: 5px;
    opacity: 0;
    transition: opacity 0.5s 0.25s; /* 👈 Change the time to meet your requirements. */;
    z-index: 1000;
    width: max-content;
    max-width: 200px;
    white-space: wrap;
    pointer-events: none;
  }

  .sja_root_holder [aria-label]:hover:after, .sja_menu_div [aria-label]:hover:after, .sja_pane [aria-label]:hover:after {
    opacity: 1;
  }
```

airenzp:
It disappears for that case, but not here. Anyways, we can talk about it tomorrow to find the best solution.
https://github.com/user-attachments/assets/84f1fc65-15e8-4889-b007-60f0488e7815

xzhou82 (dismissed):
i like it, now the tooltip no longer blocks nearby elements. thanks

gavrielm (approved):
Thanks for the fix!

siosonel (approved):

PR #2116, opened 2024-09-13
fix missing setCellProps() for condition terms in matrix plot

siosonel:
## Description

Added a condition term to dictionary terms test in [matrix.integration.spec.js](http://localhost:3000/testrun.html?dir=mass&name=matrix.integration). Not certain when this bug was introduced, the extended tw logic in `matrix.serieses.js` does not affect the selection of `setCellProps()` for non-migrated term types, it looks like condition-terms were broken even before the tw refactor.

Tested with http://localhost:3000/testrun.html?name=*.unit and `cd client && npm run test:integration`

## Checklist


airenzp (approved):
Works thanks

PR #2136, opened 2024-09-18
fix: genome browser plot now saves mds3 subtracks in state and recove…

xzhou82:
…r from session

## Description

test [link](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22MB_meta_analysis%22}). at genome browser app, launch kbtbd4, at mds3 tk, click Samples and create a subtk. save/share a session and restore it, the subtk will be present. this is good enough to generate mbmeta figure links

all mds3 tk tests pass

- order of tracks are not saved
- skewer config (up/down pointing, shape, numeric mode) are not saved and can be worked on next

## Checklist

### DIFF
--- a/client/plots/genomeBrowser.js
+++ b/client/plots/genomeBrowser.js
@@ -4,6 +4,7 @@ import { Menu } from '#dom/menu'
 import { sayerror } from '../dom/sayerror.ts'
 import { dofetch3 } from '#common/dofetch'
 import { getNormalRoot } from '#filter/filter'
+import { filterJoin } from '#shared/filter'
 import { first_genetrack_tolist } from '#common/1stGenetk'
 import { gbControlsInit, mayUpdateGroupTestMethodsIdx } from './genomeBrowser.controls'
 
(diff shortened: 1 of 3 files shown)


karishma-gangwani (approved):
works for me. Thank you @xzhou82

client/plots/genomeBrowser.js
```diff
@@ -352,6 +374,25 @@ class genomeBrowser {
...
+
+	async maySaveMds3SubtkToState(bb) {
+		/* arg is block instance
+		when a mds3 subtk is created/updated, its tk.filterObj should be saved to state so it can be recovered from session
+		*/
+		const config = structuredClone(this.state.config)
+		config.subTkFilters = []
+		for (const t of bb.tklst) {
+			if (t.filterObj) {
+				config.subTkFilters.push(t.filterObj)
```
creilly8, line 386: We might consider a `global` and `local` filters in the filterObj so users can reuse filter like this in other plots.

creilly8 (approved):

client/plots/genomeBrowser.js
xzhou82: good idea. can introduce `tk.globalFilter`

PR #2161, opened 2024-09-24
Add survival time cut off

congyu-lu:
## Description

Test with sjpp: mbmeta_add_survival_cutoff

Only mbmeta has a default survival cutoff of 10 years

## Checklist

### DIFF
--- a/client/plots/survival.js
+++ b/client/plots/survival.js
@@ -149,6 +149,13 @@ class TdbSurvival {
 							]
 						},
 						{
+							label: 'Survival Time Cut-Off',
+							type: 'number',
+							chartType: 'survival',
+							settingsKey: 'maxSurTime',
+							title: 'The max survival time to be displayed in this plot'
+						},
+						{
 							label: 'Time Factor',
 							type: 'math',
 							chartType: 'survival',
@@ -249,7 +256,8 @@ class TdbSurvival {
 		const opts = {
 			chartType: 'survival',
 			term: c.term,
-			filter: this.state.termfilter.filter
+			filter: this.state.termfilter.filter,
+			maxSurTime: this.state.config.settings.maxSurTime
 		}
 		if (c.term2) opts.term2 = c.term2
 		if (c.term0) opts.term0 = c.term0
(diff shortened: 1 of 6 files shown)


server/src/termdb.survival.js
```diff
@@ -96,7 +96,11 @@ export async function get_survival(q, ds) {
 
 		// perform survival analysis for each chart
 		for (const chartId in byChartSeries) {
-			const data = byChartSeries[chartId]
+			let data = byChartSeries[chartId]
+			if (q.maxSurTime) {
+				// Filter out the survival data with Time to event longer than survial time cut-off (defined in dataset)
```
siosonel, line 101: @congyu-lu The comment here should be `time-to-event that is less than ....`, not "longer than"?

client/plots/survival.js
```diff
@@ -148,6 +148,13 @@ class TdbSurvival {
 								{ label: 'Tick', value: 'vtick' }
 							]
 						},
+						{
+							label: 'Survival Time Cut-Off',
+							type: 'number',
+							chartType: 'survival',
+							settingsKey: 'maxSurTime',
+							title: 'The max survival time to be displayed in this plot'
```
siosonel, line 156: @congyu-lu The title should also say that if left empty or `0`, would default to not having a max time-to-event cutoff.

client/plots/survival.js
```diff
@@ -148,6 +148,13 @@ class TdbSurvival {
 								{ label: 'Tick', value: 'vtick' }
 							]
 						},
+						{
+							label: 'Survival Time Cut-Off',
+							type: 'number',
+							chartType: 'survival',
+							settingsKey: 'maxSurTime',
```
siosonel, line 155: @congyu-lu I prefer the config/setting key name to be `maxTimeToEvent`, since `maxSurTime` can mean only those who did not have an exit event. But this preference is not critical.

server/src/termdb.survival.js
congyu-lu: filter out the time-to-event, so used longer

client/plots/survival.js
congyu-lu: thanks just changed

client/plots/survival.js
congyu-lu: thanks this is a good point, just changed it

siosonel (approved):

server/shared/types/dataset.ts
```diff
@@ -777,6 +777,10 @@ type SortPriorityEntry = {
 	tiebreakers: TieBreakersEntry[]
 }
 
+type SurvivalSettings = {
+	maxTimeToEvent?: number
```
xzhou82, line 781: add comment for purpose

xzhou82 (approved):

PR #2169, opened 2024-09-27
Add docker dev container

aacic:
## Description

This PR adds a docker dev container, which enables front-end and backed nodejs, R, and Python development. Rust development is not supported yet. 

The purpose of the dev container is to enable the fast setup of the development environment. 

Testing:

There are 2 options to do container development:

1. Using text editors.
```
cd ~/dev
mkdir ppdev
cd ppdev
git clone git@github.com:stjude/proteinpaint.git 
cd  proteinpaint
git checkout devContainers 
cd container/dev
./run.sh "/path/to/ppdev/proteinpaint/"
```

This will run the container, the container will run:

```
npm install
npm build
npm run dev1

```

The project folder will be mounted inside of the container, and any changes to the source code outside will be re-bundeled inside of the container and the changes reflected on:
`http://localhost:3000/`

2. VS Code development
Execute:
```
cd ~/dev
mkdir vscode
cd vscode 
git clone git@github.com:stjude/proteinpaint.git 
cd  proteinpaint
git checkout devContainers 
```

Open VS Code
Install Dev Containers extension 
Open project inside of VS code
Click on Reopen project in Dev Container
Open terminal from VS code
the console output should be:
root@container_id:/home/root/proteinpaint# 

Run:
```
npm install
npm run build
cp container/dev/serverconfig.json . 
npm run dev1
```

All code changes should be reflected on:
`http://localhost:3000/`

 Note: when running the dev container, the `node_modules` folders will appear inside the source project directory. The project will not work outside of the container if the host and target architecture are not the same and there are differences in installed dependencies. 

## Checklist


siosonel:
I suggest moving the dev container code, configs to `proteinpaint/container/dev` dir, instead of having `.devcontainer` dir, and also deleting `build/dev` dir.

aacic:
> I suggest moving the dev container code, configs to `proteinpaint/container/dev` dir, instead of having `.devcontainer` dir, and also deleting `build/dev` dir.

The `.devcontainer`  dir is used by VSCode IDE and Webstorm IDE, it's a standard: https://containers.dev/implementors/spec/

I've migrated the dev container logic from `build/dev` to the `proteinpaint/container/dev`, but I wanted to keep `build/dev` for history purposes until we ensure the new dev container works properly. I can delete it if you insist, it's in the git history. @siosonel

siosonel:
> > I suggest moving the dev container code, configs to `proteinpaint/container/dev` dir, instead of having `.devcontainer` dir, and also deleting `build/dev` dir.
> 
> The `.devcontainer` dir is used by VSCode IDE and Webstorm IDE, it's a standard: https://containers.dev/implementors/spec/
> 
> I've migrated the dev container logic from `build/dev` to the `proteinpaint/container/dev`, but I wanted to keep `build/dev` for history purposes until we ensure the new dev container works properly. I can delete it if you insist, it's in the git history. @siosonel

Ok, with regards to using `.devcontainer`. And for sure, we should get rid of `build/dev`, the readme in there should also be migrated to `container/dev`. We should try to group related code in the same directory if possible. 

What's the difference between `.devcontainer/Dockerfile` and `container/dev/Dockefile`?

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '3'
+
+services:
+  app:
+    build:
+      context: .
+      dockerfile: Dockerfile
```
siosonel, line 7: @aacic Since the `.devcontainer/Dockerfile` only has one line to use the base dev image, should this line just be `image: ....` instead?

.github/workflows/CD-publish-dev-container.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Login to Github
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.PAT }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
```
siosonel, line 43: What is `QEMU` for? To be able to run x86/intel-built image in ARM machines?

container/dev/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ARG IMGVER=""
+
+LABEL org.opencontainers.image.source="https://github.com/stjude/proteinpaint" \
+    org.opencontainers.image.revision="$IMGREV" \
+    org.opencontainers.image.version="$IMGVER"
+
+WORKDIR /home/root/proteinpaint/
+
+EXPOSE 3000
+CMD ["sh", "-c", "npm install && npm run build && npm run dev1"]
```
siosonel, line 13: Shouldn't `npm install && npm run build` already be done in a `RUN` step before `CMD`? Or is there a reason for running it as part of `CMD`?

container/dev/run.sh
```diff
@@ -0,0 +1,40 @@
...
+
+
+if [ $# -eq 0 ]; then
+  echo "Error: No host project folder argument supplied." >&2
+  exit 1
+fi
+
+PROJECT_FOLDER=$1
+
+CONTAINER_NAME="dev-container"
```
siosonel, line 15: Please rename to `ppdev`:
- there may be other "dev" containers/process (like GFF), important the container name is associate with `pp`
- `-container` in the name is unnecessary, for sure we are dealing with a docker container here

.github/workflows/CD-publish-dev-container.yml
siosonel: Related to my question above: If it's possible to use a native arm64 build, it would be preferable than using QEMU, as mentioned in https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/. So, a user would need to use the deps/base image that's appropriate to their machine CPU  architecture. Not sure if this option has already been explored/researched for our dev use case.

.devcontainer/docker-compose.yml
aacic: I added the Docker file so someone could extend it if needed. But, let's keep it simple. I've replaced this with the image tag.

container/dev/Dockerfile
siosonel: On second thought, maybe this is done in case someone installs a new npm dependency or updates an existing dep in a branch, and the majority of the node_modules can be reused on every docker start. But not 100% sure. Maybe can leave as-is and investigate for actual use in regular, day-to-day sjpp dev.

.github/workflows/CD-publish-dev-container.yml
aacic: You are right, it works without QEMU. At one point it wasn't working, since I got an error that was ARM-related, but now it has worked out. I've built the images again.

container/dev/Dockerfile
aacic: The RUN command is used to execute commands during the image build process, and at that time the source is not attached yet. That's why I used CMD.

container/dev/run.sh
aacic: Changed.

siosonel (approved):

airenzp:
Sorry I could not test this PR on time.  I will test to set up my dev env with it though. Thanks to all.

PR #2202, opened 2024-10-07
Fixed precision in DE table, delete DE table when DE parameters change

robinpaul85:
## Description
Fixed precision in DE table, delete DE table when DE parameters change

## Checklist

### DIFF
--- a/client/plots/DEanalysis.js
+++ b/client/plots/DEanalysis.js
@@ -241,6 +241,7 @@ add:
 	const holder = self.dom.holder
 	holder.selectAll('*').remove()
 	self.dom.detailsDiv.selectAll('*').remove()
+	self.dom.tableDiv.selectAll('*').remove()
 	let minlogfc = 0,
 		maxlogfc = 0,
 		minlogpv = 0,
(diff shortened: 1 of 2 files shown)


karishma-gangwani (commented):
is it possible to use the scientific notation for the p-values that have large decimal places? @robinpaul85

robinpaul85:
> is it possible to use the scientific notation for the p-values that have large decimal places? @robinpaul85

const a=0.0000000000000000001
a.toPrecision(4) yields  "1.000e-19"
const b=0.1000000000000000001
b.toPrecision(4) yields "0.1000"

airenzp:
Also, better to determine the precision based on the number, that is why I mentioned this morning the method roundValueAuto. If you set it to 4 digits you assume that for any number 4 digits is the precision to use. Is that true for all cases? We could also customize roundValueAuto to return scientific notation if required.

airenzp:
Looking at the table I noticed that the scientific notation is already used by default. See

Screenshot 2024-10-07 at 6 46 18 PM

xzhou82 (approved):
i noticed for BP geneset gsea now returns all 6K genesets. the backend should be changed to return 1K genesets at max

PR #2209, opened 2024-10-08
fix and improve git hook scripts 

siosonel:
## Description

Include more file extensions to format, lint, detect secrets as part of the pre-commit and pre-push hooks.

Tested with:
- `pre-commit run --all-files`: to make sure there are no undetected secrets in all files in latest commit
- create a test branch to trigger the commit hooks: (1) search for `pragma: allowlist secret` and remove that string from a few .html, .sh and other non-js files. The secret detection should run on all files, previously the pre-commit hook would only run on js/ts files. Note that the pre-push hook was already checking all modified, added, copied files. 

## Checklist

### DIFF
--- a/utils/hooks/pre-commit
+++ b/utils/hooks/pre-commit
@@ -2,34 +2,39 @@
 
 set -e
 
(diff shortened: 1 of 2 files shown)


gavrielm (approved):
Works well. Thanks

PR #2234, opened 2024-10-11
Profile change cohort fixes

airenzp:
## Description

Dispatched app_refresh with subactions including clearing the filter. See corresponding PR in [sjpp](https://github.com/stjude/sjpp/pull/524) 

## Checklist

### DIFF
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -570,7 +570,7 @@ function setRenderers(self) {
 						const state = self.app.getState()
 						//the terms used in the plots are not always the same for the profile.
 						//Therefore when switching cohorts, it is necessary to delete the plots opened
-						if (state.vocab.dslabel == 'profile') {
+						if (state.termdbConfig.selectCohort.clearOnChange) {
 							const subactions = [
 								{
 									type: 'filter_replace',
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -568,8 +568,6 @@ function setRenderers(self) {
 					.style('margin-left', '0px')
 					.on('click', async () => {
 						const state = self.app.getState()
-						//The profile has clearOnChange. The terms used in the plots are not always the same for the profile.
-						//Therefore when switching cohorts, it is necessary to delete the plots opened and the global filter
 						const clearOnChange = state.termdbConfig.selectCohort.clearOnChange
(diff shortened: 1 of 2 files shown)


client/mass/nav.js
```diff
@@ -570,14 +570,32 @@ function setRenderers(self) {
 						const state = self.app.getState()
 						//the terms used in the plots are not always the same for the profile.
 						//Therefore when switching cohorts, it is necessary to delete the plots opened
-						if (state.vocab.dslabel == 'profile')
+						if (state.vocab.dslabel == 'profile') {
```
siosonel, line 573: we should not hardcode to a specific dslabel here, maybe do this instead, so need to add a new boolean flag to `sjpp/dataset/sjglobal.profile.ts: cohort.termdb.selectCohort`
```js
if (self.state.termdbConfig.selectCohort.clearFilterOnChange) {
```

airenzp: maybe clearStateOnChange, as it also removes the plots?

siosonel: I changed my suggestion to an array, so a dataset may want to clear the filter but not delete the existing plots. So, profile may want to clear both `['plots', 'filter']`, but another dataset may only want to clear one of them or maybe another part of the state, this approach will be more flexible than a boolean.

shared/types/src/dataset.ts
```diff
@@ -690,6 +690,7 @@ type SelectCohortEntry = {
 	values: SelectCohortValuesEntry[]
 	description?: string
 	asterisk?: string
+	clearOnChange?: { [index: string]: boolean }
```
xzhou82, line 693: thanks. please move the comments here to explain this property

client/mass/nav.js
airenzp: Solved

shared/types/src/dataset.ts
airenzp: Done

siosonel: todo later: it would be better to define specific names for the allowed clearOnChange property names, so that invalid props would be typechecked, but this is  not urgent, can discuss and do later

```ts
clearOnChange?: { plots?: true, filter?: true }
```

siosonel (approved):

Issue #2226, opened 2024-10-10
Code updated date same as server relaunch date


creilly8:
In the app header, the `Code updated: [Date]` is the same as when the server relaunches, not when the code is updated. If the server relaunches because of a network failure, hardware failure, etc. for example, the homepage displays the same date as the server restart even though the code wasn't updated.

creilly8:
Decided during team meeting to replace `Code updated` with the release version.

PR #2238, opened 2024-10-14
Show the release version in the app header with github link

creilly8:
## Description
Closes issue #2226 

If this is the desired replacement text and functionality, then I can add something similar to the mass UI. 
Test: Open any local page with the header -> http://localhost:3000. 

## Checklist

### DIFF
--- a/client/src/header/AppHeader.ts
+++ b/client/src/header/AppHeader.ts
@@ -128,16 +128,16 @@ export class AppHeader {
 		{
 			// a row for server stats
 			const row = headinfo.append('div').style('padding-left', '15px')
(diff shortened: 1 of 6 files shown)


siosonel (approved):

siosonel:
I kept the code date but reformatted to include build time in hours:minutes - it doesn't hurt to includes the build date-time in the healthcheck and genomes response payload even if it's not displayed. 

I also changed the version link to the github container release page, it's much more relevant than the github release page which we do not advertise or use, since the tarball in there is not considered built and not functional without the system dependencies. The container version includes a hash, which is the second word in `public/rev.txt` inside the container.

PR #2240, opened 2024-10-15
Profile user

airenzp:
## Description

Handled different user views. Hided our plots. Fixed radio buttons display in controls. Added delete all icon. Misc fixes and cleanup. See corresponding PR in [sjpp](https://github.com/stjude/sjpp/pull/528). Addresses most of the requests from Miriam and Heather. Please use latest db from hpc. Test different user views with links: 

public: http://localhost:3000/profile/index.html
admin: http://localhost:3000/profile/index.html?logged=1
site: http://localhost:3000/profile/index.html?logged=1&site=PRO_00020

## Checklist

### DIFF
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -231,7 +231,7 @@ function setRenderers(self) {
 			.style('display', 'inline-block')
 			.style('padding-top', '4px')
 			.style('vertical-align', 'middle')
-		icon_functions['trash'](deleteAllDiv, { handler: self.deletePlots, title: 'Delete all' })
+		icon_functions['trash'](deleteAllDiv, { handler: self.deletePlots, title: 'Delete all plots' })
 
 		if (appState.nav.header_mode == 'only_buttons') {
 			self.dom.tabDiv.style('display', 'none')
@@ -381,7 +381,10 @@ function setRenderers(self) {
 
 	self.deletePlots = () => {
 		const state = self.app.getState()
-		for (const plot of state.plots) self.app.dispatch({ type: 'plot_delete', id: plot.id })
+		const subactions = []
+		for (const plot of state.plots) subactions.push({ type: 'plot_delete', id: plot.id })
+
+		self.app.dispatch({ type: 'app_refresh', subactions })
 	}
 
 	self.mayShowMessage_sessionDaysLeft = () => {
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -10,6 +10,7 @@ import { dofetch3 } from '#common/dofetch'
 import { Menu } from '#dom/menu'
 import { getFilterItemByTag, filterRxCompInit } from '#filter/filter'
 import { renderTable } from '../dom/table'
+import { getProfileLogin } from '../plots/profilePlot.js'
 
 /*
 nav {}
@@ -524,11 +525,12 @@ function setRenderers(self) {
 			self.dom.cohortTitle = self.dom.subheader.about.append('h2').style('margin-left', '10px').text(selectCohort.title)
 		}
 
-		if (selectCohort.description) {
-			self.dom.cohortDescription = self.dom.subheader.about
-				.append('div')
-				.style('margin-left', '10px')
-				.html(selectCohort.description)
+		if (selectCohort.description || selectCohort.descriptionByUser) {
+			//temporary logic to get the description until the login is implemented
+			const [logged, site, user] = getProfileLogin()
+			const descriptionByUser = selectCohort.descriptionByUser[user]
+			const description = selectCohort.description || descriptionByUser
+			self.dom.cohortDescription = self.dom.subheader.about.append('div').style('margin-left', '10px').html(description)
 		}
 
 		if (selectCohort.prompt) {
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -11,6 +11,7 @@ import { Menu } from '#dom/menu'
 import { getFilterItemByTag, filterRxCompInit } from '#filter/filter'
 import { renderTable } from '../dom/table'
 import { getProfileLogin } from '../plots/profilePlot.js'
+import { icons as icon_functions } from '#dom'
 
 /*
 nav {}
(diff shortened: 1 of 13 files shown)


shared/types/src/dataset.ts
```diff
@@ -770,6 +770,7 @@ type SelectCohortEntry = {
 	values: SelectCohortValuesEntry[]
 	description?: string
 	/** subtext shown at the very bottom of the cohort/about tab subheader */
+	descriptionByUser?: { [index: string]: string }
```
xzhou82, line 773: the implementation looks good. please fully document it even if it may refactored later. the `subtext shown at the very bottom of the cohort/about tab subheader` is for the `asterisk` property

airenzp: I am not sure what you mean with the subtext. I dont show anything at the bottom.

client/plots/controls.config.js
```diff
@@ -418,7 +418,8 @@ function setRadioInput(opts) {
 				}
 		  ]
 
-	const styles = opts.styles
+	const styles = opts.styles || {}
+	styles['display'] = inputs.length > 2 ? 'block' : 'inline-block'
```
siosonel, line 422: This style change could affect a lot of other plot control menus, should verify the radio input displays still looks the same for other plots. Or, just specify `'opts.styles['display'] = 'block' | 'inline-block'` in your plot controls menu, so as not to affect other plots.

client/plots/profilePlot.js
```diff
@@ -24,13 +24,14 @@ export class profilePlot {
...
 		if (!config) throw `No plot with id='${this.id}' found`
+		const [logged, site, user] = getProfileLogin() //later on replace with real login info
 		return {
 			config,
 			termfilter: appState.termfilter,
 			dslabel: appState.vocab.dslabel,
 			vocab: appState.vocab,
-			logged: true, //later change to read login info
-			site: config.site,
+			logged: logged, //later change to read login info
```
siosonel, line 33: Could use `logged,` syntax shortcut, no need for `logged: logged,`, similar to using the `site,` shortcut.

client/plots/profilePlot.js
```diff
@@ -630,3 +629,12 @@ export function getDefaultProfilePlotSettings() {
...
 	}
 }
+
+export function getProfileLogin() {
+	const queryString = window.location.search
+	const urlParams = new URLSearchParams(queryString)
+	const logged = urlParams.get('logged')
+	const site = urlParams.get('site')
+	const user = logged && site ? 'site' : logged ? 'admin' : 'public'
+	return [logged, site, user]
```
siosonel, line 639: I like this function. Later on when the actual sign-in mechanism is available, it would make is easy to replace the code here to use `app.vocabApi()` to get the logged-in status/role.

client/plots/profilePlot.js
airenzp: Done

client/plots/profilePlot.js
siosonel: Also, could do this later, not urgent: not having a user should be equivalent to not being logged in, I don't think there is a need for a separate `logged` variable.

airenzp: Right now _public_ is a user. Yeah we can review this later.

client/mass/nav.js
```diff
@@ -224,6 +226,12 @@ function setRenderers(self) {
 			titleDiv,
 			tip: new Menu({ padding: '5px' })
 		}
+		const deleteAllDiv = controlsDiv
+			.append('div')
+			.style('display', 'inline-block')
+			.style('padding-top', '4px')
+			.style('vertical-align', 'middle')
+		icon_functions['trash'](deleteAllDiv, { handler: self.deletePlots, title: 'Delete all' })
```
siosonel, line 234: It's not clear that this applies to plots/charts/graphs, should title as `Delete all plots`. Otherwise, it may be interpreted as delete all filters or cohorts when those tabs are open.

client/mass/nav.js
```diff
@@ -371,6 +379,11 @@ function setRenderers(self) {
 		}
 	}
 
+	self.deletePlots = () => {
+		const state = self.app.getState()
+		for (const plot of state.plots) self.app.dispatch({ type: 'plot_delete', id: plot.id })
```
siosonel, line 384: Should use subactions here to optimize not having multiple dispatches, and only have one store state write step: `{type: 'app_refresh', subactions: plots.map(plot => ({type: 'plot_delete', id: plot.id}))`.

client/mass/nav.js
airenzp: yeah, just profile calls the plots graphs, I was trying to use something that works for all

client/mass/nav.js
airenzp: ok

client/mass/nav.js
```diff
@@ -223,6 +226,12 @@ function setRenderers(self) {
 			titleDiv,
 			tip: new Menu({ padding: '5px' })
 		}
+		const deleteAllDiv = controlsDiv
+			.append('div')
+			.style('display', 'inline-block')
+			.style('padding-top', '4px')
+			.style('vertical-align', 'middle')
+		icon_functions['trash'](deleteAllDiv, { handler: self.deletePlots, title: 'Delete all plots' })
```
xzhou82, line 234: this button is kind of dangerous. better to show a confirmation on clicking to avoid oops

xzhou82 (approved):
1. add confirmation to deletAllPlots()
2. when calling plot config, allow to specify certain radio options to show inline when needed or confident that it will be fine, as a way to limit menu height

client/mass/nav.js
siosonel: I also initially thought it's dangerous, but the undo button could easily recover the deleted plots.

client/plots/profilePlot.js
```diff
@@ -205,6 +204,7 @@ export class profilePlot {
 				type: 'radio',
 				chartType,
 				settingsKey: 'isAggregate',
+				styles: { display: 'inline-block' },
```
airenzp, line 207: The option is passed here @xzhou82 if this is what you meant, so it looks like this:

Screenshot 2024-10-15 at 1 24 49 PM

client/mass/nav.js
xzhou82: yes good point. then just modify the button `title` by adding `To revert, click Undo button`
no need to add the confirmation menu

PR #2264, opened 2024-10-16
When searching a term do not include hidden terms

airenzp:
## Checklist


xzhou82 (approved):

server/src/termdb.server.init.js
```diff
@@ -348,6 +348,7 @@ export function server_init_db_queries(ds) {
 			if (tmp) {
 				const r = []
 				for (const i of tmp) {
+					if (i.parent_id == '*') continue
```
siosonel, line 351: probably better to have modified the sql query to `WHERE name LIKE ? and t.parent_id != '*'`, instead of including in the query results only to be filtered out

PR #2284, opened 2024-10-18
updating build scripts

perezthedev:
## Checklist

### DIFF
--- a/server/build.sh
+++ b/server/build.sh
@@ -12,4 +12,6 @@ npx esbuild src/app.ts --bundle --platform=node --packages=external --format=esm
 
 sed -i.bk 's|clinvar.ts|clinvar.js|g' dataset/clinvar.hg19.js
 sed -i.bk 's|clinvar.ts|clinvar.js|g' dataset/clinvar.hg38.js
-rm dataset/*.bk
+
+# Non-critical cleanup step, allow it to fail silently
+rm -f dataset/*.bk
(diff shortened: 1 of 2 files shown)


server/build.sh
```diff
@@ -12,4 +12,6 @@ npx esbuild src/app.ts --bundle --platform=node --packages=external --format=esm
 
 sed -i.bk 's|clinvar.ts|clinvar.js|g' dataset/clinvar.hg19.js
 sed -i.bk 's|clinvar.ts|clinvar.js|g' dataset/clinvar.hg38.js
-rm dataset/*.bk
+
+# Non-critical cleanup step, allow it to fail silently
+rm -f dataset/*.bk
```
siosonel, line 17: `sed` will for sure create the `*.bk` file, no need to use `-f` option to `rm`

server/dockrun.sh
```diff
@@ -17,7 +17,7 @@ elif (($# == 3)); then
 	SRCDIR=$1
 	HOSTPORT=$2
 	IMAGE_NAME=$3
-	CONTAINER_ID=pp
+	CONTAINER_ID=${4:-pp} #defaults to 'pp' if not provided
```
siosonel, line 20: Since the 4th argument is not being used, can keep using the hardcoded `pp` value as $CONTAINER_ID. Also, `server/dockrun.sh` should be deleted, it was moved to `container/run.sh` and the comments in that file should reflect the file rename.

siosonel (commented):
Declining for now, unless changes are more useful. It's okay for the fixes remain simple, just trying to ensure that contributors are not spammers - please reply to this comment to verify non-spam status. Avoiding `@` mention for now for extra security.

PR #2285, opened 2024-10-18
create src/helper.ts with joinUrl()

### LINKED ISSUE #2279: use url.resolve()
do it next week

xzhou82:
## Description

closes #2279 
i tested every change in mds3.gdc.js and topVariablyExpressedGenes.ts
in new pr will finish up all other replacements

i created server/src/helpers.ts with the intent to migrate part or all server/src/utils.js there

## Checklist

### DIFF
--- a/server/src/helpers.ts
+++ b/server/src/helpers.ts
@@ -19,6 +19,7 @@ cannot do below:
 */
 export function joinUrl(p1: string, p2: string): string {
 	// p1 and p2 both shouldn't be blank string. if so, return null to alert
+	if (typeof p1 != 'string' || typeof p2 != 'string') throw `both arguments must be string type`
 	if (!p1 || !p2) throw 'blank string not allowed'
 	if (p1.indexOf('?') != -1) throw 'search string not allowed' // search string not allowed in p1. if usecase arises can support it
 	return (p1.endsWith('/') ? p1 : p1 + '/') + (p2.startsWith('/') ? p2.substring(1) : p2)
--- a/server/src/helpers.ts
+++ b/server/src/helpers.ts
@@ -1,0 +1,25 @@
+/*****************************************
+NOTE
+- here contains backend helpers that are not shared with client and not supposed to run on client
+- migrate utils.js stuff here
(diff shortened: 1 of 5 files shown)


server/src/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+when you have 3 pieces to join, do joinUrl( joinUrl(p1,p2), p3)
+
+cannot do below:
+- path.join() normalizes double-slash '://' to ':/' single-slash
+- url.resolve() replaces the path string after the domain, instead of appending to it
+*/
+export function joinUrl(p1: string, p2: string): string | null {
+	// p1 and p2 both shouldn't be blank string. if so, return null to alert
+	if (!p1 || !p2) return null
+	if (p1.indexOf('?') != -1) return null // search string not allowed in p1. if usecase arises can support it
```
siosonel, line 23: maybe `if(p1.indexOf('?')!=-1) return null // cannot handle search string` should be an error and throw?

or it can split using `?` and join `p2` to the first part and then re-join the second part with `?`

xzhou82: gdc host names won't have such. we can do it later

siosonel: Ok, but then it should just throw in line 22, since a fetch request that uses `joinUrl()` will not work anyway when the url is `null`.

xzhou82: changed to throw and covered in test

but the test cannot catch type err, e.g. when joinUrl() argument is number

siosonel: I added a test to detect non-string argument, will approve this and merge once all CI pass

xzhou82: thanks! i had thought that since the function arguments are already defined as string type, it will detect and reject non-string values at runtime?

siosonel: not at runtime, `tsc` will only run static type checks at compile time, so we have to either manually write tests to validate runtime arguments/request payloads or use `typia` which can auto-generate runtime type checker function based on typescript code - this is being fixed with augen refactor

siosonel (approved):

PR #2287, opened 2024-10-18
Fix for cohort radios not changing on click

creilly8:
## Description
In a future PR, should replace this logic with `make_radios` and only code the callback. 

Test
1. http://localhost:3000/?noheader=1&massnative=hg38-test,TermdbTest
2. http://localhost:3000/profile/index.html 

## Checklist

### DIFF
--- a/client/mass/about.ts
+++ b/client/mass/about.ts
@@ -137,12 +137,6 @@ class MassAbout {
 								subactions
 							})
 						} else await app.dispatch({ type: 'cohort_set', activeCohort: i })
-
-						nodes.forEach((n, j) => {
-							select(n)
-								.select('input')
-								.property('checked', j === state().activeCohort)
-						})
 					})
 
 				td0
@@ -242,7 +236,7 @@ class MassAbout {
 		}
 		const activeColumns = this.dom.cohortTable.selectAll(selector)
 		activeColumns.style('background-color', 'yellow')
-		this.dom.cohortInputs.property('checked', (d, i) => i === this.activeCohort)
+		//this.dom.cohortInputs.property('checked', (d, i) => i === this.activeCohort)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


xzhou82:
seems to be same. try switching cohort at http://localhost:3000/?noheader=1&massnative=hg38-test,TermdbTest

creilly8:
> seems to be same. try switching cohort at http://localhost:3000/?noheader=1&massnative=hg38-test,TermdbTest

@xzhou82, Please try it now. Note, I rebased to incorporate the changes from my other PR.

client/mass/about.ts
```diff
@@ -131,11 +132,17 @@ class MassAbout {
...
 								type: 'app_refresh',
 								subactions
 							})
-						} else app.dispatch({ type: 'cohort_set', activeCohort: i })
+						} else await app.dispatch({ type: 'cohort_set', activeCohort: i })
+
+						nodes.forEach((n, j) => {
+							select(n)
+								.select('input')
+								.property('checked', j === state().activeCohort)
```
airenzp, line 144: works now. state() looks weird and is defined up as s in line 112. I would rather use this.app.getState() than using this shortcut.

airenzp (approved):

client/mass/about.ts
airenzp: Actually, better not to await as these actions may take a while, and you know the cohort beforehand, it is i

airenzp: I thought a bit more about this logic, the node is being checked in this action, the problem is this line that runs on main, just removing the line fixes the issue
`		//this.dom.cohortInputs.property('checked', (d, i) => i === this.activeCohort)
`

xzhou82 (approved):

PR #2324, opened 2024-10-25
GDC API validation fix.

aacic:
## Description

Requires the `gdc-api-validation-handling` from the `ppgdc` repo.

To test:
Add GDC dataset to serverconfig.json
Execute: 
npm run dev

The server should start. 

Turn off WIFI connection and execute again:
npm run dev
The server should NOT start. 

Add to serverconfig:

```
 "features": {
    "runRemainingWithoutAwait": true
  }
```
Execute again:
npm run dev
The server should start. 

## Checklist


server/src/termdb.gdc.status.js
```diff
@@ -0,0 +1,45 @@
...
+ * @returns {Promise<boolean>}
+ */
+export async function testGDCApiStatus(ds) {
+	try {
+		const status = await getApiStatus(ds, 4)
+		if (status?.status !== 'OK') {
+			if (!serverconfig.features.runRemainingWithoutAwait) {
+				throw new Error('GDC API status is not OK')
+			} else {
+				return false
```
siosonel, line 18: I think this should automatically retry every 10 minutes or so, until the GDC API is back up, otherwise the GDC API will always be down in prp1 until the next restart/deployment. So, something like
```js
...
import { sleep } from '#src/utils'
...

const retries = {interval: 10 * 60 * 1000, max: 1000, total: 0}

export async function testGDCApiStatus(ds) {
	try {
		const status = await getApiStatus(ds, 4)
		if (status?.status !== 'OK') {
			if (!serverconfig.features.runRemainingWithoutAwait) {
				throw new Error('GDC API status is not OK')
			} else {
			        retries.total++
			         await sleep(retries.interval)
			         await testGDCApiStatus(ds)
			}
		}

```

siosonel (changes requested):
please see my comment about auto retry

server/src/termdb.gdc.status.js
```diff
@@ -0,0 +1,45 @@
...
+			throw new Error('failed to get GDC API status: ' + (error.message || error))
+		}
+	}
+	return true
+}
+
+export async function getApiStatus(ds, retry = 0, retryDelay = 1000) {
+	const { host, headers } = ds.getHostHeaders()
+	try {
+		return await ky(path.join(host.rest, 'status'), { headers }).json()
```
siosonel, line 34: May also need to allow for non-json-encoded response, like error 404 in html. Native `fetch` may be able to catch this error and retry.

server/src/termdb.gdc.status.js
siosonel: ah, just saw that the retry in in `getApiStatus()`. Why is this additional wrapper function needed?

server/src/termdb.gdc.status.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Tests the GDC API status and returns true if the status is OK.
+ * If serverconfig.features.runRemainingWithoutAwait is false, it will throw an error if the status is not OK.
+ * @param ds
+ * @returns {Promise<boolean>}
+ */
+export async function testGDCApiStatus(ds) {
+	try {
+		const status = await getApiStatus(ds, 4)
```
siosonel, line 13: The number of retries should be much more. Also, using the default 1 second retryDelay is too short, should be at least 10 minutes. The GDC API under maintenance status is very unlikely to change every second.

aacic: Should the server wait for GDC API to come back online before it starts, or should it restart and keep checking for GDC API to come back online?

siosonel: The latter, should restart and keep checking. The server should try to load the GDC status the first time, if it fails (not `ok`) and the `runRemainingAwait` is true, then keep retrying without blocking the loading of other datasets.

siosonel: For comparison, there are a few active loops to check GDC BAM cache and also separately, data release version to refresh the cache as needed. Both of these run without requiring the server to restart in case of needing a refresh, so I'm thinking along similar lines with regards to the status check, to have an active loop to try to automatically recover the GDC dataset loading without requiring a full server restart.

siosonel (approved):

server/src/termdb.gdc.js
```diff
@@ -1201,7 +1181,7 @@ async function caseCacheIsStale(ds) {
 
 	let value
 	try {
-		value = await getApiStatus(ds)
+		value = await ds.getStatus().data_release_version
```
xzhou82, line 1184: @siosonel @aacic coding err that should be ds.preInit.getStatus()
but how come you guys tested it to work??

PR #2325, opened 2024-10-28
Misc fixes

airenzp:
## Description

When adding a scale [here](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22MB_meta_analysis2%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22sampleScatter%22,%22name%22:%22Methylome%20TSNE%22,%22colorTW%22:{%22term%22:%20{%22gene%22:%20%22KBTBD4%22,%20%22type%22:%20%22geneVariant%22}},%22shapeTW%22:{%22id%22:%20%22Molecular%20Subgroup%20(MNPv12.8)%22}}]}) the scale legend was being clipped. Showed dslabel on about tab if no top provided. Fixed tabs space on SC. Restored facet as a default plot unless is not allowed. It was not showing before, now it shows except for the profile, that is the only one hiding all our plots.

## Checklist

### DIFF
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -293,7 +293,7 @@ function setRenderers(self) {
 			aboutTab.top = 'COHORT'
 			aboutTab.mid = ''
 			aboutTab.btm = ''
-		}
+		} else aboutTab.top = massNav?.tabs?.about?.top || appState.vocab.dslabel
 		const tabIdx = appState.termdbConfig?.selectCohort ? 0 : massNav?.tabs?.about?.order || 0
 		self.tabs.splice(tabIdx, 0, aboutTab)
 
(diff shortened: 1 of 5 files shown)


xzhou82 (approved):

xzhou82:
in sc color legend, should not use fixed column width but set column width with max item label width
also `Color by` tab name is uninformative. need a better name

xzhou82:
image
better to show violin option in the same row as gene search box

xzhou82:
in sc on overlaying gene exp, tooltip should place the color dot in gene exp row but not cluster
image

xzhou82:
restore label after clicking DE tab:
`View differentially expressed genes for cells of a cluster versus rest of the cells:`

xzhou82:
under `Plots` tab, the label `Plots:` shouldn't be bold. it's odd as other labels are not bold

Issue #2337, opened 2024-10-31
overlay barchart shows wrong data in log scale


xzhou82:
[overlay barchart](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22diaggrp%22},%22term2%22:{%22id%22:%22racegrp%22}}],%22nav%22:{%22activeTab%22:1}}), change scale to log. looks wrong.
<img width="596" alt="image" src="https://github.com/user-attachments/assets/05f522d4-d80a-44bd-97f0-33ee88cfc15c">

siosonel:
NOTE: It shouldn't be possible possible to have a scale for stacked bars (overlay term), the intermediate scales will be misleading. Technically each nested stacked bar segment scale should start from zero and have it's own scale (edited)

xzhou82:
we will disable log scale?

siosonel:
> we will disable log scale?

or do whichever is true below
- if the `log` radio option is clicked, do not show overlay option
- if there is an overlay, do not show `log` radio input

PR #2340, opened 2024-11-01
Comment out log radio button in the barchart

creilly8:
## Description
Address issue #2337. Also cleaned up import statements and wrote unit tests. 

For reference: The calculation for max domain value when the plot unit == 'log' was commented out before the repo became  public. From reviewing late in the afternoon yesterday, it looks like this issue has been unresolved for quite some time. 

## Checklist


client/plots/barchart.js
```diff
@@ -138,7 +135,12 @@ class Barchart {
 					settingsKey: 'unit',
 					options: [
 						{ label: 'Linear', value: 'abs' },
-						{ label: 'Log', value: 'log', getDisplayStyle: plot => (plot.term2 ? 'none' : 'inline-block') },
+						/** Option commented out because:
+						 * 1. The scale for the plot renders incorrectly
+						 * 2. The bars do not appear to be rendering correctly as a result
+						 * TODOs: Investigate the cause of the rendering issue and re-enable the option
+						 */
+						// { label: 'Log', value: 'log', getDisplayStyle: plot => (plot.term2 ? 'none' : 'inline-block') },
```
siosonel, line 143: I think we should keep the `log` option, but per my [comment in the associated issue](https://github.com/stjude/proteinpaint/issues/2337#issuecomment-2452061678), do whichever is currently true below
- if the log radio option is clicked, do not show overlay option
- if there is an overlay, do not show log radio input

siosonel: @creilly8 I see that the log is hidden when `plot.term2` is defined. But the overlay input has to also be hidden or disabled when the log scale is currently active.

creilly8: Addressed in https://github.com/stjude/proteinpaint/pull/2340/commits/45baf03ed659be6a654bd742b0a38417071c5de5

client/plots/barchart.js
```diff
@@ -106,7 +106,11 @@ export class Barchart {
 					vocabApi: this.app.vocabApi,
 					numericEditMenuVersion: this.opts.numericEditMenuVersion,
 					defaultQ4fillTW: term0_term2_defaultQ,
-					getDisplayStyle: () => (this.settings.unit == 'log' ? 'none' : '')
+					getDisplayStyle: () => (this.settings.unit == 'log' ? 'none' : ''),
+					processConfig: () => {
+						//Prevent log scale shown in overlay
+						if (this.settings.unit == 'log') this.settings.unit == 'abs'
+					}
```
siosonel, line 113: This is incorrect. The option should be provided with the input opts in the screenshot below. You can see that `opts.processConfig()` is used in `setTermsInput()` callback line 718 in `controls.config.js`. 

Screenshot 2024-11-05 at 10 01 35 AM

Screenshot 2024-11-05 at 10 03 19 AM

```js
{
	type: 'term',
	configKey: 'term2',
	chartType: 'barchart',
	usecase: { target: 'barchart', detail: 'term2' },
	title: 'Overlay data',
	label: 'Overlay',
	vocabApi: this.app.vocabApi,
	numericEditMenuVersion: this.opts.numericEditMenuVersion,
	defaultQ4fillTW: term0_term2_defaultQ
       processConfig(config) {
          config.settings.unit = 'abs'
      } 
},
```

siosonel: still not correct, should be `config.settings` not `this.settings` inside `processConfig()`

siosonel: I still see something wrong - does this work for you when you add an overlay term while unit is `log`? I'll be surprised if it worked that way with line 112, there's something wrong with how unit is set.

creilly8: No, this doesn't work. The only passed in config is `term2` or `term0`. That's why I used `this`. I figured you had a greater plan for the original request.

siosonel: A barchart `config.settings` can be part of the dispatch, see my `processConfig()` example above, it doesn't use `this.settings.unit == 'abs'`, it edist the config argument itself and also does not use double equal `==`.

```js
processConfig(config) {
          if (config.term2 && config.settings.unit == 'log') config.settings.unit = 'abs'
      } 
```

creilly8:
This is fixed. Also test [with this example](http://localhost:3000/?noheader=1&mass=%7B%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22id%22:%22diaggrp%22%7D,%22term2%22:%7B%22id%22:%22racegrp%22%7D%7D%5D,%22nav%22:%7B%22activeTab%22:1%7D%7D) selecting Proportion, then removing the overlay term. The scale and radio should default back to Linear.

siosonel:
@creilly8 I will fix the log scale ticks not lining up

siosonel (dismissed):

client/plots/test/barchart.render.unit.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**************
+ test sections
+***************/
+
+tape('\n', test => {
+	test.pass('-***- plots/barchart -***-')
+	test.end()
+})
+/** Difficult to test.
```
siosonel, line 48: @creilly8 I plan to delete this spec file. Unit testing of plot rendering is very difficult if it requires mocking of lots of data/components. I think integration tests are much easier to code and has very few downsides.

client/dom/radiobutton.ts
```diff
@@ -57,34 +59,37 @@ export function make_radios(opts: RadioButtonOpts): RadioApi {
...
-		.style('display', displayMode)
-		.style('padding', opts.styles && 'padding' in opts.styles ? opts.styles.padding : '3px')
-		.append('label')
+		//.style('display', opts.styles.display)
+		.style('padding', opts.styles.padding)
 
 	if (opts.styles) {
 		for (const k in opts.styles) {
-			labels.style(k, opts.styles[k])
+			// apply opts.styles to div, instead of labels
```
siosonel, line 87: @creilly8 I changed this line to apply the styles to `divs` instead of `labels`, it fixed the issue that @gavrielm noted of empty but visible divs with margin/padding. I tested in barchart and matrix, the radio input layout works as expected. Please test the radio layout/display toggling in other plots.

siosonel (approved):

PR #2352, opened 2024-11-06
fix: improve isNumeric() helper to not allow truncated alphanumeric p…

siosonel:
…arseFloat(), and create strictNumeric() helper. This also fixes intermittent session id errors dues to parseFloat() issue. It also adds a `strictNumeric()` helper for `termdb.bins.js` code that does not allow casting from string. 

To test, click `Session` button and `Share`, it should not fail intermittently. The issue was inside `isNumeric()` helper, `parseFloat(n)` was evaluating to `true` when a string starts with a number even if it's followed by non-numeric characters. This caused the subsequent `Number()` to be `NaN` and then converted as a `null` session id. 

Also tested with all unit and integration tests.

## Checklist

### DIFF
--- a/client/mass/test/summary.integration.spec.js
+++ b/client/mass/test/summary.integration.spec.js
@@ -31,6 +31,10 @@ const runpp = helpers.getRunPp('mass', {
 	debug: 1
 })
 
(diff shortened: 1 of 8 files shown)


creilly8 (dismissed):

airenzp (approved):
Thanks!

PR #2357, opened 2024-11-07
fix: barchart scale min should be zero even when unit='log'

siosonel:
since a linear bar scale is still used

## Description

In case the `log` scale option gets enabled in barchart, this fix would address the issue of bars not lining up to the matching log-scale tick value. In TermdbTest, avg dose to heart in log scale should show `1500 to <2000` line up to 100 and `3000 to <3500` bar length should be about 3/4 of `1500 to <2000` bar.

Screenshot 2024-11-07 at 4 12 09 PM

Tested with http://localhost:3000/testrun.html?dir=mass&name=barchart.integration and unit tests.

## Checklist


gavrielm (approved):
Looks good. Bars are not correct when overlay term is used, but it seems like @creilly8 will address that issue in her PR.

Issue #2229, opened 2024-10-10
`getSupportedChartTypes()` should assess ds.cohort{addCharts{},hideCharts{}} and replace allowedChartTypes[]


xzhou82:
getSupportedChartTypes() does two things:
1. compute chart types that can work for a ds, based on term type and data type availability. e.g. if survival term is present, allow survival chart. this computing only generate common chart types.
2. return allowed chart types per subcohort. mass client will show chart buttons solely based on this
3. optionally, apply ds overrides from a new structure below. addCharts will add things, hideCharts will hide things. this is a better design than allowedChartTypes[] that does two things and lead to excessive config on datasets

```
cohort {
  addCharts: {
     bySubcohort: {
       <cohort1>: [ 'chartype1', ...]
     }
     list: [ 'chartype1', ...]
   }
   hideCharts: {
      // same structure as addCharts
   }
}
```
1. `bySubcohort{}` is chart type list per subcohort, this will be used for profile which has a subcohort-specific chart type we have to support
2. `list[]` is a flat list
3. both bySubcohort and list can be used at same time

xzhou82:
may hold this off and coordinate with Airen's efforts

siosonel:
why is `hideCharts` required? Would not listing a chart in `addCharts` be sufficient?

xzhou82:
say if one ds wants to hide barchart, if we use just one structure, then
barchart has to be added to ALL other ds, this is too repetitive
compared to if using two structures for add/hide, only need a small change
to this ds

PR #2359, opened 2024-11-08
refactor getSupportedChartTypes() to clarify the rules by chartType

siosonel:
## Description
... instead of using procedural code. Closes https://github.com/stjude/proteinpaint/issues/2229
see sjpp `pp.issue.2229` for corresponding ds change

Tested with TermdbTest only, other datasets will need to be migrated in `sjpp` repo. The chart buttons in localhost should match the ones in prp1 (11 charts).

Screenshot 2024-11-08 at 4 46 24 PM

Tested with http://localhost:3000/testrun.html?name=*.unit and `cd client && npm run test:integration`.

## Checklist


server/src/termdb.server.init.js
```diff
@@ -473,93 +473,102 @@ export function server_init_db_queries(ds) {
...
-				}
-			}
-			if (ds.queries.geneExpression) {
-				for (const cohort in supportedChartTypes) {
-					supportedChartTypes[cohort].push('geneExpression')
-					supportedChartTypes[cohort].push('sampleScatter')
-				}
-				// TODO support clustering on dict terms
+		mayComputeTermtypeByCohort(ds) // ds.cohort.termdb.termtypeByCohort[] is set
+		const specialCharts = ds.cohort.specialCharts || []
```
xzhou82, line 551: please declare such in dataset.ts

server/src/termdb.server.init.js
```diff
@@ -473,93 +473,102 @@ export function server_init_db_queries(ds) {
...
-				// TODO: move draft charts out of flag once stable
-				supportedChartTypes[r.cohort].add(...serverconfig.features.draftChartTypes)
-			}
-			if (r.type == 'survival' && !supportedChartTypes[r.cohort].has('survival'))
-				supportedChartTypes[r.cohort].add('survival')
-			if (r.type == 'condition' && !supportedChartTypes[r.cohort].has('cuminc'))
-				supportedChartTypes[r.cohort].add('cuminc')
-			if (numericTypes.has(r.type)) numericTypeCount[r.cohort] += r.samplecount
-		}
+		const isSupportedChart = {
```
xzhou82, line 478: this is a good design. it can be improved by removing profile charts. this should only define general chart types that can be determined by term type availability. profile charts and any other special purpose charts are only defined in their own ds.cohort.specialCharts[]. and will be added to `supportedChartTypes` as-is.

siosonel: So that would mean that any chartType that's listed in `specialCharts`, is not meant to be be listed in `hiddenCharts`, right?

airenzp: Maybe we can have a method that does a switch with cases for the charts with more complex logic and a default behaviour, something like:
`
isSupported(chart)
{
switch(chart):
...

default:
return specialCharts.includes(chart) && !hiddenCharts.includes(chart)
}
`

shared/types/src/dataset.ts
```diff
@@ -1555,6 +1554,15 @@ export type Mds3 = BaseMds & {
...
 	queries?: Mds3Queries
 	cohort?: Cohort
+	/** supply "isSupported()" kind of callback per chart type,
+	that will overwrite default logic in getSupportedChartTypes()
+	- the callback can have arbitrary logic based on requirements from this ds
+	- can supply ()=>false to hide charts that will otherwise shown
+	- can define arbitrary chart type names for purpose-specific charts
+	*/
+	isSupportedChartOverride?: {
+		[chartType: string]: (f: any) => void
```
siosonel, line 1564: this has to return `boolean`, not void

xzhou82 (approved):
i suggest we merge it now and continue with new branch to fix below:

getSupporteChartType() will need to compute and return `dynamicScatter`, which is required by sampleScatter to decide if to show X/Y selector in chart btn menu. `dynamicScatter` can be viewed as childType or parameter of `sampleScatter`. this may become a general mechanism that getSupportedChartTypes() will return both main types and subtypes

PR #2365, opened 2024-11-11
Brain imaging add legends

congyu-lu:
## Description
will allows users to customize color in another branch

## Checklist

### DIFF
--- a/client/plots/brainImaging.js
+++ b/client/plots/brainImaging.js
@@ -495,7 +495,6 @@ async function getTableData(self, samples, state, refKey) {
 	return [rows, columns]
 }
 
-// TODO: when there is only one legend left, do not show the option to hide it
 function setInteractivity(self) {
 	self.legendLabelMouseup = event => {
 		const targetData = event.target.__data__
--- a/client/plots/brainImaging.js
+++ b/client/plots/brainImaging.js
@@ -2,10 +2,14 @@ import { getCompInit, copyMerge } from '#rx'
 import { controlsInit } from './controls'
 import { dofetch3 } from '#common/dofetch'
 import { renderTable } from '../dom/table.ts'
+import svgLegend from '#dom/svg.legend'
+import { scaleLinear } from 'd3-scale'
+import { Menu } from '#dom/menu'
 class BrainImaging {
 	constructor(opts) {
 		this.opts = opts
 		this.type = 'brainImaging'
+		setInteractivity(this)
 	}
 
 	async init(appState) {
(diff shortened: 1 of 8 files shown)


airenzp:
Thanks for working on this, age overlay looks like this:
Screenshot 2024-11-12 at 2 47 14 PM

It does not seem right

client/plots/brainImaging.js
```diff
@@ -186,9 +205,19 @@ class BrainImaging {
 			config.overlayTW?.term?.id !== previousConfig?.overlayTW?.term?.id ||
 			config.divideByTW?.term?.id !== previousConfig?.divideByTW?.term?.id
 
-		const updateL = config.settings.brainImaging.brainImageL != this.settings?.brainImageL || overlayOrDivideByChange
-		const updateF = config.settings.brainImaging.brainImageF != this.settings?.brainImageF || overlayOrDivideByChange
-		const updateT = config.settings.brainImaging.brainImageT != this.settings?.brainImageT || overlayOrDivideByChange
+		const legendFilterChange = JSON.stringify(previousConfig?.legendFilter) !== JSON.stringify(config.legendFilter)
```
airenzp, line 208: I wonder why this is needed, seems that with the other variables: updateL, updateF, updateT and overlayOrDivideByChange all the changes in the state can be tracked.

congyu-lu:
> Thanks for working on this, age overlay looks like this:
> 
> Screenshot 2024-11-12 at 2 47 14 PM
> It does not seem right

yes will need support numeric and gene term in another PR

client/plots/brainImaging.js
congyu-lu: legendFilterChange is for tracking the legendFilter change. e.g.Hide Male, then we need to re-render images and legends

airenzp (approved):
Looks good overall, thanks.

airenzp (approved):

siosonel:
should remove tracking of `update*` and jsut rely on dofetch/browser caching + rx state diff logic

airenzp:
If we do that it would rerender all the images on every request and it would compromise performance.

siosonel:
I tested it with Congyu, the network response is cached and the image is just the usual `` tag that doesn't have to rerender. It's not readily apparent to me which rendering code is expensive, I'll need more detail to understand the performance concern.

airenzp:
There are three images that would be rerendered, instead of one, or more if we consider divide by or overlay use cases. As of now, rendering has improved but is still slow. Yes, the cache helps if you request the same image, but it does not help otherwise. We can talk further about it. It may be easier to huddle.

congyu-lu:
tested with Edgar and when we don't check for updateL/F/T, changing slice index for sagital plane will only trigger one request for sagital (no requests for coronal or axial). And the response size is about 100kb so not too big to cache

airenzp:
I think I see what you mean. With the cache when you update one coordinate image even if all are requested only one request is done to the server.

PR #2380, opened 2024-11-15
Removed tracking of specific state changes in brain imaging as is not needed

airenzp:
## Description

Refactor. As discussed in previous PR with @siosonel and @congyu-lu removed tracking of specific state changes. Even when all the images are requested images that do not need to be rerendered will be cached already. Thanks for the heads up. Please check.

## Checklist

### DIFF
--- a/client/plots/brainImaging.js
+++ b/client/plots/brainImaging.js
@@ -45,9 +45,6 @@ class BrainImaging {
 			tdL,
 			tdF,
 			tdT,
-			imagesF: [],
-			imagesL: [],
-			imagesT: [],
 			legendHolder,
 			legendMenu
 		}
@@ -65,10 +62,11 @@ class BrainImaging {
 		}
 		this.components.controls.on('downloadClick.brainImaging', () => {
 			const urls = []
-			for (const category in this.dataUrlL) urls.push(this.dataUrlL[category].url)
-			for (const category in this.dataUrlF) urls.push(this.dataUrlF[category].url)
-			for (const category in this.dataUrlT) urls.push(this.dataUrlT[category].url)
-			console.log(urls)
+			for (const key in this.imagesData)
+				for (const category in this.imagesData[key].dataUrls) {
+					const dataUrl = this.imagesData[key].dataUrls[category].url
+					urls.push(dataUrl)
+				}
 			this.downloadImage(urls)
 		})
 		this.legendRenderer = svgLegend({ holder: this.dom.legendHolder })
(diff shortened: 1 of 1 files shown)


client/plots/brainImaging.js
```diff
@@ -247,153 +223,152 @@ class BrainImaging {
...
+			genome: this.state.genome,
+			dslabel: this.state.dslabel,
+			refKey: this.state.config.queryKey,
+			f: this.settings.brainImageF,
+			selectedSampleFileNames: this.state.config.selectedSampleFileNames,
+			divideByTW,
+			overlayTW,
+			legendFilter: this.state.config.legendFilter
+		}
+		data = await dofetch3('brainImaging', { body })
```
siosonel, line 276: reusing `body` and `data` variable names is a bit risky, it would be better to reuse a separate function for the different image requests, so that body and data could be constant inside the function. If done this way, could also parallelize the 3 requests like this with `Promise.all(requests)`.

client/plots/brainImaging.js
```diff
@@ -246,154 +220,98 @@ class BrainImaging {
...
-						.html('&nbsp;')
-						.style('font-weight', '600')
-						.style('font-size', '24px')
-						.style('margin-bottom', '5px')
-						.style('margin-top', '5px')
-						.style('display', 'block')
-				const img = this.dom.tdT.append('div').append('img').attr('src', result.url)
-				this.dom.imagesT.push(img)
-			}
+			promises.push(await dofetch3('brainImaging', { body }))
```
siosonel, line 279: Do not await here when pushing to `promises` array, because that will block the loop on each iteration. Only await on the `Promise.all()`.

client/plots/brainImaging.js
```diff
@@ -246,154 +220,98 @@ class BrainImaging {
...
+
+	async requestImages(keys) {
+		const promises = []
+		for (const key of keys) {
 			const body = {
 				genome: this.state.genome,
 				dslabel: this.state.dslabel,
 				refKey: this.state.config.queryKey,
-				t: this.settings.brainImageT,
+				l: this.settings[key],
```
siosonel, line 273: I think `l` should be `f` or `t`, depending on the key? Instead of just passing array of string `keys` as argument, can pass an array of object that includes this single letter key as a property, 
```js
...
this.imagesData = {
	brainImageL: { dataUrls: {}, td: this.dom.tdL, key: 'l', value: this.settings.brainImageL },
	brainImageF: { dataUrls: {}, td: this.dom.tdF, key: 'f', value: this.settings.brainImageF },
	brainImageT: { dataUrls: {}, td: this.dom.tdT, key: 't', value: this.settings.brainImageT }
}
...
this.datas = await this.requestImages(Object.values(this.imagesData))
...
async requestImages(images) {
    const promises = []
		for (const img of images) {
			const body = {
				genome: this.state.genome,
				dslabel: this.state.dslabel,
				refKey: this.state.config.queryKey,
				[img.key]: [img.value],
                         }  
}
```

airenzp:
Thanks for the careful revision @siosonel. I was in a hurry to finish this on friday evening. Everything looks good to me now.

siosonel (approved):

Issue #2403, opened 2024-11-19
boxplot not displaying 0 values in legend entries


xzhou82:
[test link](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22ALL-pharmacotyping%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22summary%22,%22childType%22:%22boxplot%22,%22term%22:{%22id%22:%22Asparaginase_normalizedLC50%22,%22q%22:{%22mode%22:%22continuous%22}}}]}) at legend below, Min and 1st quartile value is 0 but not displayed

Total: 809
Minimum
1st quartile
Median: 0.57
Mean: 0.5

PR #2404, opened 2024-11-19
Fix for zeros not appearing in descriptive stats

creilly8:
## Description
Closes issue #2403

Please test with [this example](http://localhost:3000/?noheader=1&mass=%7B%22genome%22:%22hg38%22,%22dslabel%22:%22ALL-pharmacotyping%22,%22nav%22:%7B%22activeTab%22:1%7D,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22childType%22:%22boxplot%22,%22term%22:%7B%22id%22:%22Asparaginase_normalizedLC50%22,%22q%22:%7B%22mode%22:%22continuous%22%7D%7D%7D%5D%7D). Also fixed issue with values appearing blank in the tooltip when plot value count is less than five. Please test with [this example](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg19%22,%22dslabel%22:%22PNET%22,%22nav%22:{%22header_mode%22:%22hidden%22},%22plots%22:[{%22chartType%22:%22summary%22,%22childType%22:%22boxplot%22,%22term%22:{%22id%22:%22Age%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22Mutational%20Burden%22}}]}), the 11-20 mutations plot.

## Checklist


server/routes/termdb.boxplot.ts
```diff
@@ -163,17 +163,25 @@ function setDescrStats(boxplot: BoxPlotData, sortedValues: number[]) {
 	const sd = Math.sqrt(s / (sortedValues.length - 1))
 	const squareDiffs = sortedValues.map(x => (x - mean) ** 2).reduce((a, b) => a + b, 0)
 	const variance = squareDiffs / (sortedValues.length - 1)
+
+	/** boxplot_getvalue won't return these values for small # of samples
+	 * Calculate value subset for tooltips. */
```
siosonel, line 168: Instead of recalculating below by copying code from `boxplot_getvalue()`, it's less redundant and less error-prone to support an optional second argument in `boxplot_getvalue()`, that could override the minimum 5 value requirement. However, I'm not sure if it's still valid to render a boxplot in our case  with less than 5 values, maybe @xzhou82 could verify. Also, calculating quartiles with less than 5 values seems to not be related to [the linked issue](https://github.com/stjude/proteinpaint/issues/2403), where there are 800+ values, so this extra code seems optional to me since the fix is really within the LegendRender.ts updates?
```js
export function boxplot_getvalue(lst, minNumValues=5) {
	/* ascending order
    each element: {value}
    */
	const l = lst.length
	if (l < minNumValues) {
		// less than minNumValues/items, won't make boxplot
		return { out: lst }
	}
```

client/plots/boxplot/viewModel/LegendDataMapper.ts
```diff
@@ -23,21 +27,22 @@ export class LegendDataMapper {
 		if (config.term.q?.descrStats) {
 			this.legendData.push({
 				label: `Descriptive Statistics${isTerm2 ? `: ${config.term.term.name}` : ''}`,
-				items: config.term.q.descrStats
+				items: this.setDescrStatArr(config.term.q.descrStats)
 			})
 		}
 		if (isTerm2 && isTerm2.q?.descrStats) {
```
siosonel, line 33: why is `isTerm2` variable required, and not simply use optional chaining `if (config?.term2?.q?.descrStats)`?

server/routes/termdb.boxplot.ts
xzhou82: i agree here it shouldn't re-calculate when there're less than 5 data points --  just do not show these stats on the client is okay

shared/utils/src/test/roundvalue.unit.spec.js
```diff
@@ -23,7 +23,7 @@ tape('roundValue tests', function (test) {
 	test.equal(roundValueAuto(value), rounded, `should return ${rounded}`)
 
 	value = 10549.23556789
-	rounded = 10549.24
+	rounded = '1.1e+4'
```
xzhou82, line 26: lacks test for big negative numbers, e.g. -9876 and -19876

server/routes/termdb.boxplot.ts
creilly8: This is fixed.

shared/utils/src/test/roundvalue.unit.spec.js
creilly8: Another test has been added.

siosonel:
Trying with filter with less than 5 samples, I get errors for boxplot. The Network error for `termdb/boxplot` request is 
```json
{
    "error": "absMin or absMax is null [termdb.boxplot init()]"
}
```

Screenshot 2024-11-20 at 1 52 27 PM

creilly8:
> a0c7690

This is fixed in [a0c7690](https://github.com/stjude/proteinpaint/pull/2404/commits/a0c7690b1ec5ee0d007aca9f7ec0c80d229a9927)

siosonel:
@creilly8 Minor re-wording request, instead of `No visible box plot data to render`, since the filter tab shows 2 samples, so there is data but it's insufficient in our way of rendering boxplot. Maybe `Insufficient or no visible plot data to render`?

creilly8:
> @creilly8 Minor re-wording request, instead of `No visible box plot data to render`, since the filter tab shows 2 samples, so there is data but it's insufficient in our way of rendering boxplot. Maybe `Insufficient or no visible plot data to render`?

@siosonel I used the same wording as the barchart. I'm fine with changing it to `Insufficient or no visible plot data to render` and updating the barchart and violin error messages. Otherwise all three messages are worded differently. @xzhou82, does that work for you?

siosonel:
@creilly I see, I think in the particular test link and filter that I have in my screenshot, both the barchart and violin also do not have data for `asparaginase`. But in a filter where there is at least 1 annotated sample for barchart, then the `No visible data` is an accurate message for barchart, but may not be accurate for boxplot. In other words, there may be times when barchart can render while the boxplot does not, I'm trying to look for an example.

PR #2411, opened 2024-11-21
Profile templates2

airenzp:
## Description

Showed filters, improved legend and misc fixes. Handled session recovery by passing the state to getPlotConfig. This allows to know the selected cohort when recovering a session. Please check.

## Checklist


client/plots/profilePlot.js
```diff
@@ -598,19 +605,15 @@ export function makeChartBtnMenu(holder, chartsInstance, chartType) {
...
 
-export async function loadFilterTerms(config, app, opts) {
-	const state = app.getState()
-	const activeCohort = state ? state.activeCohort : opts.activeCohort //when recovering a session the state is not available, use the opts provided
-	const cohortPreffix = activeCohort == FULL_COHORT ? 'F' : 'A'
+export async function loadFilterTerms(config, state, app) {
+	const cohortPreffix = state.activeCohort == FULL_COHORT ? 'F' : 'A'
 	const twlst = []
 	config.countryTW = { id: cohortPreffix + 'country' }
 	config.regionTW = { id: cohortPreffix + 'WHO_region' }
```
siosonel, line 619: why is `cohortPrefix` needed for `tw.id`? and this is for the tw.id, not for term.id, right?

client/mass/store.js
```diff
@@ -97,7 +97,7 @@ class TdbStore {
 				let plot
 				try {
 					const _ = await import(`../plots/${savedPlot.chartType}.js`)
-					plot = await _.getPlotConfig(savedPlot, this.app)
+					plot = await _.getPlotConfig(savedPlot, this.app, this.state)
```
siosonel, line 100: this is not a good idea, `this.state` is not fully computed until after state rehydration. So getPlotConfig() when called as part of state rehydration should not expect a full state.

airenzp: Ok, how can I get the active cohort in this case?

siosonel (approved):

client/mass/about.ts
```diff
@@ -135,9 +135,17 @@ export class MassAbout {
 					.style('margin-left', '0px')
 					.on('click', async () => {
 						const state = app.getState()
-						const clearOnChange = state.clearOnChange
+						const clearOnChange = state.termdbConfig.selectCohort.clearOnChange
```
xzhou82, line 138: purpose and structure of `clearOnChange` is not clear in dataset.ts

client/plots/profileBarchart.js
```diff
@@ -298,9 +298,10 @@ class profileBarchart extends profilePlot {
 	}
 }
 
-export async function getPlotConfig(opts, app) {
+export async function getPlotConfig(opts, app, _activeCohort) {
 	try {
-		const defaults = getProfilePlotConfig(app, opts)
+		const activeCohort = _activeCohort === undefined ? app.getState().activeCohort : _activeCohort
```
xzhou82, line 303: please document why the 3rd param `_activeCohort` is needed and it can't be derived from app.getState().activeCohort?

client/mass/about.ts
airenzp: I will add comment in a next PR. This allows to specify if the dataset should delete the plots or the filter when changing cohorts. So far it is only done for the profile, Edgar suggested to add this configuration to the dataset so any dataset with this config would behave the same.

client/plots/profileBarchart.js
airenzp: When recovering a session getState() returns null and that is why we need to pass the activeCohort.

PR #2432, opened 2024-11-27
attempt to recache if a caching error is detected besides a stale ver…

siosonel:
…sion; catch all awaited functions that are not in a try-catch block

Requires related branches from sjpp and gdc: `git unite git unite gdc-await-catch`

## Description

Since GDC relies on network requests as part of dataset initialization, it is more prone to startup errors compared to local file-based datasets. This PR branch tries to recover from all initial GDC caching errors, by setting a `ds.__gdc.cacheError` string that the stale cache detector can use in addition to the GDC API version query. 

TODO later: use a different test server or container process for GDC and any network-based datasets in the future?

Test steps:
1. set `serverconfig.gdcCacheCheckWait` to 5-10 seconds
2. turn off dev machine wifi
3. restart `npm run start`
4. when the server log starts to listen to port 3000, and there are GDC-related caching messages, turn the wifi back on, the gdc caching should try recover
5. before the caching completes (can be fast if `extApiCache` is used), turn the wifi off again. This simulates an error that occurs after the GDC API returns a healthy status and before the caching is complete. The caching should hang again.
6. Turn your wifi back on again. The caching should restart with a message like `detected caching error: cacheMappingOnNewRelease() call to fetchIdsFromGdcApi(), getCasesWithGeneExpression(), or getAnalysisTsv2loom4scrna()`

`cached gdc data version is up-to-date, but there was a caching error, will restart cache`

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+Features:
+- support MassNav.tabs.about.activeItems
+
 
 ## 2.88.2
 
--- a/release.txt
+++ b/release.txt
@@ -1,2 +1,1 @@
-Features:
-- support MassNav.tabs.about.activeItems
+


server/src/initGdc.cache.js
```diff
@@ -96,7 +96,11 @@ async function getOpenProjects(ds, ref) {
 	}
 
 	const { host, headers } = ds.getHostHeaders()
-	const { body: re } = await cachedFetch(joinUrl(host.rest, 'files'), { method: 'POST', headers, body: data })
+	const { body: re } = await cachedFetch(joinUrl(host.rest, 'files'), { method: 'POST', headers, body: data }).catch(
```
xzhou82, line 99: could you comment on why needs `.catch()`?

server/src/mds3.init.js
```diff
@@ -117,7 +117,9 @@ export async function init(ds, genome, _servconfig) {
 	}
 
 	// must validate termdb first
-	await validate_termdb(ds)
+	await validate_termdb(ds).catch(e => {
```
xzhou82, line 120: line 44 `import { preInit } from '#src/mds3.preInit.ts'` why is `#src/` needed?

PR #2476, opened 2024-12-06
Gdc await catch

siosonel:
## Description

This update attempts to expand auto-recovery for recoverable errors besides the initial healthy GDC API detection.  Previously, retries only applied to detecting a healthy `/status` response, and the server crashed if any subsequent caching requests failed due to connectivity issues or API 5xx status (server maintenance). 

To test:
- set `serverconfig.features.gdcCacheCheckWait=9000` 
- use `npx tsx server.ts` from sjpp instead of `npm start` to more clearly see server crash
- and uncomment one of the test `.then()` lines in `initGdc.cache.js`
  - line ~136: HTTP 4xx should cause an immediate crash on initial cache attempt, client request must be fixed
  - line ~450: HTTP 5xx should not cause a crash, API may be under maintenance and will be healthy again
  - line ~552: error code='ENOTFOUND' is assumed to be caused by a temporary network issue that will be resolved
  - line ~82: one of the preceding `.then()` recoverable errors has to be active, which will be converted to a non-recoverable error on retry: should cancel all subsequent retries

## Checklist

### DIFF
--- a/server/src/mds3.preInit.ts
+++ b/server/src/mds3.preInit.ts
@@ -2,15 +2,17 @@ import serverconfig from '#src/serverconfig.js'
 import type { Mds3 } from '#types'
 
 /**
- * Tests the GDC API status and returns true if the status is OK.
- * If serverconfig.features.runRemainingWithoutAwait is false, it will throw an error if the status is not OK.
- * If serverconfig.features.runRemainingWithoutAwait is true, it will retry the request up to retryMax times.
+ * Tests the dataset readiness for initialization and data queries,
+ * for example checking if the GDC API is healthy before querying data for caching.
+ * If a raw dataset entry (in serverconfig.genomes[].datasets[]) has
+ * awaitMds3Init === true, it will throw an error if the status is not OK
+ * otherwise, it will retry the request up to retryMax times.
  * @param ds
  * @param retryDelay
  * @param retryMax
  */
 
-export async function preInit(ds: Mds3): Promise<any> {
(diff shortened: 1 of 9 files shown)


server/src/mds3.variant2samples.js
```diff
@@ -27,8 +27,10 @@ export async function validate_variant2samples(ds) {
 	const vs = ds.variant2samples
 	if (!vs) return
 
-	if (ds.preInit?.isReady) {
-		const response = await ds.preInit.isReady()
+	if (ds.preInit?.getStatus) {
```
xzhou82, line 30: may explain

server/src/mds3.preInit.ts
```diff
@@ -20,17 +22,22 @@ export async function preInit(ds: Mds3): Promise<any> {
...
+		if (response.status != 'recoverableError') {
+			// should not retry on fatal error
+			console.log(`fatal error: ${ds.label} ds.preInit.getStatus()`, response)
+			throw new Error('status is not OK: ' + response.message)
+		}
+
 		if (retryMax < 1) {
-			throw new Error('GDC API status error: ' + (error.message || error))
+			console.log('35 mayRetryDsPreInit()', response)
+			throw new Error('GDC API status error: ' + (response.message || response))
```
xzhou82, line 37: if ds.preInit() is intended to be generic but not gdc-specific, may delete mentioning of GDC

xzhou82 (approved):

PR #2479, opened 2024-12-08
fix: scrna gene exp violin disables brushing and label menu

### LINKED ISSUE #2438: violin plot using scrna gene exp data, `List sample` over a range can list all cells rather by range
right now on scrna gene exp violin, it has the original capability to select a range to `List samples`. if this function is to be kept, it should be called `List cells`. however, this function will result in all cells being loaded without observing exp value range

xzhou82:
## Description

closes #2438 
violin CI passes, scrna views (native and gdc) works

## Checklist

### DIFF
--- a/client/plots/violin.interactivity.js
+++ b/client/plots/violin.interactivity.js
@@ -1,9 +1,9 @@
 import { filterJoin, getFilterItemByTag } from '#filter'
-import { renderTable } from '../dom/table'
+import { renderTable, niceNumLabels } from '#dom'
 import { to_svg } from '#src/client'
 import { roundValueAuto } from '#shared/roundValue.js'
 import { rgb } from 'd3'
-import { niceNumLabels } from '../dom/niceNumLabels'
+import { TermTypes } from '#shared/terms.js'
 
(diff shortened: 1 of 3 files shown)


karishma-gangwani:
thanks, can you add link or steps on how to test it

client/plots/violin.interactivity.js
```diff
@@ -25,9 +25,8 @@ export function setInteractivity(self) {
 	}
 
 	self.displayLabelClickMenu = function (t1, t2, plot, event) {
-		if (!t2 || self.data.plots.length === 1) {
-			return
-		}
+		if (!t2 || self.data.plots.length === 1) return // when no term 2 and just one violin, do not show options on the sole violin label
```
airenzp, line 28: you may want to see the t1 values, @karishma-gangwani requested this feature.

airenzp (approved):
I am approving, thanks for the fix. I would suggest to allow showing the sample values even when there is no overlay.

airenzp:
@karishma-gangwani you can test this on any SC plot, select the gene expression tab, search a gene, then open the violin associated.

Issue #2474, opened 2024-12-06
dom/table.ts allows to sort rows by a column


xzhou82:
should be an optional feature. may need this for next gdc release

PR #2507, opened 2024-12-12
Prototype table sort option

creilly8:
## Description
Closes issue #2474 

Changes: 
1. Reformatted the comments in `dom/table` to work with typedoc
2. Prototype sorting function for table

Test: 
In `client/dom/test/table.unit.spec.js`, comment out line 303. Open http://localhost:3000/testrun.html?dir=dom&name=table.unit. Click on the up/down icons to sort the data. 

## Checklist

### DIFF
--- a/client/gdc/maf.js
+++ b/client/gdc/maf.js
@@ -304,7 +304,7 @@ async function getFilesAndShowTable(obj) {
 	for (const f of result.files) {
 		const row = [
 			{
-				html: `<a href=https://portal.gdc.cancer.gov/cases/${f.case_uuid} target=_blank>${f.case_submitter_id}</a>`,
+				html: `<a href=https://portal.gdc.cancer.gov/cases/${f.case_uuid} target=_blank onclick="event.stopPropagation()">${f.case_submitter_id}</a>`,
 				value: f.case_submitter_id
 			},
 			{ value: f.project_id },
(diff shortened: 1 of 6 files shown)


xzhou82:
i pushed a change. could you test at http://localhost:3000/?gdcmaf=1 to see if table can be sorted?

creilly8:
> i pushed a change. could you test at http://localhost:3000/?gdcmaf=1 to see if table can be sorted?

I added another attribute, `sortable`, to the columns json so not every column is sorted. For example, sorting columns with only .html:"string" would be useless for the user. I see that none of the columns include this attribute. That's why this isn't working here.

If I change line 18 in `client/gdc/maf.js` to `const tableColumns = [{ label: 'Case', sortable: true }, { label: 'Project', sortable: true }, { label: 'Samples', sortable: true }, { label: 'File Size', sortable: true }]
`, it works.

client/gdc/maf.js
```diff
@@ -304,7 +304,7 @@ async function getFilesAndShowTable(obj) {
 	for (const f of result.files) {
 		const row = [
 			{
-				html: `<a href=https://portal.gdc.cancer.gov/cases/${f.case_uuid} target=_blank>${f.case_submitter_id}</a>`,
+				html: `<a href=https://portal.gdc.cancer.gov/cases/${f.case_uuid} target=_blank onclick="event.stopPropagation()">${f.case_submitter_id}</a>`,
```
xzhou82, line 307: if this cannot be implemented in table.ts, should make a note there that clickable link in html must have ...

xzhou82 (approved):

PR #2522, opened 2024-12-16
Barchart optional mult testing correction

congyu-lu:
## Description
Adding a checkbox for performing multiple testing correction as described in https://docs.google.com/drawings/d/1OuQCiS-i4VkeAg4_7dz7BemYeBCEfoZag67bf6kSWos/edit 

Could test with analysis paper Supplementary Figure S1-C: 
[barchart_mtc.txt](https://github.com/user-attachments/files/18156274/barchart_mtc.txt)
Without multiple testing correction, G3 shows asterisks 
with multiple testing correction, G3 shows no asterisks

## Checklist

### DIFF
--- a/client/plots/barchart.js
+++ b/client/plots/barchart.js
@@ -194,7 +194,8 @@ export class Barchart {
 				}
 			]
 
-			if (this.app.getState().termdbConfig.multipleTestingCorrection) {
+			const multipleTestingCorrection = this.app.getState().termdbConfig.multipleTestingCorrection
+			if (multipleTestingCorrection) {
 				// a checkbox to allow users to show or hide asterisks on bars
 				inputs.push({
 					label: 'Asterisks',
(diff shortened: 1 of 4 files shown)


client/plots/barchart.js
```diff
@@ -204,6 +204,14 @@ export class Barchart {
 					settingsKey: 'asterisksVisible',
 					title: 'Display the asterisks'
 				})
+				inputs.push({
+					label: 'Multiple testing correction',
```
karishma-gangwani, line 208: what type of multiple testing correct is applied? Can we specify that on the portal somewhere?

congyu-lu: thanks added it to label

karishma-gangwani (approved):
This looks good. Thanks @congyu-lu Do you think somewhere we could show the adjusted alpha value? Perhaps instead of the hardcoded 'p-value' because if something is no longer significant a user might want to see the actual adjusted value. Hope that makes sense.

congyu-lu:
> This looks good. Thanks @congyu-lu Do you think somewhere we could show the adjusted alpha value? Perhaps instead of the hardcoded 'p-value' because if something is no longer significant a user might want to see the actual adjusted value. Hope that makes sense.

Yes makes sense, could discuss this in the meeting.

xzhou82 (approved):

PR #2527, opened 2024-12-17
Added profile fixes discussed in the last meeting

airenzp:
## Description

Solves issues:
Name the top label HOME
Use darker text background and blue background for the tabs
If there is only one choice don't show a menu when selecting radars
Show tooltip on radar table column _difference_ instead of table footnote
Highlight cohorts selection in the home page.

See corresponding PR in sjpp

Home page looks like this now:

Screenshot 2024-12-17 at 12 38 11 PM

## Checklist

### DIFF
--- a/client/dom/table.ts
+++ b/client/dom/table.ts
@@ -33,7 +33,7 @@ export type Column = {
 	/** left, center, right. If missing it is aligned to the left by default */
 	align?: string
 	/** tooltip describing column content */
-	title?: string
+	tooltip?: string
 	/** Used for sorting function
 	 * Do not use this field for html columns */
 	sortable?: boolean
@@ -236,7 +236,7 @@ export function renderTable({
 		for (const [i, c] of columns.entries()) {
 			const th = theadRow.append('th').text(c.label).attr('class', 'sjpp_table_item sjpp_table_header')
 			if (c.width) th.style('width', c.width)
-			if (c.title) th.attr('title', c.title)
+			if (c.tooltip) th.attr('title', c.tooltip)
 			if (header?.allowSort) {
 				//Only create sort button for columns with data
 				//(i.e. not html columns)
--- a/client/mass/about.ts
+++ b/client/mass/about.ts
@@ -147,6 +147,7 @@ export class MassAbout {
 				const radioId = radioName + '-' + i
 				td0
 					.append('input')
+					.style('scale', '1.2')
 					.attr('type', 'radio')
 					.attr('name', radioName)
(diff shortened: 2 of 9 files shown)


client/dom/table.ts
```diff
@@ -33,7 +33,7 @@ export type Column = {
 	/** left, center, right. If missing it is aligned to the left by default */
 	align?: string
 	/** tooltip describing column content */
-	title?: string
```
xzhou82, line 36: with this change, can u try to verify if plot code calling renderTable() is not still supplying `.title`?

airenzp: This property was added by me for the profile, I dont think anyone else is using it. Is more clear to name it tooltip.

shared/types/src/dataset.ts
```diff
@@ -785,7 +785,6 @@ export type SelectCohortEntry = {
 	//Therefore when switching cohorts, it is necessary to delete the plots opened and the global filter
 	clearOnChange?: { [index: string]: boolean }
 	//Allows to customize the color of the selected column on the cohort features table. Used by the PrOFILE
```
xzhou82, line 787: should move comment along with property

client/mass/about.ts
```diff
@@ -91,14 +91,14 @@ export class MassAbout {
...
 		}
-
 		if (opts.selectCohort?.prompt) {
 			this.dom.cohortPrompt = this.subheader
 				.append('div')
 				.style('margin-left', '10px')
 				.style('padding-top', '30px')
 				.style('padding-bottom', '10px')
 				.style('font-weight', 'bold')
+				.style('font-size', '1.2em')
```
xzhou82, line 101: this causes radio buttons look noticeably small compared to text
try .style.zoom="120%" on the div element wrapping these

airenzp: When I scale, the way to zoom is to use the scale property, it looked like this, I wanted to keep the radio buttons aligned with the text. I will see how.
Screenshot 2024-12-17 at 1 03 54 PM

client/mass/about.ts
```diff
@@ -257,7 +258,7 @@ export class MassAbout {
 			}
 		}
 		const activeColumns = this.dom.cohortTable.selectAll(selector)
-		const color = state.termdbConfig.activeColor || 'yellow'
+		const color = state.termdbConfig.massNav.activeColor || 'yellow'
```
xzhou82, line 261: massNav is optional

xzhou82 (approved):

PR #2572, opened 2025-01-07
Box plot UI improvements and violin fix

creilly8:
## Description

Changes: 
1. Fixed broken `List samples` label menu options for numeric terms in both the violin and box plot. This also fixed the broken filter menu option in the violin plot. 
2. Added color picker and filter menu options to the box plot label menu. 
3. Fixed the download when dark mode is applied. 

Test: 
1. [Numeric example with mbmeta](http://localhost:3000/?noheader=1&mass={%22genome%22:%22hg38%22,%22dslabel%22:%22MB_meta_analysis2%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22summary%22,%22childType%22:%22boxplot%22,%22term%22:{%22id%22:%22Age%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22Confidence%20Score%22}}]}) -> test the 'List samples' and 'Add filter' options for both the violin and box plot. 
2. [Categorical example with sjlife](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22summary%22,%22childType%22:%22boxplot%22,%22term%22:{%22id%22:%22agedx%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22diaggrp%22}}]}) -> should work exactly the same as master. 

## Checklist

### DIFF
--- a/client/plots/boxplot/interactions/ListSamples.ts
+++ b/client/plots/boxplot/interactions/ListSamples.ts
@@ -3,7 +3,6 @@ import type { RenderedPlot } from '../view/RenderedPlot'
 import type { TermWrapper } from '#types'
 import type { AnnotatedSampleData } from '../../../types/termdb'
 import { roundValueAuto } from '#shared/roundValue.js'
-import { isNumericTerm } from '#shared/terms.js'
 
 export class ListSamples {
 	app: MassAppApi
@@ -136,9 +135,6 @@ export class ListSamples {
 				this.createTvsTerm(tw2, tvslst)
 				tvslst.lst[0].tvs.ranges = this.assignPlotRangeRanges()
 				if (getRange) this.createTvsLstRanges(tw1, tvslst, rangeStart, rangeStop, 1)
-			} else if (isNumericTerm(tw1.term) && isNumericTerm(tw2.term)) {
-				this.createTvsLstRanges(tw2, tvslst, rangeStart, rangeStop, 0)
-				if (getRange) this.createTvsLstRanges(tw1, tvslst, rangeStart, rangeStop, 1)
 			} else {
 				this.createTvsLstValues(tw2, tvslst, 0)
 				if (getRange) this.createTvsLstRanges(tw1, tvslst, rangeStart, rangeStop, 1)
@@ -146,6 +142,7 @@ export class ListSamples {
 		} else {
 			if (getRange) this.createTvsLstRanges(tw1, tvslst, rangeStart, rangeStop, 0)
 		}
+
 		return tvslst
 	}
 }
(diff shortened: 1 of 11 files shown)


client/plots/boxplot/interactions/ListSamples.ts
```diff
@@ -12,17 +12,25 @@ export class ListSamples {
...
 		this.app = app
 		this.plot = plot
+
 		const plotConfig = state.plots.find((p: { id: string }) => p.id === id)
 		if (!plotConfig) throw 'Box plot not found [ListSamples]'
-		const tvslst = this.getTvsLst(min, max, plotConfig.term, plotConfig.term2)
+
+		const min = plot.descrStats.find(d => d.id === 'min')!.value
+		const max = plot.descrStats.find(d => d.id === 'max')!.value
+		if (Number.isNaN(min) || Number.isNaN(max)) throw 'Min or max not found [ListSamples]'
```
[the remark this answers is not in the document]
creilly8: That wouldn't work in this case. A value, number or null, is always returned for min and max from the server. This is check is purely to avoid a type error. Also using optional chaining doesn't avoid assigning the constant as undefined or null, causing an error somewhere else.

creilly8:
> line 43, async getData function can simply use 'return'. can remove the await. it still works

That line calls an async vocab function for a server request. It's best to give it time to complete.

client/plots/boxplot/viewModel/ViewModel.ts
karishma-gangwani, line 1: both Disco and boxplot contain similarly named files. Could you change to keep these separate?

creilly8: I don't think that's necessary.

server/routes/termdb.boxplot.ts
karishma-gangwani, line 1: around line 51 and line 61 in termdb.boxplot.ts, the logic for checking uncomputable values is repetitive. Is it possible to clean that up? even in termdb.violin it is repeated to check for term1 and term2 uncomputable values. Do you think we could create a handler for this since this check is done in many places in our codebase?

client/plots/boxplot/interactions/ListSamples.ts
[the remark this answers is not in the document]
xzhou82: line 25 throw will crash client. if this can actually happen should enclose it in try-catch and print msg somewhere to avoid crashing

server/routes/termdb.boxplot.ts
xzhou82: no need to change this. a better refactor will be to combine violin computing logic into `termdb.boxplot.ts` and delete `termdb.violin.js` to remove lots of redundancy. route can be named as /termdb/boxplotViolin to indicate it computes either box/violin depending on param

xzhou82 (approved):

PR #2575, opened 2025-01-08
check if cache map is empty before accessing so to return a better er…

xzhou82:
…r msg

## Description

can test by adding `stopGdcCacheAliquot: true` to serverconfig features{} to prevent caching from happening, and test at gdc scrna gene query. previous message says cannot map id which is uninformative

does not impact normal gdc operation

## Checklist

### DIFF
--- a/server/routes/termdb.singlecellSamples.ts
+++ b/server/routes/termdb.singlecellSamples.ts
@@ -314,6 +314,10 @@ function gdc_validateGeneExpression(G, ds, genome) {
 		*/
 		try {
 			const fileid = q.sample.eID
+			if (ds.__gdc.scrnaAnalysis2hdf5.size == 0) {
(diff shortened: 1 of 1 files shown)


compbiolover:
When I add `stopGdcCacheAliquot: true` I do get the expected behavior with the more elegant error message. Unfortunately, I also get the error message when I delete `stopGdcCacheAliquot: true` from my serverconfig features{}. I strongly think this issue relates only to my system as I can see when I start my server that `GDC: sample IDs are not cached! No periodic check will take place!`. I am good to approve this pull request.

compbiolover (approved):

congyu-lu (approved):
saw the error message when adding "stopGdcCacheAliquot": true
did not see the error without "stopGdcCacheAliquot": true

PR #2580, opened 2025-01-09
Allow adding pre built numberic dicterm cluster

congyu-lu:
## Description

Please test with sjpp branch: ALL_prebuilt_matrix_numericDictTermCluster

## Checklist

### DIFF
--- a/server/src/termdb.matrix.js
+++ b/server/src/termdb.matrix.js
@@ -649,11 +649,11 @@ works with "canned" NumericDictionaryTerm plot in a dataset, e.g. data from a te
 called in mds3.init
 */
 export async function mayInitiateNumericDictionaryTermplots(ds) {
-	if (!ds.cohort.numericDictTermClusterPlots) return
-	if (!Array.isArray(ds.cohort.numericDictTermClusterPlots.plots))
-		throw 'cohort.numericDictTermClusterPlots.plots is not array'
-	for (const p of ds.cohort.numericDictTermClusterPlots.plots) {
-		if (!p.name) throw '.name missing from one of numericDictTermClusterPlots.plots[]'
+	if (!ds.cohort.termdb?.numericDictTermCluster?.plots) return
+	if (!Array.isArray(ds.cohort.termdb.numericDictTermCluster.plots))
+		throw 'cohort.termdb.numericDictTermCluster.plots is not array'
+	for (const p of ds.cohort.termdb.numericDictTermCluster.plots) {
+		if (!p.name) throw '.name missing from one of numericDictTermCluster.plots[]'
 		if (p.file) {
 			const numericDictTermClusterConfig = await read_file(path.join(serverconfig.tpmasterdir, p.file))
 			p.numericDictTermClusterConfig = JSON.parse(numericDictTermClusterConfig)
--- a/server/src/termdb.matrix.js
+++ b/server/src/termdb.matrix.js
@@ -644,6 +644,26 @@ export async function mayInitiateMatrixplots(ds) {
 	}
 }
 
+/*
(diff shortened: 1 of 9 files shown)


server/src/termdb.matrix.js
```diff
@@ -644,6 +644,26 @@ export async function mayInitiateMatrixplots(ds) {
 	}
 }
 
+/*
+works with "canned" NumericDictionaryTerm plot in a dataset, e.g. data from a text file
+called in mds3.init
+*/
+export async function mayInitiateNumericDictionaryTermplots(ds) {
+	if (!ds.cohort.numericDictTermClusterPlots) return
```
xzhou82, line 652: please define type def for `ds.cohort.numericDictTermClusterPlots`

congyu-lu: thanks just added

xzhou82:
what do u think about combining the existing ds.cohort.termdb.numericDictTermCluster{} into the new property, thus gather related things in one place
```
ds.cohort. numericDictTermCluster{
  appName
  settings{}
  exclude
  plots[]
}
```

congyu-lu:
> what do u think about combining the existing ds.cohort.termdb.numericDictTermCluster{} into the new property, thus gather related things in one place
> 
> ```
> ds.cohort.numericDictTermCluster{
>   appName
>   settings{}
>   exclude
>   plots[]
> }
> ```

thanks that's a good idea, just combined

karishma-gangwani (dismissed):
works for me. so list samples on spanning on matrix doesn't show values in second column. is that intentional or a bug?

server/src/termdb.matrix.js
xzhou82: did you commit ur changes? i don't see it combined

xzhou82:
did u commit? i don't see the two are combined

congyu-lu:
> did u commit? i don't see the two are combined

sorry my git push failed, just pushed again, please check

shared/types/src/dataset.ts
```diff
@@ -1028,6 +1030,15 @@ type MatrixPlotsEntry = {
 	getConfig?: (f: any) => void
 }
 
+type NumericDictTermClusterPlotsEntry = {
+	name: string
+	file: string
+	settings?: {
+		[key: string]: any
+	}
+	getConfig?: (f: any) => void
```
xzhou82, line 1039: what's purpose of getConfig()?

xzhou82:
thanks a lot this looks better. please update sjpp branch in order to test

congyu-lu:
> works for me. so list samples on spanning on matrix doesn't show values in second column. is that intentional or a bug?

Its intentional to only list the samples, cannot show values for the samples in another column, because in a big matrix, there will be many values (terms) for a specific sample, will be a very big table. To find all the values for a specific sample, users are expected to use sampleView

shared/types/src/dataset.ts
congyu-lu: just added comment

congyu-lu:
> thanks a lot this looks better. please update sjpp branch in order to test

thanks sjpp already updated

karishma-gangwani:
> > works for me. so list samples on spanning on matrix doesn't show values in second column. is that intentional or a bug?
> 
> Its intentional to only list the samples, cannot show values for the samples in another column, because in a big matrix, there will be many values (terms) for a specific sample, will be a very big table. To find all the values for a specific sample, users are expected to use sampleView

okay yes. maybe later we can think of a way to select samples and 'download data' and add additional option for 'add to a group and filter'.

xzhou82 (approved):

PR #2592, opened 2025-01-10
Hot app refresh

siosonel:
## Description

Implement hot module replacement for a pp app, without triggering a page reload. This helps developer workflow, especially when integrating with external portal code like GFF.

~~NOTE: When swtiching back to master branch, delete all files from `proteinpaint/.sse/messages/` dir to prevent endless flashing browser refresh.~~

Must stop and restart `npm run dev` after pulling this branch. Tested with any plot, but "real" partial refresh only works with mass, termdb, lollipop, and gdc tool wrappers (in `client/gdc`).  Other apps will always clear `arg.holder` and call `runproteinpaint()` again. Saving client or server code changes should notify the user of rebundling status.

Also, `bindProteinPaint()` in `client/src/app.js` will be merged with `runproteinpaint()` in a separate PR, to limit the scope of these changes, which includes fixes to SSE issues (that increased the scope of these changes). I will add release notes in the next PR, before releasing to GDC.

The SSE refresh option can be selected in the dropdown in the top right corner, where the notification message is shown.
```js
        partial: 'call runproteinpaint() on the same DOM element, with the original argument + any HMR state',
	partial_except_on_unhide: 'same as reload, except when a hidden browser tab is made visible',
	full:
		'call window.reload() to refresh from URL params and/or runpp() arguments, ' +
		'this may have less memory leaks than the default approach',
	full_except_on_unhide: 'same as reload, except when a hidden browser tab is made visible',
	none: 'do not refresh the app view'
```

## Checklist


congyu-lu (dismissed):
works in local and GFF

karishma-gangwani (approved):
everything seems to work. thanks @siosonel

siosonel:
@congyu-lu , @creilly8 , and @karishma-gangwani : I updated the code to not persist message files, so there is no need to delete `proteinpaint/.sse/messages/` files even when switching to `master` branch. The server code in `server/src/test/routes/sse.js` is also easier to read now.

congyu-lu (approved):

Issue #2583, opened 2025-01-09
list samples should correspond to samples hidden in legend category


karishma-gangwani:
After hiding categories on barchart and listing samples still shows the hidden category. 

![image](https://github.com/user-attachments/assets/f83e4b75-fe63-4382-8d88-15d18d922916)

creilly8:
@karishma-gangwani, I'm going to change the label from `good first issue` to `bug`. It would be difficult for someone new to ProteinPaint to figure out how our server calls work and then how our rendering code processes the response.

karishma-gangwani:
> @karishma-gangwani What is the link/dataset you used to create this example?

Oh I used a custom term for it. Let me share a session with you for this offline.

PR #2593, opened 2025-01-10
Do not show hidden values in the barchart `List sample` label menu option results

creilly8:
## Description
Closes issue #2583.

Catches custom hidden variables as categories before displaying them on the client. Will send data to use for testing in All-pharmacotyping example: http://localhost:3000/?noheader=1&massnative=hg38,ALL-pharmacotyping. 

## Checklist

### DIFF
--- a/client/plots/barchart.events.js
+++ b/client/plots/barchart.events.js
@@ -540,11 +540,22 @@ async function listSamples(event, self, seriesId, dataId, chartId) {
 	for (const sample of data.lst) {
 		const sampleName = data.refs.bySampleId[Number(sample.sample)].label
 		const row = [{ value: sampleName }]
+		/** Don't show hidden values in the results
+		 * May not be caught in server request for custom variables
+		 * with user supplied keys */
+		if (self.config.term.q?.hiddenValues && `${sample[self.config.term.$id].key}` in self.config.term.q.hiddenValues)
+			continue
 		if (termIsNumeric) {
 			const value = sample[self.config.term.$id]?.value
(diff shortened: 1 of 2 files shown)


karishma-gangwani:
seeing this error when 'RNA-seq' from assay availability is overlaid with 'sex' in all-pharma and categories are clicked for hiding. 

image

client/plots/barchart.events.js
```diff
@@ -540,11 +540,22 @@ async function listSamples(event, self, seriesId, dataId, chartId) {
 	for (const sample of data.lst) {
 		const sampleName = data.refs.bySampleId[Number(sample.sample)].label
 		const row = [{ value: sampleName }]
+		/** Don't show hidden values in the results
+		 * May not be caught in server request for custom variables
+		 * with user supplied keys */
+		if (self.config.term.q?.hiddenValues && `${sample[self.config.term.$id].key}` in self.config.term.q.hiddenValues)
```
karishma-gangwani, line 546: do you want to add additional check with optional chaining to check against the .key?

karishma-gangwani (approved):

client/plots/barchart.events.js
creilly8: Not really. The in operator make this safer since the string value itself could be anything: str | undefined | null

xzhou82: i suggest to look into a backend fix instead

xzhou82:
not such a priority but if the problem really comes from backend, should look at a proper backend fix

PR #2598, opened 2025-01-14
Scatter improvements

airenzp:
## Description

Supported adding Z coordinate in continuous mode for premade plots (3D plot). If no color specified, colored particles building color scale from z coords.
If color in continous mode enabled for 2D scatter plots used it when calculating contour maps  to weight coordinates.

## Checklist

### DIFF
--- a/client/plots/sampleScatter.js
+++ b/client/plots/sampleScatter.js
@@ -332,7 +332,7 @@ class Scatter {
 				chartType: 'sampleScatter',
 				settingsKey: 'showContour',
 				title:
-					"Shows the density of point clouds. If 'Z/Divide by' is added in continous mode, it uses this to weight the points when calculating the density contours"
+					"Shows the density of point clouds. If 'Color' is used in continous mode, it uses it to weight the points when calculating the density contours. If 'Z/Divide by' is added in continous mode, it used it instead."
 			}
 		]
 		if (this.settings.showContour)
--- a/client/plots/sampleScatter.js
+++ b/client/plots/sampleScatter.js
@@ -138,7 +138,7 @@ class Scatter {
 			else this.createChart(key, data, 0)
 		}
 		this.initRanges()
-		this.is3D = this.config.term && this.config.term0?.q.mode == 'continuous'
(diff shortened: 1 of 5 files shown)


compbiolover:
A couple of things. When first launching the test the genes are in lower case. When searching in the dictionary they are shown and displayed as uppercase. 
gene_title_difference

In addition, I noticed that when I have turned on contour map in the 2D view and I select a variable from divide by (in my example age at index) it keeps the 2D view until I change the mode of the variable to continuous. Since I am new to the team I'm not sure if the plot is supposed to remain in 2D when using bin sizes. 
remains_in_2d_view
 
When I do change it to continuous it appears to remove the axes even though I have not unchecked the show axes box.
axes_removed
 
Finally, when I uncheck show contour map it behaves as expected with the display of the 3D plot colored by each axis. [
proper_3d_behavior_age_at_index
](url)

This holds true for other variables where I simply don't show the contour map
proper_3d_behavior

xzhou82 (approved):
recommend to name branch `issue.2594`
can fix unrelated issues in new branches

airenzp:
Thanks Andrew for the detailed revision with screenshots. Looks nice and easy to reproduce. The genes could be uppercase I suppose but that is something to be addressed later. There is a logic to use continuous mode if another mode is not set, but in this case the term comes with discrete mode by default. Maybe we could change to continuous mode in this case if X and Y are already continuous like in this case.

airenzp:
About showing the axes for the contour map in the 3D plot, I see it may be confusing. I may look into it later. It has some technical difficulties as I need to integrate it to the contour image generated for the 3D plots and also I am not sure it would be meaningful to interpret the contours.

compbiolover:
Sounds good on the gene casing. I like your suggestion of defaulting to continuous mode if X and Y are already in continuous mode. With regard to the axes on the 3D contour I follow what you mean. Even if it isn't meaningful to have the axes for the contour I think it would be helpful to have a little more explanation to the user as I found the sudden removal of the axes confusing. Thanks!

PR #2610, opened 2025-01-15
Login in the profile

airenzp:
## Description

Passed getDatasetAccessToken to runproteinpaint to login. Updated login logic related.

## Checklist

### DIFF
--- a/client/mass/about.ts
+++ b/client/mass/about.ts
@@ -64,6 +64,7 @@ export class MassAbout {
 	subheader: Elem // where all contents are rendered
 	type: string
 	opts: MassAboutOpts
+	state: any
 
 	constructor(opts: MassAboutOpts) {
 		this.opts = opts
--- a/client/mass/about.ts
+++ b/client/mass/about.ts
@@ -63,8 +63,10 @@ export class MassAbout {
 	selectCohort: SelectCohortEntry | null
 	subheader: Elem // where all contents are rendered
 	type: string
+	opts: MassAboutOpts
 
 	constructor(opts: MassAboutOpts) {
+		this.opts = opts
 		this.type = 'about'
 		this.app = opts.app
 		this.subheader = opts.subheader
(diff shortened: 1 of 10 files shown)


server/routes/termdb.config.ts
```diff
@@ -68,17 +70,14 @@ returns:
...
 	// add attributes to this object to reveal to client
 
-	// TODO: may fill this in later with other request-specific credentials
-	const auth = { embedder: q.embedder }
-
 	// add required attributes
 	const c: any = {
 		selectCohort: tdb.selectCohort, // optional
-		supportedChartTypes: tdb.q?.getSupportedChartTypes(auth),
+		supportedChartTypes: tdb.q?.getSupportedChartTypes(req),
```
airenzp, line 80: Here we need to pass the auth to getSupportedChartTypes. Also, eventually, the cohort selected should be passed

airenzp: it seems that getTermdbConfig would need to be called when the selected cohort changes in order to update the charts that can be seen

client/mass/about.ts
```diff
@@ -114,6 +105,19 @@ export class MassAbout {
 	since there's no reason to update about tab contents while user is interacting with mass ui
 	*/
 	init(appState) {
+		console.log(appState)
+		if (this.opts.selectCohort?.description || this.opts.selectCohort?.descriptionByUser) {
+			//temporary logic to get the profile description until the login is implemented
+			const [logged, site, user] = getProfileLogin(this.app, appState.activeCohort)
```
xzhou82: getProfileLogin() is outdated

please replace it with the new vocabApi method to directly return clientAuthResult

client/mass/about.ts
```diff
@@ -124,6 +115,16 @@ export class MassAbout {
 
 	async main() {
 		await this.renderCohortsTable()
+		if (this.opts.selectCohort?.description || this.opts.selectCohort?.descriptionByUser) {
+			//temporary logic to get the profile description until the login is implemented
+			const [logged, site, user] = getProfileLogin(this.app, this.state.activeCohort)
```
xzhou82, line 120: getProfileLogin() is outdated

please replace it with the new vocabApi method to directly return clientAuthResult

client/plots/profilePlot.js
```diff
@@ -27,7 +27,7 @@ export class profilePlot {
 	getState(appState) {
 		const config = appState.plots.find(p => p.id === this.id)
 		if (!config) throw `No plot with id='${this.id}' found`
-		const [logged, site, user] = getProfileLogin() //later on replace with real login info
+		const [logged, site, user] = getProfileLogin(this.app, appState.activeCohort) //later on replace with real login info
```
xzhou82, line 30: can clientAuthResult be used here instead?

client/mass/about.ts
airenzp: I still need to post process the auth result, I was centralizing the post processing in getProfileLogin. 
See, this still needs to be done somehow:
`const auth = app.vocabApi.getClientAuthResult()
	if (!auth) return [false, null, 'public']
	const logged = true
	const auth_info = cohort == FULL_COHORT ? auth.full : auth.abbrev
	if (!auth_info) return [false, null, 'public'] //no login info for the cohort, treat it as not logged in, public view for that cohort
	const user = auth_info.role
	const site = user == 'admin' ? null : auth_info.site //site only matters if you are not admin
	return [logged, site, user]`

airenzp:
This code breaks other datasets with this error

`TypeError: Cannot read properties of undefined (reading 'label')
    at Object.authApi.getNonsensitiveInfo (/Users/azaldiva/dev/sjpp/proteinpaint/server/src/auth.js:585:51)
    at Object.q.getSupportedChartTypes (/Users/azaldiva/dev/sjpp/proteinpaint/server/src/termdb.server.init.ts:491:28)
    at make`

client/mass/about.ts
xzhou82: good solution is to compute it on server using ds callback based on auth. will implement it in next branch

xzhou82 (approved):

PR #2612, opened 2025-01-15
add new auth api to pass clientAuthResult to chart callback override

xzhou82:
## Description

please advise and test the implementation

## Checklist

### DIFF
--- a/server/routes/termdb.config.ts
+++ b/server/routes/termdb.config.ts
@@ -77,7 +77,7 @@ function make(q, req, res, ds: Mds3WithCohort, genome) {
 	// add required attributes
 	const c: any = {
 		selectCohort: tdb.selectCohort, // optional
-		supportedChartTypes: tdb.q?.getSupportedChartTypes(getClientAuth(q, req)),
+		supportedChartTypes: tdb.q?.getSupportedChartTypes(req),
 		renamedChartTypes: ds.cohort.renamedChartTypes,
 		allowedTermTypes: getAllowedTermTypes(ds),
 		termMatch2geneSet: tdb.termMatch2geneSet,
@@ -304,12 +304,4 @@ function getAllowedTermTypes(ds) {
 	if (ds?.queries?.metaboliteIntensity) typeSet.add(TermTypes.METABOLITE_INTENSITY)
 
 	return [...typeSet]
-}
-
-function getClientAuth(q, req) {
-	// optional auth results based on the request (if this ds requires it). pass to getSupportedChartTypes to customize chart list based on user role
-	return {
-		embedder: q.embedder,
-		clientAuthResult: authApi.getClientAuthResult(req)
-	}
 }
--- a/server/routes/termdb.config.ts
+++ b/server/routes/termdb.config.ts
@@ -30,7 +30,7 @@ function init({ genomes }) {
 			if (!genome) throw 'invalid genome'
 
 			const [ds] = get_ds_tdb(genome, q)
-			return make(q, res, ds, genome)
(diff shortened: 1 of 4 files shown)


server/src/auth.js
```diff
@@ -485,6 +485,16 @@ async function maySetAuthRoutes(app, basepath = '', _serverconfig = null) {
 		}
 	})
 
+	// helper with some code duplication just to get clientAuthResult from jwt
+	authApi.getClientAuthResult = function (req) {
```
siosonel, line 489: I think this should be generalized and also return `forbiddenRoutes`, so that `getForbiddenRoutesForDsEmbedder()` does not have to be called separately inside `q.getSupportedChartTypes()` in `termdb.server.init.js`. 

So, maybe call this function `authApi.getNonsensitiveInfo()`, and it would return `{forbiddenRoutes, clientAuthResults}`, and can be extended easily later.

server/routes/termdb.config.ts
```diff
@@ -306,3 +305,11 @@ function getAllowedTermTypes(ds) {
 
 	return [...typeSet]
 }
+
```
siosonel, line 308: If `forbiddenRoutes` is returned together with `clientAuthResults` by a new authApi method, then no need for this extra function, since `embedder` is used only to determined forbidden routes within `getSupportedChartTypes()`.

server/routes/termdb.config.ts
```diff
@@ -68,17 +70,14 @@ returns:
...
 	// add attributes to this object to reveal to client
 
-	// TODO: may fill this in later with other request-specific credentials
-	const auth = { embedder: q.embedder }
-
 	// add required attributes
 	const c: any = {
 		selectCohort: tdb.selectCohort, // optional
-		supportedChartTypes: tdb.q?.getSupportedChartTypes(auth),
+		supportedChartTypes: tdb.q?.getSupportedChartTypes(getClientAuth(q, req)),
```
siosonel, line 80: If `forbiddenRoutes` is returned together with `clientAuthResults` by a new authApi method, then should remain as one argument.

siosonel:
@xzhou82 I think it would be better to not edit `termdb.config.js`. Instead, the new authApi method should be called within `termdb.server.init.js`, and should still return `forbiddenRoutes` together with `clientAuthResults` as I previously suggested. So, `getSupportedChartTypes()` could remain as is, it could grab request info from the request params like before.

server/src/termdb.server.init.ts
```diff
@@ -490,15 +490,15 @@ export function server_init_db_queries(ds) {
...
+		const authInfo = authApi.getNonsensitiveInfo(req)
 		const supportedChartTypes = {} // key: subcohort string, value: list of chart types allowed for this cohort
 
 		for (const [cohort, cohortTermTypes] of Object.entries(ds.cohort.termdb.termtypeByCohort.nested)) {
 			supportedChartTypes[cohort] = []
 
 			for (const [chartType, isSupported] of Object.entries(commonCharts)) {
 				// auth second argument may be ignored by isSupported() callback
-				if (isSupported({ ds, forbiddenRoutes, cohortTermTypes, clientAuthResults })) {
+				if (isSupported(Object.assign({ ds, cohortTermTypes }, authInfo))) {
```
siosonel, line 501: can use object srpead syntax, `{  ds, cohortTermTypes , ...authInfo }`, instead of Object.assign()

airenzp (approved):

airenzp:
I merged these changes into my branch so we can continue to work from there

Issue #2599, opened 2025-01-14
when all mass nav tabs are closed, it is not saved in session


xzhou82:
at http://localhost:3000/?noheader=1&massnative=hg38-test,TermdbTest, click About tab to hide. "share" a session and click link to restore, about tab is still open

PR #2613, opened 2025-01-15
Save subheader display option in the state

creilly8:
## Description
Closes issue #2599. Whether or not the subheader should be shown is saved in the state. Test with http://localhost:3000/?noheader=1&massnative=hg38-test,TermdbTest -> click About tab -> Share session -> shouldn't see the subheader. 

## Checklist

### DIFF
--- a/client/mass/nav.js
+++ b/client/mass/nav.js
@@ -180,6 +180,7 @@ class TdbNav {
 					this.samplecounts[this.filterJSON] = n
 				}
 			}
+			this.displaySubheader = this.state.nav.displaySubheader
 		}
 		this.updateUI()
 	}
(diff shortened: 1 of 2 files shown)


xzhou82:
thanks. what do you think about setting activeTab=-1 instead of introducing and managing a new flag?

creilly8:
> thanks. what do you think about setting activeTab=-1 for 'no tabs open' instead of introducing and managing a new flag?

Since activeTab is used as an index lookup, setting it equal to undefined will likely create logic problems later. I get where you're going with this but will that allow the next person to be successful? Maybe not.

xzhou82:
still a bug: [at link](http://localhost:3000/?noheader=1&massnative=hg38-test,TermdbTest), click about tab to hide subheader. then, click on charts tab. charts tab doesn't turn on and subheader not shown

xzhou82 (approved):

PR #2620, opened 2025-01-17
correlation volcano

xzhou82:
## Checklist


client/mass/types/mass.ts
```diff
@@ -63,9 +63,10 @@ export type MassState = {
 	activeCohort: number
 	customTerms?: any[]
 	groups: any[]
-	plots: PlotConfig[]
+	plots: BasePlotConfig[]
 	termfilter: any
 	termdbConfig: {
+		correlationVolcano: any
```
xzhou82, line 69: this should be optional

server/utils/corr.R
```diff
@@ -0,0 +1,36 @@
...
+
+# Adjusting for multiple testing correction
+adjust_p_values <- p.adjust(pvalues, method = "fdr")
+output <- data.frame(ids, coeffs, -log10(pvalues), -log10(adjust_p_values), sample_sizes)
+names(output)[1] <- "id"
+names(output)[2] <- "correlation"
+names(output)[3] <- "original_p_value"
+names(output)[4] <- "adjusted_p_value"
+names(output)[5] <- "sample_size"
+cat(paste0("adjusted_p_values:", toJSON(output)))
```
xzhou82, line 36: change it to just return stringnified json, rather that nodejs has to cut a line and check what it is

client/mass/types/mass.ts
creilly8: Thanks. This is fixed.

server/utils/corr.R
robinpaul85: This makes it easier to print variables inside the R script for debugging. If this is not added, nodejs considers the entire output (including other print statements from R) as JSON which breaks PP during debugging. Most of my other R and rust scripts are designed this way so its easier to debug R and rust codes.

server/utils/corr.R
```diff
@@ -0,0 +1,36 @@
...
+for (i in 1:length(v1)) {
+    cor <- cor.test(as.numeric(unlist(v1[i])), as.numeric(unlist(v2[i])), method = input$method)
+    coeffs <- c(coeffs, cor$estimate)
+    pvalues <- c(pvalues, cor$p.value)
+    sample_sizes <- c(sample_sizes, length(as.numeric(unlist(v1[i]))))
+}
+
+# Adjusting for multiple testing correction
+adjust_p_values <- p.adjust(pvalues, method = "fdr")
+output <- data.frame(ids, coeffs, -log10(pvalues), -log10(adjust_p_values), sample_sizes)
```
xzhou82, line 30: just return pvalue but not -log10

server/utils/corr.R
xzhou82: either use `message(msg)` to print to stderr (don't do if this throws in node)

or better and i highly prefer this, you should have a test script to run the r script with test data on commandline, that will be more efficient than having to refresh the web browser each time, and be able for script to cleanly return json

Issue #2648, opened 2025-01-23
reusable dot size legend component with control menu for corrVol, as implemented in samplescatter


xzhou82:
see [example](http://localhost:3000/?noheader=1&mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22sampleScatter%22,%22name%22:%22TermdbTest%20TSNE%22,%22colorTW%22:{%22id%22:%22diaggrp%22},%22scaleDotTW%22:{%22id%22:%22agedx%22}}]})
so that we maintain a uniform legend appearance in two apps and allow for customization which could be important

unlike samplescatter which allows reversal of dot size scaling, corrvol won't need to reverse it

PR #2683, opened 2025-01-28
Implement legend component into correlation volcano plot

creilly8:
## Description
Closes issue #2648 . 

Changes: 
1. Refined the legend circle reference component. The component uses simpler arguments. 
2. Implemented the component into the correlation volcano plot. The radius can be updated from both the legend by clicking on the `Sample size` element to the right of the plot as well as the control panel. 
3. Updated the unit tests. 

Test with the [all pharma example](http://localhost:3000/?mass={%22dslabel%22:%22ALL-pharmacotyping%22,%22nav%22:{%22activeTab%22:1},%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22correlationVolcano%22,%22featureTw%22:{%22term%22:{%22type%22:%22geneExpression%22,%22gene%22:%22KRAS%22}}}]}). 

## Checklist

### DIFF
--- a/client/dom/LegendCircleReference.ts
+++ b/client/dom/LegendCircleReference.ts
@@ -16,7 +16,9 @@ export class LegendCircleReference {
 	inputMax: number
 	inputMin: number
 	isAscending: boolean
+	maxLabel: number
 	maxRadius: number
+	minLabel: number
 	minRadius: number
 	menu?: {
 		minMaxLabel?: string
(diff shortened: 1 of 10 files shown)


client/dom/LegendCircleReference.ts
```diff
@@ -16,12 +16,16 @@ export class LegendCircleReference {
 	inputMax: number
```
xzhou82, line 16: may rename the script as `DotSizeScale`, similar to `ColorScale`

client/plots/corrVolcano/CorrelationVolcano.ts
```diff
@@ -110,6 +114,26 @@ class CorrelationVolcano extends RxComponentInner {
 				settingsKey: 'threshold',
 				debounceInterval: 0.05
 			},
+			{
```
xzhou82, line 117: since it's already controllable via legend which is highly visible, may not need to add these 2 opts

xzhou82 (approved):
thanks
the dot size scaling component should not only allow to set pixel size, but also the scale via auto/fixed method same as ColorScale

creilly8:
> thanks the dot size scaling component should not only allow to set pixel size, but also the scale via auto/fixed method same as ColorScale

@xzhou82, It can be whatever value the caller wants. The corr volcano uses "pixel" but the sample scatter could set the label to "scale" as an example.

PR #2753, opened 2025-02-06
blur an auto-focused tab, so that there is not distracting border hig…

siosonel:
…hlight around a tab

## Description

This should stop the automatic border highlight for the `Samples` (first tab) in the singlecell plot. When navigating by keyboar (`tab` or `shift+tab`), the browser's autohighlight will still show up.

## Checklist


siosonel:
> Thanks the samples tab highlight on load is gone. When using tab I see the blue border as expected, just I do not see that the focus fades with blur, just wanted to let you know.

The `blur()` only happens when the tabbed content is first rendered, after line 170 forced `focus()`, and which happens without a user having to navigate by keyboard. If a user does navigate by keyboard (`tab` or `shift+tab`), then the border highlight should persist per usual browser section 508 behavior, it shouldn't be blurred/lose focus.

PR #2763, opened 2025-02-07
SC fixes

airenzp:
## Description

Do not preselect  a sample but show only Samples tabs until a sample is selected. Improved sample heading. Aligned legend with plot for GDC. Added  subtype in BALL [sjpp](https://github.com/stjude/sjpp/pull/656). I used the idea about hiding other tabs that we discussed @siosonel to do as GDC asked.

Screenshot 2025-02-07 at 12 18 17 PM

Two plots in a row with the legend besides the plot:

Screenshot 2025-02-07 at 1 36 25 PM

## Checklist

### DIFF
--- a/client/dom/table.ts
+++ b/client/dom/table.ts
@@ -325,6 +325,12 @@ export function renderTable({
 						}
 					})
 
+				if (i === selectedRows[0]) {
+					setTimeout(() => {
+						td.node().parentNode.scrollIntoView({ behavior: 'smooth', block: 'center' })
+					}, 500)
+				}
+
 				const checked = checkbox.property('checked')
 				for (const key in selectedRowStyle) {
 					tr.style(key, checked ? selectedRowStyle[key] : '')
--- a/client/dom/table.ts
+++ b/client/dom/table.ts
@@ -326,6 +326,8 @@ export function renderTable({
 					})
 
 				if (i === selectedRows[0]) {
+					// if there is at least one selected row, scroll to that table row,
+					// so that it's visible and obvious to the user which rows are pre-selected
 					setTimeout(() => {
 						td.node().parentNode.scrollIntoView({ behavior: 'smooth', block: 'center' })
 					}, 500)
--- a/client/plots/singleCellPlot.js
+++ b/client/plots/singleCellPlot.js
@@ -84,6 +84,9 @@ class singleCellPlot {
 		})
 		this.tabs = []
 		const activeTab = state.config.activeTab
+		// shared isVisible function for tabs that require config.sample
+		const isVisible = () => this.state?.config.sample
+
 		this.tabs.push({
 			label: 'Samples',
 			id: SAMPLES_TAB,
@@ -94,7 +97,7 @@ class singleCellPlot {
 			label: 'Plots',
 			id: PLOTS_TAB,
 			active: activeTab == PLOTS_TAB,
-			isVisible: () => state.config.sample != undefined || this.state?.config.sample,
+			isVisible,
 			callback: () => this.setActiveTab(PLOTS_TAB)
 		})
 		if (state.termdbConfig.queries?.singleCell?.DEgenes) {
@@ -102,7 +105,7 @@ class singleCellPlot {
 				label: 'Differential Expression',
 				id: DIFFERENTIAL_EXPRESSION_TAB,
 				active: activeTab == DIFFERENTIAL_EXPRESSION_TAB,
-				isVisible: () => state.config.sample != undefined || this.state?.config.sample,
+				isVisible,
 				callback: () => this.setActiveTab(DIFFERENTIAL_EXPRESSION_TAB)
 			})
 			if (this.app.opts.genome.termdbs)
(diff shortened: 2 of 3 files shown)


client/plots/singleCellPlot.js
```diff
@@ -95,13 +95,15 @@ class singleCellPlot {
...
 			active: activeTab == PLOTS_TAB,
+			isVisible: () => initialSample || this.state?.config.sample,
 			callback: () => this.setActiveTab(PLOTS_TAB)
 		})
 		if (state.termdbConfig.queries?.singleCell?.DEgenes) {
 			this.tabs.push({
 				label: 'Differential Expression',
 				id: DIFFERENTIAL_EXPRESSION_TAB,
 				active: activeTab == DIFFERENTIAL_EXPRESSION_TAB,
+				isVisible: () => initialSample || this.state?.config.sample,
```
siosonel, line 106: I think this should be `state.config.sample !== undefined`, and not have to use an `initialSample` variable.

client/plots/singleCellPlot.js
```diff
@@ -118,13 +120,15 @@ class singleCellPlot {
 			label: 'Gene Expression',
 			id: GENE_EXPRESSION_TAB,
 			active: activeTab == GENE_EXPRESSION_TAB,
+			isVisible: () => initialSample || this.state?.config.sample,
```
siosonel, line 123: I think this should be `state.config.sample !== undefined`, and not have to use an `initialSample` variable. Similar for the other `isVisible()` callbacks below.

client/plots/singleCellPlot.js
```diff
@@ -82,7 +82,7 @@ class singleCellPlot {
 			if (result == 1 || result == -1) return result
 			else return elem1.sample.localeCompare(elem2.sample)
 		})
-
+		const initialSample = state.config.sample != undefined
```
siosonel, line 85: I don't this variable is needed, could just check within the `isVisible()` callback.

client/plots/singleCellPlot.js
airenzp: Solved now

client/plots/singleCellPlot.js
```diff
@@ -84,6 +84,9 @@ class singleCellPlot {
 		})
 		this.tabs = []
 		const activeTab = state.config.activeTab
+		// shared isVisible function for tabs that require config.sample
+		const isVisible = () => this.state?.config.sample
```
airenzp, line 88: if the sample is preselected, like in the session, this would not capture it. That is why I check the init state

siosonel: if it's preselected, then it would be in the state as config.sample

siosonel: also, `showActiveTab()` is called in `main()`, so it would be checked/set against current state on every dispatch, not only on init

airenzp: but on init this.state is undefined, state uses the const state = this.getState(appState)

compbiolover (commented):
Overall this looks good. However, I noticed a layout issue: When viewing the UMAP and t-SNE plots side by side, then switching to the differential expression tab to run an analysis, the output table only takes up half of the sandbox width instead of the full width like the samples table does. While there's a resize handle in the corner of the table, it's not functional when dragged.

client/plots/singleCellPlot.js
airenzp: It would check the active tab, not all the tabs @siosonel

siosonel: Please let me know if there tab visibility is not working as expected, otherwise not sure if what you are pointing out is relevant. `self.update()` in `toggeButtons.js` sets tab display in line 292, so it's checking more than which tab is active.
```js
if (tab.isVisible) tab.wrapper.style('display', tab => (config && tab.isVisible() ? '' : 'none'))
```

compbiolover:
> > Overall this looks good. However, I noticed a layout issue: When viewing the UMAP and t-SNE plots side by side, then switching to the differential expression tab to run an analysis, the output table only takes up half of the sandbox width instead of the full width like the samples table does. While there's a resize handle in the corner of the table, it's not functional when dragged.
> 
> Thanks Andrew. I made the DE table resizable now. Also I have reduced the table heights for the GDC Frontend Framework as they add one extra heading and the long height on the tables makes them scroll

You're welcome. The table can now be resized to be smaller or larger. I didn't realize there was a limit on the width of the DE table. At about ~half way across the page the table can't be made larger. My previous comment can be discarded as I thought the DE table could be expanded to fill the full width of the page but I see that isn't the case

airenzp:
Thanks for the scrolling @siosonel . Please see my last commit with this [link](http://localhost:3000/?mass={%22genome%22:%22hg38%22,%22dslabel%22:%22BALL-scrna%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22singleCellPlot%22,%22sample%22:%22SJBALL030036_D1%22}]}) where a sample is preselected. It works if I add the state check

client/dom/table.ts
```diff
@@ -325,6 +325,12 @@ export function renderTable({
 						}
 					})
 
+				if (i === selectedRows[0]) {
+					setTimeout(() => {
```
xzhou82, line 329: could you comment on purpose?

siosonel (approved):

airenzp (commented):
Looks good thanks @siosonel

PR #2787, opened 2025-02-11
GSEA: Implemented FDR cutoff, top FDR pathways and number of permutations

robinpaul85:
## Description
Implemented FDR cutoff, top FDR pathways and number of permutations

## Checklist

### DIFF
--- a/client/plots/gsea.js
+++ b/client/plots/gsea.js
@@ -42,7 +42,7 @@ class gsea {
 		this.dom.controlsDiv.selectAll('*').remove()
 		const inputs = [
 			{
-				label: 'P-value Filter Cutoff (Linear Scale)',
+				label: 'FDR Filter Cutoff (Linear Scale)',
 				type: 'number',
 				chartType: 'gsea',
 				settingsKey: 'pvalue',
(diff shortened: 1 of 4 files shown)


client/plots/gsea.js
```diff
@@ -51,24 +51,40 @@ class gsea {
...
+				type: 'number',
 				chartType: 'gsea',
-				settingsKey: 'adjusted_original_pvalue',
-				title: 'Toggle between original and adjusted pvalues for volcano plot',
-				options: [
-					{ label: 'Adjusted', value: 'adjusted' },
-					{ label: 'Original', value: 'original' }
-				]
+				settingsKey: 'top_genesets',
+				title: 'Number of top gene sets to be displayed',
```
karishma-gangwani, line 58: can simply label as "top gene set cut-off"

client/plots/gsea.js
```diff
@@ -42,7 +42,7 @@ class gsea {
 		this.dom.controlsDiv.selectAll('*').remove()
 		const inputs = [
 			{
-				label: 'P-value Filter Cutoff (Linear Scale)',
+				label: 'FDR Filter Cutoff (Linear Scale)',
```
karishma-gangwani, line 45: just label as 'FDR' it is self-explanatory. also change the definition on the hover.

karishma-gangwani: include step size in FDR cut off for incremental increase by 0.1 units

client/plots/gsea.js
```diff
@@ -42,33 +42,47 @@ class gsea {
...
```
karishma-gangwani, line 45: if no. of gene sets is changed then changing FDR cutoff will not make any difference. The options should be provided as either-or using radio buttons. if 'FDR' is selected, then a user can input their FDR cutoff, else if 'top genesets' is selected then a user can input a number (say 10) and the top 10 genesets/pathways that show enrichment plus the top 10 showing downregulation should be printed in the output table after a ranking is performed on the basis of FDR.

client/plots/gsea.js
```diff
@@ -77,9 +73,44 @@ class gsea {
...
+				type: 'number',
+				chartType: 'gsea',
+				settingsKey: 'fdr_cutoff',
+				title: 'P-value significance',
+				min: 0,
+				max: 1
+			})
+		} else if (this.settings.fdr_or_top == 'top') {
+			inputs.push({
+				label: 'Number of top Gene Sets by FDR',
```
compbiolover, line 102: This seems a little unclear to me. The conditional is checking FDR or top but the label says it is top gene sets by FDR implying that FDR is being used in both. Also I think top should be capitalized to Top to match the casing of the rest of the label

karishma-gangwani (approved):
thanks @robinpaul85 please add comments in the code when you have a chance. and we can improve the UI sequentially based on everyone's inputs.

client/plots/gsea.js
karishma-gangwani: yes FDR is used by both but the Top will rank the genesets based on that and output only the top (number specified by the user) most significant ones in both the enriched and downregulated sets. Please check the latest commit.

karishma-gangwani (dismissed):

client/plots/gsea.js
compbiolover: No problem. The meaning was simply a little unclear to me. Thanks for clarifying

compbiolover (dismissed):

karishma-gangwani:
at gdc [link](http://localhost:3000/?mass=%7B%22genome%22:%22hg38%22,%22dslabel%22:%22GDC%22,%22nav%22:%7B%22header_mode%22:%22hidden%22%7D,%22plots%22:%5B%7B%22chartType%22:%22singleCellPlot%22,%22sample%22:%222409%22,%22experimentID%22:%229f155433-3c2e-4b67-a452-eb32f06c93f7%22%7D%5D%7D) the previous analyses for testing the MHC II example breaks currently when top geneset number is changed from 40 to 80, because key at line 239 of gsea.js becomes undefined.

robinpaul85:
> at gdc [link](http://localhost:3000/?mass=%7B%22genome%22:%22hg38%22,%22dslabel%22:%22GDC%22,%22nav%22:%7B%22header_mode%22:%22hidden%22%7D,%22plots%22:%5B%7B%22chartType%22:%22singleCellPlot%22,%22sample%22:%222409%22,%22experimentID%22:%229f155433-3c2e-4b67-a452-eb32f06c93f7%22%7D%5D%7D) the previous analyses for testing the MHC II example breaks currently when top geneset number is changed from 40 to 80, because 'key' at line 239 of gsea.js becomes undefined. can be fixed here or another pr.

My latest commit fixes that

karishma-gangwani:
Based on the discussion with @xzhou82 here are the suggested changes:

1. make Top gene sets as the default option in the radio button instead of FDR.
2. For option 2 (now being FDR), the default cutoff will still remain at 0.05 but it will no longer be applied/linked to the first option which is 'Top gene sets'
3. For the first default (Top gene sets) option now, rank the list of gene sets based on FDR and send the first 40 default significant genesets regardless of the NES scores.

Let me know if there are questions @robinpaul85

karishma-gangwani (approved):

xzhou82 (approved):

PR #2798, opened 2025-02-12
Implementation for making GSEA stateless

robinpaul85:
## Description
PLEASE read how to test, before testing!

Implementation for making GSEA stateless.

For mimicking statelessness on a local pp (which is technically not stateless) the pickle file needs to be deleted after the initial GSEA computation but BEFORE clicking on a row in the GSEA table. This can be achieved by running the following computation after the initial GSEA computation `rm /path/to/your/serverconfig/cachedir/*.pkl`

NOTE: In the event of a re-computation, the NES values will be slightly different than what is mentioned in the table. This is because blitzgsea uses permutations to calculate the pvalues and are not completely reproducible.

Minor fixes:

1) The GSEA button in the DE app has been restored.
2) In GSEA table, ES and Sidak columns have been commented out.
   

## Checklist

### DIFF
--- a/server/utils/gsea.py
+++ b/server/utils/gsea.py
@@ -24,6 +24,7 @@ try:
             cachedir = json_object['cachedir']  # Get the cache directory from the JSON object
             genes = json_object['genes']  # Get the genes from the JSON object
             fold_change = json_object['fold_change']  # Get the fold change values from the JSON object
+            num_permutations = json_object['num_permutations'] # Number of permutations for GSEA analysis
             table_name = json_object['geneset_group']  # Get the gene set group from the JSON object
             filter_non_coding_genes = json_object['filter_non_coding_genes']  # Get the filter_non_coding_genes flag from the JSON object
             db = json_object['db']  # Get the database path from the JSON object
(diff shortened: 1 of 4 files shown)


server/utils/gsea.py
```diff
@@ -84,8 +85,12 @@ def extract_symbols(x):
...
                 geneset_name = json_object['geneset_name']  # Get the gene set name from the JSON object
                 pickle_file = json_object['pickle_file']  # Get the pickle file name from the JSON object
-                result = pd.read_pickle(os.path.join(cachedir, pickle_file))  # Load the result from the pickle file
-                fig = blitz.plot.running_sum(signature, geneset_name, msigdb_library, result=result.T, compact=True)  # Generate the running sum plot
+                if os.path.isfile(os.path.join(cachedir, pickle_file)): # Check if the pickle file exists as it may not be in the same server that did the original GSEA computation
+                    result = pd.read_pickle(os.path.join(cachedir, pickle_file))  # Load the result from the pickle file
+                    fig = blitz.plot.running_sum(signature, geneset_name, msigdb_library, result=result.T, compact=True)  # Generate the running sum plot
+                else: # If pickle file is not found, redo the GSEA computation from scratch
+                    result = blitz.gsea(signature, msigdb_library, permutations=num_permutations).T  # Perform GSEA computation and transpose the result
```
xzhou82, line 92: on re-computing, it should also write the result into pickle file by the name, so that subsequent queries can benefit. otherwise there's no point in using pickle file

karishma-gangwani (approved):

PR #2800, opened 2025-02-12
improve joinUrl() to handle optional 3+ arguments and not require nes…

siosonel:
…ted joinUrl() calls

## Description

Tested with unit and integration tests.

## Checklist

### DIFF
--- a/shared/utils/src/joinUrl.js
+++ b/shared/utils/src/joinUrl.js
@@ -1,7 +1,14 @@
-function joinUrl(p1, p2) {
-	if (typeof p1 != 'string' || typeof p2 != 'string') throw `both arguments must be string type`
-	if (!p1 || !p2) throw 'blank string not allowed'
+function joinUrl(p1, ...p2) {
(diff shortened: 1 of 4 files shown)


shared/utils/src/joinUrl.js
```diff
@@ -1,7 +1,14 @@
-function joinUrl(p1, p2) {
```
xzhou82, line 1: why are both js and ts files needed?

siosonel: `proteinpaint/shared/utils` is published as `@sjcrh/proteinpaint-shared`, so all `ts` files in there needs to be built as `js` for packaging. If that's not done, it may not work if its excluded from bundling by nodejs/browser builds. This `js` transpilation requirement is somewhat similar to need to convert dataset `ts` files to `js` files for prod use.

xzhou82: thanks. i see it's explained in shared/utils/README.md

xzhou82 (approved):

PR #2802, opened 2025-02-13
Improved wait styling in the SC plot

airenzp:
## Checklist


xzhou82 (approved):
thanks looks nice
minor things that can be optimized
1. at DE panel, selecting a cluster should display the loading box too
2. when a gene exp has been selected, switch a sample, then go to DE tab, it will still trigger a wait, which is not necessary. i suppose the app can also clear the already selected gene on changing sample, same as DE cluster

PR #2804, opened 2025-02-13
Reduced default shape size and set min font to 0.7 in the scatter plot

### LINKED ISSUE #2696: samplescatter: dynamically calculate font size for each column separately
[test link](http://localhost:3000/?mass={%22dslabel%22:%22ASH%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22sampleScatter%22,%22name%22:%22Transcriptome%20TSNE%22,%22colorTW%22:{%22id%22:%22MolecularSubtype%22},%22shapeTW%22:{%22id%22:%22Sex%22}}]})

not reasonable Sex column is shown in such tiny font size

in each legend item, symbol (dot) size should't be constant and should scale with font size of that column

also the column header (term name) should be shown in a fixed font size and must not gets small as with category items

airenzp:
## Description

Closes #2696

## Checklist

### DIFF
--- a/client/plots/sampleScatter.renderer.js
+++ b/client/plots/sampleScatter.renderer.js
@@ -775,7 +775,7 @@ export function setRenderers(self) {
 		const top = 20
 		if (chart.colorLegend.size > top || chart.shapeLegend.size > top) {
 			fontSize = Math.min(0.9, top / chart.colorLegend.size, top / chart.shapeLegend.size)
-			if (fontSize < 0.5) fontSize = 0.5
+			if (fontSize < 0.7) fontSize = 0.7
 		}
 		return fontSize
 	}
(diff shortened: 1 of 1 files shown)


xzhou82:
thanks; i would like it to be further improved. these are not urgent
1. in the example, no need to keep sex legend font size "in sync" with molecular subtype; it should be calculated on its own
2. legend header font size should be calculated independently from legend items so it can be bigger
3. in the example deleting color tw and leaving only shape tw, the legend should not maintain a blank space for colortw

airenzp:
> thanks; i would like it to be further improved. these are not urgent
> 
> 1. in the example, no need to keep sex legend font size "in sync" with molecular subtype; it should be calculated on its own
> 2. legend header font size should be calculated independently from legend items so it can be bigger
> 3. in the example deleting color tw and leaving only shape tw, the legend should not maintain a blank space for colortw

I think now all these items are addressed. Please check.

xzhou82 (approved):

PR #2819, opened 2025-02-16
fix error when term1label is not a string

congyu-lu:
## Checklist


xzhou82:
thanks but not the fix i want. this chart makes no sense!
image



to do gene exp summary, the geneExpression tw.q.mode should be continous by default so as to show violins.

the

Issue #2828, opened 2025-02-18
numeric binary tw adjusting cutoff breaks


xzhou82:
[test link](http://localhost:3000/?mass={%22genome%22:%22hg38-test%22,%22dslabel%22:%22TermdbTest%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22regression%22,%22regressionType%22:%22logistic%22,%22outcome%22:{%22id%22:%22agedx%22}}]}), at outcome term edit menu, drag the line and server breaks

PR #2830, opened 2025-02-18
Convert custom binary value to numeric in binconfig

gavrielm:
## Description

Fixes https://github.com/stjude/proteinpaint/issues/2828

## Checklist

### DIFF
--- a/client/termsetting/handlers/numeric.binary.ts
+++ b/client/termsetting/handlers/numeric.binary.ts
@@ -263,7 +263,7 @@ function processCustomBinInputs(self) {
 	const inputDivs = self.dom.bins_table.node().querySelectorAll('input')
 	//let prevBin
 	const val = self.q.lst![0].stop // should not get value from dom.customBinBoundaryInput as value can be percentile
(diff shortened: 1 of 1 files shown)


karishma-gangwani (approved):
works for me.

client/termsetting/handlers/numeric.binary.ts
```diff
@@ -263,16 +263,17 @@ function processCustomBinInputs(self) {
 	const inputDivs = self.dom.bins_table.node().querySelectorAll('input')
 	//let prevBin
 	const val = self.q.lst![0].stop // should not get value from dom.customBinBoundaryInput as value can be percentile
+	if (!val) throw 'val is undefined'
```
siosonel, line 266: should be more strict, `if (!val && val !== 0) 'val is undefined`, should not throw on `0`, which is also an empty value

gavrielm: Thanks for noticing this. Fixed in new commit.

PR #2868, opened 2025-02-22
In GSEA fixed bug on download image and updated labels requested by GDC

airenzp:
## Description

Fix on the the image downloaded from the burguer menu icon,  it is updated every time a table row is selected but it needs to be cleared on main to avoid downloading a wrong image (jira-sv-2571) .

## Checklist

### DIFF
--- a/client/plots/gsea.js
+++ b/client/plots/gsea.js
@@ -185,6 +185,7 @@ class gsea {
 	async main() {
 		this.config = structuredClone(this.state.config)
 		this.settings = this.config.settings.gsea
+		this.imageUrl = null // Reset the image URL
 		await this.setControls()
 		if (this.dom.header)
 			this.dom.header.html(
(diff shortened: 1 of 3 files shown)


client/plots/gsea.js
```diff
@@ -185,6 +185,7 @@ class gsea {
 	async main() {
 		this.config = structuredClone(this.state.config)
 		this.settings = this.config.settings.gsea
+		this.imageUrl = null // Reset the image URL
```
siosonel, line 188: Should create a key to hold all instance props that will be cleared in `main()`, to make it easier to track/remember, and may be able to automate this later as part of `rx` dispatch cycle.

```js
   this.clearedProps = {.}
  
 main() {
    ...
    for(const k of Object.keys(this.clearedProps)) delete this.clearedProps[k]
    ..
 }
 
 // in other functions, can set arbitrary props under this.clearedProps
 ```

airenzp: This is not my plot. I just added the minimum fixes I needed to solve the jira ticket. As these are improvements I suggest we talk about them when you are back and we include Robin in the conversation? I am not sure I follow you completely. Aslo I dont want to take more time from you and I would rather minimize changes prone to errors as I am not familiar with the GSEA logic.

xzhou82 (approved):

PR #2872, opened 2025-02-24
Updated Survival Test to Work with Modern R Runner

compbiolover:
## Description
Updated `survival.R` test to work with `run_R.js`. Removed `hwe.R` and its related test as it is no longer used. Also removed an unused Fisher exact test now that it is done in Rust. Wrote a simple rounding script to round numeric values because the old expected file and the actual file had slightly different rounded values in the `upper` value (e.g. Expected: upper: 0.96245719974324 vs. Actual: upper: 0.962457199743239). This difference in rounding might be due to R versions etc. By rounding to 10 decimals for both before comparing we ensure we don't have superfluous test failures. 

## To Test
From `~/dev/sjpp/proteinpaint/server` run `npx tape -r '@babel/register' utils/test/Rscripts.spec.js`. All tests should pass. 

## Future Directions
Continue updating discussed tests in future PRs.

## Checklist


gavrielm (approved):

server/utils/test/Rscripts.spec.js
```diff
@@ -207,19 +185,61 @@ tape('km.R', async function (test) {
...
+	out = roundEstimates(out)
+	exp = roundEstimates(exp)
+
+	// Test if they match
+	test.deepEqual(out, exp, 'survival analysis results should match expected output')
 	test.end()
-})
 
+	// Function to round estimate values to fixed number of digits
+	function roundEstimates(data, digits = 10) {
```
karishma-gangwani, line 230: can this be moved to shared/utils/src/roundValue.js?

gavrielm: This function is specific to this test, so it is better to keep it here. I did point Andrew to `shared/utils/src/roundValue.js` so that he can see if that can be implemented here instead.

PR #2886, opened 2025-02-27
In scatter card omit host to use same host as client code and avoid issues

### LINKED ISSUE #2879: restore sjlife-related examples in app cards
e.g. http://localhost:3000/?appcard=Scatter%20Plot

airenzp:
## Description

Closes #2879

## Checklist

### DIFF
--- a/front/public/cards/scatterplot.json
+++ b/front/public/cards/scatterplot.json
@@ -37,7 +37,6 @@
         {
             "label": "Age and Heart",
             "runargs": {
-                "host": "https://survivorship.proteinpaint.stjude.org/",
                 "mass": {
                     "state": {
                         "vocab": {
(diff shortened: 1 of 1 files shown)


front/public/cards/scatterplot.json
```diff
@@ -37,7 +37,6 @@
         {
             "label": "Age and Heart",
             "runargs": {
-                "host": "https://survivorship.proteinpaint.stjude.org/",
```
xzhou82, line 40: cannot delete host; sjlife is no longer hosted on prp2

airenzp: I would think that if I omit the host it will use the host of the server where it runs, it does not mean is going to fail. But maybe I am missing something?

airenzp: You want this example to work on the public server I see. Then discard this PR and once we deploy to the survivorship server it should work

PR #2891, opened 2025-02-27
Sort violin plots by median 

creilly8:
## Description
Closes issue #2786. 

Changes: 
If more than one violin plot is present, a new `Order by` control appears. This allows the user to sort by the default order returned from the server or median value in descending order. 

Test: 
Any violin plot from http://localhost:3000/url.html. Please test that the control ought not to appear when only one plot is present and that it works properly otherwise. 

## Checklist

### DIFF
--- a/client/plots/violin.js
+++ b/client/plots/violin.js
@@ -145,6 +145,18 @@ class ViolinPlot {
 				]
 			},
 			{
+				label: 'Order by',
+				title: 'Order violin plots by parameters',
+				type: 'radio',
+				chartType: 'violin',
+				settingsKey: 'orderByMedian',
+				options: [
+					{ label: 'Default', value: false },
+					{ label: 'Median', value: true }
+				],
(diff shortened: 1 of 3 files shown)


karishma-gangwani (approved):
works nicely. thanks @creilly8 

this is separate but there's a color option in the burger menu which doesn't really work for multiple categories of the violin. Should we only show it when there's no overlay term?

client/plots/violin.js
```diff
@@ -144,6 +144,18 @@ class ViolinPlot {
...
 				]
 			},
+			{
+				label: 'Order by',
+				title: 'Order violin plots by parameters',
+				type: 'radio',
+				chartType: 'violin',
+				settingsKey: 'orderByMedian',
+				options: [
+					{ label: 'Default', value: false },
```
xzhou82, line 154: need a few new options

xzhou82:
may also allow to sort by sample size
if a categorical term is ordinal needs to observe the category order too. need to find an example

PR #2898, opened 2025-02-28
Tentative implementation for quasi-likelihood pipeline in edgeR

robinpaul85:
## Description

Implementation for quasi-likelihood pipeline in edgeR.

Notes:
1) The QL pipeline is quite fast than the maximum likelihood (as expected).

2) The edgeR author recommended to use QL pipeline irrespective of sample size. So, only the R code was changed without any change in server routes or client side code.

3) Added support to show QL dispersion plot and MDS plot. Currently saves the image to serverconfig.cachedir. The MDS plot saved in cachedir does not render correctly. But that does not break anything. I shall fix that in my following branch. 

     The images are deleted once they are read in nodejs. Please comment out fs.unlink statement in line 280 of server/routes/termdb.DE.ts to see the png files externally. 

## Checklist


server/utils/edge.R
```diff
@@ -156,27 +160,29 @@ if (length(input$conf1) == 0) { # No adjustment of confounding factors
...
-    dispersion_time <- system.time({
-        y <- estimateDisp(y, design)
-    })
-    #cat("Dispersion time: ", dispersion_time[3], " seconds\n")
-
     fit_time <- system.time({
-        fit <- glmFit(y, design)
+        suppressWarnings({
+            suppressMessages({
+                fit <- glmQLFit(y,design)
```
karishma-gangwani, line 173: could you clean up the code a bit, and add a note about how QLfit works and what it is doing differently from the regular glm fit (can go at the top of the script).

karishma-gangwani:
ok, I looked more closely into this dispersion effects by specific genes. After you do the fitting, we can check for the gene-specific dispersion. Probably we can print it as a table on the backend and see or add it in the UI table as a quality check for the users (something we would need to discuss with Xin for the UI related). Because if we need to include in a filtering step based on overdispersion we could do that in case we do see very high dispersions but I want to crosscheck. You could do something like this and we can test it out:

# Inspect dispersion estimates
summary(dge$tagwise.dispersion)

# which genes show high dispersion??
high_dispersion_genes <- which(dge$tagwise.dispersion > 1)

server/utils/edge.R
```diff
@@ -115,27 +115,31 @@ filter_time <- system.time({
 
 normalization_time <- system.time({
     y <- y[keep, keep.lib.sizes = FALSE]
-    y <- calcNormFactors(y, method = "TMM")
+    y <- normLibSizes(y, method="TMM") # Using TMM method for normalization
```
karishma-gangwani, line 118: you don't need to use TMM normalization here anymore. Please remove the parameter.

server/utils/edge.R
```diff
@@ -115,27 +115,42 @@ filter_time <- system.time({
...
 })
 #cat("Normalization time: ", normalization_time[3], " seconds\n")
 
+# Saving MDS plot image
+set.seed(as.integer(Sys.time())) # Set the seed according to current time
+cachedir <- input$cachedir # Importing serverconfig.cachedir
+random_number <- runif(1, min = 0, max = 1) # Generating random number
+mds_image_name <- paste0("edgeR_mds_temp_",random_number,".png") # Generating random image name so that simultaneous server side requests do NOT generate the same edgeR file name
+png(filename = paste0(cachedir,"/",mds_image_name), width = 1000, height = 1000, res = 200) # Opening a png device
+par(oma = c(1, 1, 1, 1)) # Creating a margin
```
karishma-gangwani, line 128: can they both use the same margins instead of repeating?

robinpaul85: I will put them into a common function in my following branch.

karishma-gangwani (dismissed):
currently supports downloading the MDS and QL dispersion plots in the cache directory. MDS plot still needs some work and would be done in a separate pr along with displaying both on the UI next to the volcano.  Thanks @robinpaul85 and @creilly8

karishma-gangwani:
There is also a bug with using numeric discrete terms for grouping as well as a confounding variable. This error message handling fix will also be done in the next pr @robinpaul85

xzhou82 (approved):

PR #2911, opened 2025-03-04
Watch server.ts on start

aacic:
## Description

`npm run dev1` is broken on master.

This is a fix. 

To test, from pp dir:

`npm run dev1` on:

- hostmachine

- dev container

Run inside of sjpp:

`npm run dev`

## Checklist


server/src/serverconfig.js
```diff
@@ -6,12 +6,17 @@
 import fs from 'fs'
 import path from 'path'
 
+// import.meta.dirname is undefined when using docker dev environment
+const url = import.meta.url
+const directoryURL = new URL('.', url)
+const directoryPath = directoryURL.pathname
```
siosonel, line 12: Should be simplified as `const directoryPath = import.meta.dirname || (new URL('.', import.meta.urll)).pathname`.

Also, not sure why `import.meta.dirname` is undefined in dev container, since it uses the same `ppbase` image that `ppserver` and `ppfull` image use, and `server/src/app.ts` works in those containers.  I think there is an underlying bug that is specific to the dev container, that should be addressed, instead of modifying `app.ts` and `serverconfig.js`.

siosonel (approved):

siosonel:
@aacic Please pull. my changes and test in dev container. If it works, please rebase/merge.

PR #2912, opened 2025-03-05
Use (max-min) if p98 == p5 to compute bin size

gavrielm:
## Description

In `shared/utils/src/termdb.initbinconfig.js`, if 98th percentile == 5th percentile, use max and min to compute bin size. Solves issue when data contains mostly one value.

## Checklist


airenzp (approved):
Thanks. We were troubleshooting the edge cases and this solution and looks good to me. It would work like before when the percentiles can't be used.

PR #2937, opened 2025-03-10
Violin using R density

airenzp:
## Description

We decided to use the R density method to ensure we render a proper violin from a trusted library. Now the density is calculated using  getDensity in termdb.violin.js that runs the density.R script. getDensity replaces getBinsDensity and returns the density as before. The filter and the termsetting editor are updated to use this density. Updated tests that now need to wait one more second before timing out. 

## Checklist


server/routes/correlationVolcano.ts
```diff
@@ -144,3 +143,14 @@ export function validate_correlationVolcano(ds: any) {
 		throw 'unknown cv.variables.type'
 	}
 }
+
+export function mean(data) {
```
creilly8, line 147: @airenzp, We probably should move these to a new #shared/utils/src/stats or something similar. That way others can find these functions easier. What do you think?

siosonel: Definitely not in `#shared/` dir, the code there should be able to run in both nodejs and browser, and R code is only available in the server/utils right now. The only way to share between nodejs and browser is to use web assembly, which may be too much work.

airenzp: Yes we can move these functions to another place to make them more visible and reusable. Let me know what would be the right place.

siosonel: @creilly8 ah, ok,  I didn't notice the comment was directly under `mean()`, that makes sense now, and I do see that function in `shared`
```text
/Users/esioson/dev/sjpp/proteinpaint/shared/utils/src/descriptive.stats.js:
   16  
   17  	//compute mean
   18: 	function mean(arr) {
   19  		if (arr.length === 0) throw new Error('No data provided')
   20  		return arr.reduce((a, b) => a + b) / n
```

creilly8: @siosonel Ok, perfect. I can make the change to use the above and move `stdev` to that file in a separate PR so Airen isn't held up.

siosonel: @creilly8 that reduce function can stay outside of `mean()`, so that it's reused instead of creating a new function every call to `mean()`, so
```js
function mean(arr) {
   if (arr.length === 0) throw new Error('No data provided')
   return arr.reduce(getArrSum) /n
}

// reusable array reducer function, instead of creating a new one every time,
// and since there is no closured variable inside mean() that's used in the reducer 
function getArrSum(a, b) { 
  return a + b
}
```

creilly8: @siosonel. Sounds good.

gavrielm: Note that `stdev` is also computed in `descriptive.stats.js`:
```
const squareDiffs = arr.map(x => (x - mean_arr) ** 2).reduce((a, b) => a + b, 0)
// compute variance
const variance = squareDiffs / (n - 1)
// compute standard deviation
const stdDev = Math.sqrt(variance)
```

May consider merging into a single `stdev` function.

airenzp: Sounds good, thank you all. Now the median is fixed so this PR should be ready to merge.

creilly8 (approved):

xzhou82:
[sjlife violin+condition term](http://localhost:3000/?mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22hrtavg_s%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22Auditory%20System_s%22}}],%22nav%22:{%22activeTab%22:1}}) breaks wilcox rust code, the boxplot of this works but shows a group `undefined n=1`

termdbtest violin+condition works, even after changing minSampleCutoff from 5 to 1 so 2+ violins are rendered and triggered wilcox rust.

@gavrielm might be this is sjlife data issue and unrelated to code changes here? if so can merge this and fix the breakage in new branch

server/utils/density.R
```diff
@@ -0,0 +1,13 @@
+library(jsonlite)
```
xzhou82, line 1: would like a comment line e.g. `Rscript density.R <some test data here>`

gavrielm:
> sjlife violin+condition term

@xzhou82 I am not seeing an error when running [sjlife violin+condition term](http://localhost:3000/?mass=%7B%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22id%22:%22hrtavg_s%22,%22q%22:%7B%22mode%22:%22continuous%22%7D%7D,%22term2%22:%7B%22id%22:%22Auditory%20System_s%22%7D%7D%5D,%22nav%22:%7B%22activeTab%22:1%7D%7D)

airenzp:
Me neither. Both links work for me. I added a comment on how to run the script after checking it it with @gavrielm. We also discussed the possibility of speeding up the rendering by running a single R process when there are several categories. I can do that in another PR.

server/utils/density.R
```diff
@@ -1,13 +1,18 @@
 library(jsonlite)
 library(ggplot2)
+# This script reads in a json string from stdin, calculates the density of the data and returns the density as a json string
+# The input json string should be a list of numbers
```
gavrielm, line 4: Should be: "The input json string is an array of numbers"

server/utils/density.R
```diff
@@ -1,13 +1,18 @@
...
+# The output json string will be a dictionary with two keys x and y that are number arrays
+# The x array will contain the x values of the density
+# The y array will contain the y values of the density
```
gavrielm, line 7: May be simplified to: "The output json string is an object {x: [x values of the density], y: [y values of the density]}"

server/utils/density.R
```diff
@@ -1,13 +1,18 @@
...
 con <- file("stdin", "r")
 json <- readLines(con)
 close(con)
 data <- fromJSON(json)
 den = density(x = data)
 x = den$x
 y = den$y
-
-toJSON(list(x=x, y=y), digits = NA, na = "string") # will  return a json like {x:[...], y: [...]}
+result = list(x=x, y=y)#dictionary with two keys x and y that are number arrays
```
gavrielm, line 17: The data is still in R format here, so can revise the comment to say: `# list with two keys x and y that are number vectors`

xzhou82:
thanks my sjlife db is out of date and that probably is the cause

server/utils/density.R
```diff
@@ -1,18 +1,19 @@
...
 # In order to test it you can run from the command line: echo "[1.2, 2, 3]" | Rscript ./density.R
 
 con <- file("stdin", "r")
 json <- readLines(con)
 close(con)
-data <- fromJSON(json)
+valuesPerPlot <- fromJSON(json)
+for(i in 1:length(valuesPerPlot)){
+  data = valuesPerPlot[[i]]
+}
```
gavrielm, line 14: Looks like this code change causes violin plot to break. Why is this change needed?

airenzp: Sorry I added those changes by mistake, they are reverted now. I was trying to test passing all the arrays to calculate all the densities  in a single script.

gavrielm (approved):
Looks good. In next PR, can implement density calculation for multiple plots within a single R process.

@airenzp make sure to rebuild the deps image in docker because `ggplot2` is now added as a new dependency.

xzhou82:
great idea to run multiple density calculations in one r call. thanks Gavriel!

PR #2938, opened 2025-03-10
Prototype implementation of mds and ql images

robinpaul85:
## Description
1) Fixed edgeR mds plot render. Now case samples are labelled as "T" and control samples are labelled as "C". This is to avoid overwriting of labels.

2) Removed option for "Variable gene cutoff" as it gives inaccurate results.

3) Now filtering for genes using cpm. May add these as UI parameters in the future.
## Checklist

### DIFF
--- a/client/plots/volcano/VolcanoControlInputs.ts
+++ b/client/plots/volcano/VolcanoControlInputs.ts
@@ -137,7 +137,7 @@ export class VolcanoControlInputs {
 				settingsKey: 'varGenesCutoff',
 				title: 'Top number of genes with the highest variability to include in analysis',
 				min: 1000,
-				max: 4000
+				max: 7000
 			},
 			{
 				label: 'Rank Genes by',
--- a/client/plots/volcano/VolcanoControlInputs.ts
+++ b/client/plots/volcano/VolcanoControlInputs.ts
@@ -130,15 +130,6 @@ export class VolcanoControlInputs {
 				min: -10,
 				max: 10
 			},
-			// {
-			// 	label: 'Variable genes cutoff',
-			// 	type: 'number',
-			// 	chartType: 'volcano',
-			// 	settingsKey: 'varGenesCutoff',
-			// 	title: 'Top number of genes with the highest variability to include in analysis',
-			// 	min: 1000,
-			// 	max: 7000
-			// },
 			{
 				label: 'Rank Genes by',
 				type: 'radio',
--- a/client/plots/volcano/VolcanoControlInputs.ts
+++ b/client/plots/volcano/VolcanoControlInputs.ts
@@ -130,15 +130,15 @@ export class VolcanoControlInputs {
 				min: -10,
 				max: 10
 			},
-			{
-				label: 'Variable genes cutoff',
-				type: 'number',
-				chartType: 'volcano',
-				settingsKey: 'varGenesCutoff',
-				title: 'Top number of genes with the highest variability to include in analysis',
-				min: 1000,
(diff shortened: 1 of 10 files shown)


karishma-gangwani (approved):

server/utils/edge.R
```diff
@@ -119,14 +126,28 @@ normalization_time <- system.time({
...
 
+# Cutoffs for cpm, will add them as UI options later
+if (length(samples_indices) > 100) {
+    gene_cpm_cutoff <- 15
+    sample_cpm_cutoff <- 30
+    count_cpm_cutoff <- 100000
+} else {
+    gene_cpm_cutoff <- 5
+    sample_cpm_cutoff <- 15
+    count_cpm_cutoff <- 100000
```
karishma-gangwani, line 137: you could remove the repetition later since in both blocks its the same value.

client/plots/volcano/VolcanoControlInputs.ts
```diff
@@ -137,7 +137,7 @@ export class VolcanoControlInputs {
 				settingsKey: 'varGenesCutoff',
 				title: 'Top number of genes with the highest variability to include in analysis',
 				min: 1000,
-				max: 4000
+				max: 7000
```
karishma-gangwani, line 140: We should also remove this variance cutoff now. and perhaps work on enabling DEseq2 for higher sample sizes.

creilly8: @karishma-gangwani, I'll do that in this PR not this one https://github.com/stjude/proteinpaint/pull/2944.

server/utils/edge.R
robinpaul85: I could, but we may want to have different cutoffs for different sample sizes.

client/plots/volcano/view/VolcanoPlotView.ts
```diff
@@ -25,9 +25,15 @@ export class VolcanoPlotView {
 		this.settings = settings
 		this.viewData = viewData
 		const actions = this.dom.holder.append('div').attr('id', 'sjpp-volcano-actions').style('display', 'block')
+		const images = this.dom.holder
+			.append('div')
+			.style('display', 'inline-block')
+			.style('vertical-align', 'top')
+			.attr('id', 'sjpp-volcano-images')
```
xzhou82, line 32: should use `data-testid` but not id

creilly8: @xzhou82, the ids in this case are used elsewhere, not just for test scripts.

karishma-gangwani:
Are the images still supposed to get cached in the cache dir if confounders are applied to an existing visualization??

creilly8:
> Are the images still supposed to get cached in the cache dir if confounders are applied to an existing visualization??

@karishma-gangwani, I don't see the images in the cache dir but I do see the images not changing when adding confounders. Is that what you are referring to?

karishma-gangwani:
> > Are the images still supposed to get cached in the cache dir if confounders are applied to an existing visualization??
> 
> @karishma-gangwani, I don't see the images in the cache dir but I do see the images not changing when adding confounders. Is that what you are referring to?

okay, the caching is not taking place, that's fine. I do see it get logged on the server side with the path. Maybe we can clean that up? And I see the volcano changes, I don't expect the MDS or the dispersion plot to change with adding the confounders. Dispersion may or may not change. But that we can continue to test in later pr's.

xzhou82 (approved):

PR #2944, opened 2025-03-11
Edge r plot2

robinpaul85:
## Checklist

### DIFF
--- a/client/plots/volcano/VolcanoControlInputs.ts
+++ b/client/plots/volcano/VolcanoControlInputs.ts
@@ -137,7 +137,7 @@ export class VolcanoControlInputs {
 				settingsKey: 'varGenesCutoff',
 				title: 'Top number of genes with the highest variability to include in analysis',
 				min: 1000,
-				max: 4000
+				max: 7000
 			},
 			{
 				label: 'Rank Genes by',
--- a/client/plots/volcano/VolcanoTypes.ts
+++ b/client/plots/volcano/VolcanoTypes.ts
@@ -62,6 +62,7 @@ export type VolcanoPlotDom = {
 	yAxis: SvgG
 	/** Y axis label */
 	yAxisLabel: SvgText
+	images: any
 }
 
 /** Settings for the differential analysis volcano */
@@ -102,4 +103,5 @@ export type VolcanoViewData = {
 	plotDim: VolcanoPlotDimensions
 	pointData: DataPointEntry[]
 	pValueTableData: any
+	images: any
 }
--- a/client/plots/volcano/view/VolcanoPlotView.ts
+++ b/client/plots/volcano/view/VolcanoPlotView.ts
@@ -118,9 +118,8 @@ export class VolcanoPlotView {
 		const holder = this.dom.holder
 			.append('div')
 			.attr('id', 'sjpp-volcano-stats')
-			.style('display', 'inline-block')
+			.style('display', 'block')
 			.style('vertical-align', 'top')
-			.style('margin-top', '50px')
 		const table = table2col({ holder })
 		for (const d of statsData) {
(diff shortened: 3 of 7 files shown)


karishma-gangwani:
for implementing the CPM cutoff:

pg 57 - BCV plot 

pg 96 - filtering documentation for CPM

pg 56 - common/global dispersion

Basically....

## before your matrix model is designed:

`table(y$samples$group)`

# Filter out lowly expressed genes
## Filter out genes with CPM greater than 5 in 10 or more samples and atleast 100,000 reads (dropped some 4k genes) (for a smaller sample size <100). This affects the global dispersion value. It should decrease but not so that the increasing trendline starts to show a higher dispersion way above the global dispersion. A global dispersion <0.5 is generally a good indicator of a good filtering strategy (print y$common.dispersion to get the value of global dispersion, check documentation for this on pg 56 of edgeR).

```
print("y$counts0:")
print(nrow(y$counts))
```

```
selr <- rowSums(cpm(y$counts)>5)>=15
selc <- colSums(cpm(y$counts))>=100000
```

## for a larger sample size >100 Filter out genes with CPM greater than 15 in 30 or more samples:

```
selr <- rowSums(cpm(y$counts)>15)>=30
selc <- colSums(cpm(y$counts))>=100000
```

## generate y again with the new filtered genes:
`y <- y[selr, selc]`

## can print nrow(y$counts) before and after to check how many total gene rows are being dropped. 
```
print("y$counts1:")
print(nrow(y$counts))
```

##I have additionally generated BCV plots. They are also in the ppt. and help assess whether CPM cutoff has been applied accurately or not. 

##code for bcv plot:

## Extract BCV data
```
bcv_data <- data.frame(
  AveLogCPM = y$AveLogCPM,
  BCV = sqrt(y$tagwise.dispersion)
)
```

# Extract the common dispersion
`common_bcv <- sqrt(y$common.dispersion)`

# Add a smoothed trend lineand common dispersion line
```
bcv_plot_with_smooth <- ggplot(bcv_data, aes(x = AveLogCPM, y = BCV)) +
  geom_point(alpha = 0.5) +
  geom_hline(yintercept = common_bcv, color = "red", linetype = "dashed", linewidth = 1) +  # Common dispersion line
  geom_smooth(method = "loess", color = "blue", se = FALSE, linewidth = 1) +  # Smoothed trend line with updated linewidth
  theme_minimal() +
  labs(title = "BCV Plot with Common Dispersion and Smoothed Trend",
       x = "Average Log CPM",
       y = "Biological Coefficient of Variation (BCV)")
```

## Save the plot
`ggsave("~/path/BCV_plot_with_smoothed_trend.png", plot = bcv_plot_with_smooth, width = 8, height = 6)
`

To summarize:

1. MDS plot allows user to check sampling and grouping criteria
2. BCV plots help users understand if CPM cutoffs are correctly applied (so this could very well become a UI option)
3. QL dispersion plot shows the degree of dispersion. 

Either way, my computation became faster (haven't calculated the runtime) but this should help speed up the process. For 1000+ samples I haven't run any test yet. Those would require even a more stringent cpm cutoff I am assuming. We can keep a default for speeding up the computation and remove the global variance cut off)

Please crosscheck the documentation. This is based on my test outside of PP.

client/plots/volcano/VolcanoControlInputs.ts
```diff
@@ -137,7 +137,7 @@ export class VolcanoControlInputs {
 				settingsKey: 'varGenesCutoff',
 				title: 'Top number of genes with the highest variability to include in analysis',
 				min: 1000,
-				max: 4000
+				max: 7000
```
karishma-gangwani, line 140: we should remove this variance cutoff and perhaps work towards enabling DEseq2 now.

PR #2945, opened 2025-03-11
on api failure, GDC cache init will retry (Jira sv 2590)

siosonel:
## Description

This fix will force the initial GDC caching to be retried, regardless of error being fatal or recoerable.

To test, need a way to simulate the case where GDC API is not ready yet when the PP server starts up. Suggested temp changes to simulate
- add `let numTries = 0` in line 10
- change to `joinUrl(host.rest, numTries < 3 ? '--files' : 'files'); numTries++` in line 141
- change to `5000` in line 108, so `cacheCheckWait` doesn't too long to observer caching retry loop
- save and when the pp server restarts, there should be two times when `"force retries of cacheMappingOnNewRelease()"` will be logged, then `GDC: cache is stale. Re-caching... 0` to proceed to the normal caching.

## Checklist


xzhou82:
`in line 141 change to "joinUrl(host.rest, numTries < 3 ? 'invalidEndpoint' : 'files'); numTries++"`
can add this and related comments to code as self-reminder on how to simulate failed api

congyu-lu (approved):

PR #2947, opened 2025-03-12
Vioin r speedup

airenzp:
## Description

When the violin is divided into categories calculate all of them in the same R process. When all values are the same do not plot a violin. Reduced timeouts increased now that the violin with categories is rendered faster.

## Checklist

### DIFF
--- a/server/utils/density.R
+++ b/server/utils/density.R
@@ -1,18 +1,29 @@
 library(jsonlite)
 library(ggplot2)
-# This script reads in a json string from stdin, calculates the density of the data and returns the density as a json string
-# The input json string is an array of numbers
-# The output json string is an object with the following fields {x: [x density values], y: [y density values]}
-# In order to test it you can run from the command line replacing the array with your own: echo "[1.2, 2, 3]" | Rscript ./density.R
+# This script reads in a json string from stdin, calculates the densities of each plot and returns the densities as a json string
+# The input json string is a dictionary where each field maps to an array of numbers
+# The output json string is a dictionary  with the density for each plot. The density is represented like {x: [x density values], y: [y density values]}
+# In order to test it you can run this from the command line replacing the arrays with your own: 
+# echo "{plotA: [1.2, 2, 3], plotB: [4.5, 5, 6]}" | Rscript ./density.R
 
 con <- file("stdin", "r")
 json <- readLines(con)
 close(con)
 data <- fromJSON(json)
-den = density(x = data)
-x = den$x
-y = den$y
-result = list(x=x, y=y) #This will be read in javascript like a dictionary with two keys x and y that are number arrays
-toJSON(result, digits = NA, na = "string") # will  return a json like {x:[...], y: [...]}
+densities <- list()
+for(plot in names(data)){
+    values = data[[plot]]
+    if(length(values) <= 5){
+        y = rep(0, length(values))
+        densities[[plot]] <- list(x=values, y=y)
+        next
+    }
+    den = density(x = values)
+    x = den$x
+    y = den$y
+    result = list(x=x, y=y) #This is an object  with two keys x and y that are number arrays
+    densities[[plot]] <- result
+}
+toJSON(densities, digits = NA, na = "string") # will  return a json like { plotA: {x:[...], y: [...]}}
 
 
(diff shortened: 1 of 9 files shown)


gavrielm:
For this [link](http://localhost:3000/?mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22hrtavg_s%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22racegrp_s%22}}],%22nav%22:{%22activeTab%22:1}}), the `Unknown` category has only zero values but the rendered median value is a non-zero value:

image

server/utils/density.R
```diff
@@ -1,18 +1,30 @@
...
-y = den$y
-result = list(x=x, y=y) #This will be read in javascript like a dictionary with two keys x and y that are number arrays
-toJSON(result, digits = NA, na = "string") # will  return a json like {x:[...], y: [...]}
+densities <- list()
+for(plot in names(data)){
+    values = data[[plot]]
+    # If the plot has less than 5 values or all the values are the same, we will return a flat line
+    if(length(values) <= 5 | length(unique(values)) == 1){
+        y = rep(0, length(values))
+        densities[[plot]] <- list(x=values, y=y)
```
gavrielm, line 19: It seems like this will create a flat line at `0`, even if values are duplicated non-zero values.

For example, for this [link](http://localhost:3000/?mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22hrtavg_s%22,%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22id%22:%22racegrp_s%22}}],%22nav%22:{%22activeTab%22:1}}), when I hardcode `plot2Values['Unknown'] = [ 10, 10, 10, 10, 10, 10 ]` in `termdb.violin.ts` the plot will still show a flat line at 0:

image

xzhou82:
inspecting additional examples now

server/utils/density.R
karishma-gangwani: Yes, the medians are not rendering correctly to the density. The plots sped up as required.

airenzp:
Thank you all for the comments. Hopefully now all the issues are solved. Thanks @gavrielm for your help with R, passing the min and max to the density helped to solve all the scaling issues and I could cleanup the logic related. 
As you pointed out the ggplot2 library is not needed to use the density function (it was used more for the violin and for plotting). I am keeping it though in the Dockerfile since Robin is planning to use it.

server/utils/density.R
gavrielm: Medians are now rendering correctly, but issue I pointed out above it still present.

image

airenzp: > It seems like this will create a flat line at `0`, even if values are duplicated non-zero values.
> 
> For example, for this [link](http://localhost:3000/?mass=%7B%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22id%22:%22hrtavg_s%22,%22q%22:%7B%22mode%22:%22continuous%22%7D%7D,%22term2%22:%7B%22id%22:%22racegrp_s%22%7D%7D%5D,%22nav%22:%7B%22activeTab%22:1%7D%7D), when I hardcode `plot2Values['Unknown'] = [ 2000, 2000, 2000, 2000, 2000, 2000 ]` in `termdb.violin.ts` the plot will still show a flat line at 0:
> 
> image

You need to also overwrite plot.values to see the bins change

siosonel:
@airenzp 
fyi, related to speed up, I pushed parallelized `getDescrStats()` and `getViolinPlotData()` to master, to fix the integration tests

server/utils/density.R
gavrielm: I see ok thanks. The grey line that represents the plot gets hidden by the red median line. Maybe in next PR can make grey line taller so that user can see all values are the same.

gavrielm (dismissed):

xzhou82 (approved):

PR #2951, opened 2025-03-13
Improved median and samples opacity in violin

airenzp:
## Description

As pointed out by @gavrielm the median should have opacity. Also improved lines visualization by adding some more opacity. When using circles to represent samples they were being filled black when few samples, removed to use same appearance in all plots. Improved symbol size as 3 makes the circles to be clipped.
Before
Screenshot 2025-03-13 at 4 32 49 PM
After
Screenshot 2025-03-13 at 4 33 26 PM

## Checklist


gavrielm:
I wasn't necessarily thinking that the median line should have opacity. I was thinking that the straight vertical line of plot with duplicated values should be taller than the median line.

airenzp:
> I wasn't necessarily thinking that the median line should have opacity. I was thinking that the straight vertical line of plot with duplicated values should be taller than the median line.

Discussed with @gavrielm and we agreed on using some opacity for the median.

gavrielm (approved):
Higher opacity looks nice. Thanks for implementing this.

xzhou82 (approved):

PR #2957, opened 2025-03-14
fix: detect debug lines from R output and console.log

### LINKED ISSUE #2931: r server log: debug mode-only logic to split output and do string detection in lines
output debugging lines must start with special string `debugging` to be detected and logged out, and excluded from r output

using debugmode avoids running such logic in prod

**big catch:**
 prod will break if an r script is merged with debugging lines! because the check is disabled on prod and such lines go into output
solution is to add linting to check for it

xzhou82:
… prod

## Description

closes #2931 

to test, add lines 276, 278 and 279 to server/utils/cuminc.R, and [test link](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22plots%22:[{%22chartType%22:%22cuminc%22,%22settings%22:{%22cuminc%22:{}},%22term%22:{%22id%22:%22Cardiovascular%20System%22,%22q%22:{%22breaks%22:[1]}}}],%22nav%22:{%22activeTab%22:1}}) to see the debug line outputted in server console

```R
276 t1<-Sys.time()
277 ci_results <- mclapply(X = dat, FUN = run_cuminc, startTime = startTime, mc.cores = cores)
278 t2<-Sys.time()
279 cat("debug: time:",difftime(t2,t1,units='secs'),"\n")
```

please evaluate if this is good approach and verify all portal features using R are working

## Checklist

### DIFF
--- a/server/src/run_R.js
+++ b/server/src/run_R.js
@@ -15,8 +15,18 @@ import serverconfig from './serverconfig.js'
 import { spawn } from 'child_process'
 import { Readable } from 'stream'
 
-// hardcoded debug line marker, to identify which lines of Rscript stdout are debugging/diagnostic and to be separated from actual output e.g. stringified JSON
-// NOTE that a debugging line must ends with line break
+/*
+hardcoded debug line marker, to identify which lines of Rscript stdout are debugging/diagnostic,
+and to be separated from actual output e.g. stringified JSON or any text
+
+**********
+** NOTE **
+**********
+- a debugging line must ends with line break
+- a normal line MUST NOT begins with the marker, or it will be skipped!!
+
+TODO when R is moved into workspace, add these to README over there also!!
+*/
 const debugLineMarker = 'debug:'
 
 export default async function run_R(path, data, args) {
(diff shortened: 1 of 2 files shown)


karishma-gangwani (approved):
I tested survival, regression and edgeR. printing with 'debug:' seems to work for me. Thanks for the fix. Let's wait for Gav's or Robin's comments.

server/src/run_R.js
```diff
@@ -60,7 +64,33 @@ export default async function run_R(path, data, args) {
...
+						console.log('<R>', line)
+						continue
+					}
+					// line doesn't begin with marker and is kept
+					actualOutputLines.push(line)
+				}
+				resolve(actualOutputLines.join('\n'))
+			} else {
+				// not dev environment; do this detection to guard against accidental introduction of debug lines
+				if (stdout.indexOf(debugLineMarker) != -1) {
```
robinpaul85, line 87: Isn't this still string search? If you have to string search why not this implementation https://github.com/stjude/proteinpaint/compare/master...DE_test_external_scripts

Secondly, even if we do decide to add string search in prod environment, stdout.startsWith is safer since its only doing string search at the beginning of the string search.

robinpaul85: Although its has been mentioned in the comments, the risk still remains with this implementation of somebody committing an R script with uncommented out print lines. Maybe add some CI check that can flag this issue while running tests in a PR?

xzhou82: please check latest change

robinpaul85: Thank you for addressing my concern, however its still difficult to print entire objects (which can be thousands of line) because its hard to add "debug:" to every line.

gavrielm: Yes it will not be straightforward to add `debug:` to every line of multi-line objects. Also, another concern is that `print()` will not work with this method because by default it prepends `[number]` to its output, which prevents us from prepending `debug:`. As a result, we need to use `cat()` and this does not work with some objects. For example, adding this to `cuminc.R`:
```
cat("debug:", input, "\n")
```
will cause this error:
image

xzhou82: is it possible that an object such as `input` can be serialized into json to print?

gavrielm: Yes, but not all R objects can be serialized. For example, if we want to examine a regression model object, that cannot be serialized, but it can be viewed using `print()`.

PR #3005, opened 2025-03-21
Added support for limma, made mds generation available only for smaller queries

robinpaul85:
## Description

1) Added support for limma voom
2) Made mds generation available only for smaller queries. On server side `mds_cutoff` is defined . When number of samples * number of genes < mds_cutoff, only then the mds image is generated. This substantially decreases the compute time.
3) Deleted `client/plots/DEanalysis.js`
## Checklist


server/utils/edge.R
```diff
@@ -174,56 +180,141 @@ if (length(input$conf1) == 0) { # No adjustment of confounding factors
...
+       if (dim(read_counts)[1] * dim(read_counts)[2] < as.numeric(input$mds_cutoff)) { # If the dimensions of the read counts matrix is below this threshold, only then the mds image will be generated as its very compute intensive
+           final_output$edgeR_mds_image_name <- mds_image_name
+       }
+       #cat("Time for multiple testing correction: ", as.difftime(multiple_testing_correction_time, units = "secs")[3], " seconds\n")
+
+       # Output results
+       toJSON(final_output)
+} else if (DE_method == "limma") {
+       # Do voom transformation
+       voom_transformation_time <- system.time({
```
karishma-gangwani, line 238: Is voom transformation the only thing that Limma does and the rest is the same as edgeR? Can you add a link to the documentation and paper here in the pr description?

karishma-gangwani:
around line 180-200 some of the logic when DE method is edgeR can be combined together to get rid of repetitive blocks.

karishma-gangwani:
there's also fit_time and test_time repeated in several blocks when the method is edgeR. can also simplify and clean up that.

karishma-gangwani:
if we're not going to be using text files for the counts' data we might as well just remove that block and just keep the hdf5 related code.

karishma-gangwani:
multiple testing correction parts also can be deduplicated.

robinpaul85:
> The confounder logic in the code can be improved and made concise. if a user has not selected a confounder then matrix design is to be modeled only on conditions which is fine but if a user has selected a single confounder: a. check whether continuous or categorical b. check if no confounder 2 present
> 
> then go ahead and design the matrix only considering confounder 1 else if a user has made both selections for two confounders then design the matrix based on both (after checking whether they are numeric or categorical of course).
> 
> In the future, I think we could allow a user to input a comma-separated text input for multiple confounders which the design matrix should be able to handle keeping some sort of upper limit on the number of confounders once the performance is optimized.

I have added more commits, where the code has been streamlined. First the design matrix is generated based upon the number of confounding variables provided. Then "edgeR"/"limma" is used based on user input.

robinpaul85:
> around line 180-200 some of the logic when DE method is edgeR (for running model fitting and MDS plot) can be combined together to get rid of repetitive blocks.

Now all edgeR (and limma) specific code are under one specific block.

robinpaul85:
> if we're not going to be using text files for the counts' data we might as well just remove that block and just keep the hdf5 related code.

This is a separate issue where we need to benchmark the text vs the HDF5 implementation. This will need to happen in another branch since this directly is not related to limma.

robinpaul85:
> multiple testing correction parts also can be deduplicated.

Now there is only 1  multiple testing correction / final_data_generation step

karishma-gangwani:
can also get rid of repetetive logic around line 161 logic. your coefficients column will always be the treatment group.

karishma-gangwani:
> can also get rid of repetitive logic around line 161 of edgeR script. your coefficients column will always be the treatment group.

client/plots/DEanalysis.js
```diff
@@ -131,33 +131,23 @@ class DEanalysis {
 				type: 'radio',
 				chartType: 'DEanalysis',
 				settingsKey: 'method',
-				title: 'Toggle between edgeR and Wilcoxon rank-sum test',
+				title: 'Toggle between edgeR, limma and Wilcoxon rank-sum test',
```
xzhou82, line 134: see if possible to avoid changing DEanalysis.js and do such in DA instead
i hope the old script can be deleted soon

robinpaul85:
> can also get rid of repetitive logic around line 161 logic. your coefficients column will always be the treatment group.

Done

client/plots/DEanalysis.js
robinpaul85: Done

server/utils/edge.R
```diff
@@ -164,46 +152,121 @@ if (length(input$conf1) == 0) { # No adjustment of confounding factors
...
+                  genes_matrix <- str_split_fixed(unlist(top_table$genes), "\t", 2)
+                  geneids <- unlist(genes_matrix[, 1])
+                  genesymbols <- unlist(genes_matrix[, 2])
+                  adjust_p_values <- top_table$adj.P.Val
+             })
+           })
+       })
+       #cat("Time for selecting top genes: ", as.difftime(top_genes_selection_time, units = "secs")[3], " seconds\n")
+} else { # Should not happen
+       stop(paste0("Unknown method:", DE_method))
```
karishma-gangwani, line 266: why is this check needed? I don't think there are any other options to select from apart from Limma, edgeR, and Wilcoxon.

karishma-gangwani (approved):
looks good to me.

server/utils/edge.R
robinpaul85: So that if an unknown method is added, the error message shows that an unsupported DE method was entered (rather than some other random error method which cannot be easily understood). This is just trying to make the code more fool-proof.

client/mass/groups.js
```diff
@@ -223,7 +223,6 @@ class MassGroups {
 
 		//TODO: need to 'diffAnalysis' to `currentCohortChartTypes` in the future
 		if (this.state.currentCohortChartTypes.includes('DEanalysis') && samplelstTW.q.groups.length == 2) {
-			addDEPlotMenuItem(menuDiv, this, this.state, samplelstTW)
```
xzhou82, line 226: am adding minor improvement here

xzhou82 (approved):

client/mass/groups.js
```diff
@@ -301,71 +300,56 @@ class MassGroups {
...
-			const groups = []
-			for (const group of samplelstTW.q.groups) {
-				if (group.values && group.values.length > 0) {
-					groups.push(group)
-				} else {
-					throw 'group does not contain samples for DE analysis'
+function addDiffAnalysisPlotMenuItem(div, self, samplelstTW) {
+	// DA app can be applied to multiple datatypes. show options based on availability for each datatype
+	const allowedTypes = [
+		TermTypes.GENE_EXPRESSION,
```
xzhou82, line 306: ha made mistake here!! will open new pr to fix

xzhou82: fixed ab0e1469777bf92205d545c49ee421d398373c53

Issue #3003, opened 2025-03-21
Disabled terms are clickable from search results in tree


creilly8:
[TermdbTest](http://localhost:3000/?massnative=hg38-test,TermdbTest) -> create group -> click AND and search for previous term. The term is clickable in the search results. 

In this example `Diagnosis Group` is disabled in the tree but a clickable search result: 
<img width="617" alt="Image" src="https://github.com/user-attachments/assets/e0b5d399-36d9-4325-9c4c-e95fda0631db" />

xzhou82:
have to standardize that `disable_terms[]` should be array of term objects, but not array of strings/ids

siosonel:
> have to standardize that `disable_terms[]` should be array of term objects, but not array of strings/ids

this will likely require updating example urls, or alternatively, can rehydrate `disable_terms[]` strings to full `tw` objects

creilly8:
Fixed in https://github.com/stjude/proteinpaint/pull/3020.

PR #3020, opened 2025-03-25
Fix for disabled terms arguments

creilly8:
## Description
Closes issue #3003. 

Changes: 
All instances of `opts.disable_terms` for tree init() pass an array of term objs, not an array of term.ids. Types reflect this change as well. 

Test all changed instances:
1. [Tree integration tests](http://localhost:3000/testrun.html?dir=termdb&name=tree.integration)
2. MultiTermWrapperEditUI. Use the default test in the [DA spec file](http://localhost:3000/testrun.html?dir=plots/diffAnalysis&name=diffAnalysis) -> Click on Confounders btn -> add the first term -> add the second term but search for the first term. Should see the first term appear as disabled in the tree and search result. 
3. Filter. Use [termdb test](http://localhost:3000/?massnative=hg38-test,TermdbTest) -> Add first term to filter -> click add and try to add the same term. Should appear as disabled. 
4. Disabling the group terms in the volcano. Use the [all pharma](http://localhost:3000/?massnative=hg38,ALL-pharmacotyping) -> create two groups. I suggest something with a small number of sample such as small agedx ranges -> launch DA app -> Click on Confounders btn -> add the first term -> add the second term but search for the first term. Should see the first term appear as disabled in the tree and search result. 

## Checklist

### DIFF
--- a/client/termdb/tree.js
+++ b/client/termdb/tree.js
@@ -3,7 +3,7 @@ import { select, selectAll } from 'd3-selection'
 import { getNormalRoot } from '#filter'
 import { isUsableTerm } from '#shared/termdb.usecase.js'
 import { termInfoInit } from './termInfo'
-import { TermTypeGroups } from '#shared/terms.js'
+import { TermTypeGroups, equals } from '#shared/terms.js'
 
 const childterm_indent = '25px'
 export const root_ID = 'root'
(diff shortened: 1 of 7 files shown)


client/termdb/tree.js
```diff
@@ -327,7 +327,7 @@ function setRenderers(self) {
 			return
 		}
 
-		const termIsDisabled = self.opts.disable_terms?.includes(term.id)
+		const termIsDisabled = self.opts?.disable_terms?.some(t => t.id == term.id)
```
xzhou82, line 330: should use `equals()` here

Issue #3021, opened 2025-03-25
DA/volcano: hover over row of p-value table to highlight corresponding item in volcano


xzhou82:
nice to have and not urgent

PR #3025, opened 2025-03-25
Event listeners for p value table

creilly8:
## Description
Closes issue #3021. 

Hovering over the p value table highlights the gene in the volcano plot. Clicking on the row adds a persistent highlight. Test with [the spec file](http://localhost:3000/testrun.html?dir=plots/diffAnalysis&name=diffAnalysis). 

## Checklist

### DIFF
--- a/client/plots/volcano/interactions/VolcanoInteractions.ts
+++ b/client/plots/volcano/interactions/VolcanoInteractions.ts
@@ -90,6 +90,18 @@ export class VolcanoInteractions {
 		}
 	}
 
+	async highlightDataPoint(value: string) {
+		const config = this.app.getState().plots.find((p: VolcanoPlotConfig) => p.id === this.id)
+		const highlightedData = config.highlightedData.includes(value)
+			? config.highlightedData.filter(d => d !== value)
+			: [...config.highlightedData, value]
+		await this.app.dispatch({
+			type: 'plot_edit',
+			id: this.id,
+			config: { highlightedData }
+		})
+	}
+
 	/** When clicking on a data point, launches the box plot in a separate sandbox
 	 * For geneExpression, value == gene symbol */
 	async launchBoxPlot(value: string) {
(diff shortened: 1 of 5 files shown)


karishma-gangwani:
looks nice @creilly8 but if the highlight color is the same as the group color then the dots don't show up in a densely populated area on the the volcano. Maybe make the highlight color black or a stark red?

client/plots/volcano/interactions/VolcanoInteractions.ts
```diff
@@ -90,6 +90,18 @@ export class VolcanoInteractions {
 		}
 	}
 
+	async highlightDataPoint(value: string) {
+		const config = this.app.getState().plots.find((p: VolcanoPlotConfig) => p.id === this.id)
```
xzhou82, line 94: simplify `this.app.getState().plots.find` through some way to directly access plot state?

xzhou82: i see this is only called on clicking a row to commit the hl gene into state. this should be fine

karishma-gangwani:
> > looks nice @creilly8 but if the highlight color is the same as the group color then the dots don't show up in a densely populated area on the the volcano. Maybe make the highlight color black or a stark red?
> 
> The default plot, if no colors are specified, is black and red. If a user selected red for one of the groups, its the same problem. I'm not against changing the color. It feels like a fool's errand however when I can't control the groups color. That's why there's a control to change the highlight color.
> 
> The only thing I can think of is to make the data point color always a shade darker and the highlight a shade lighter to solve the problem.

Sure, maybe a neon color or two shades darker and reduce the opacity of the other dots when hovering over the p-value table rows.

xzhou82 (approved):
very good!
1. need to limit height for the p-value table and scroll it, so volcano and table will both be visible together
2. table.ts unit tests are falling behind

PR #3031, opened 2025-03-26
Run chart

airenzp:
## Description

Added show run chart option to the scatter to support the run chart analysis.  Supporting the run chart plot in the scatter allows to include all the scatter features while using it as a run chart. UI:

Screenshot 2025-03-26 at 11 27 28 AM

## Checklist

### DIFF
--- a/shared/utils/src/descriptive.stats.js
+++ b/shared/utils/src/descriptive.stats.js
@@ -70,7 +70,7 @@ export function getStdDev(data) {
 	return Math.sqrt(variance)
 }
 
-export function getMedian(lst) {
-	lst.sort((a, b) => b - a)
-	return lst[Math.floor(lst.length / 2)]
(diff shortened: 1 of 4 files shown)


shared/utils/src/descriptive.stats.js
```diff
@@ -69,3 +69,8 @@ export function getStdDev(data) {
 	const variance = getVariance(data)
 	return Math.sqrt(variance)
 }
+
+export function getMedian(data) {
```
xzhou82, line 73: not needed?

airenzp: Yeah but I thought it was good to have it in this file. It may be needed later. I can remove it though

airenzp: Removed now

PR #3033, opened 2025-03-27
Refactor geneVariant groupsetting

gavrielm:
## Description

This PR refactors geneVariant groupsetting by allowing groups to be created using stateless filters. These filters are created using new term types for each dt (e.g., `dtsnvindel`, `dtcnv`, `dtfusion`, `dtsv`) that are declared in `shared/utils/src/common.js`. All filter functionalities (IS/NOT, AND/OR, tvs, tvslst) can be used to group samples according to their mutation status. Samples that are not tested for a dt will never pass the filter for that dt (for more details about the design, see these [slides](https://docs.google.com/presentation/d/1fCsz3oxYA9HsI73rVNf907GjCVKkB5wv1LNjeQLLbxs/edit#slide=id.g3459e848b33_0_0)).

Can test with [TP53 in TermdbTest](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22term%22:{%22type%22:%22geneVariant%22,%22gene%22:%22TP53%22}}}]}) -> Open edit menu -> Select the `Assign variants to groups` radio button

Please also test with other datasets that support the geneVariant term type.

## Checklist


karishma-gangwani:
Thanks @gavrielm
just a few questions on this:
1) how are splice variant assignments being handled? I am looking at 'KDM6A' as an example in the mbmeta portal and there is a splice variant in the exon 27 region but is that being considered as truncating by default in our UI? Maybe, for the splice ones we can just move them to the 'rest' since those may or may not be truncating in nature. Also, do you think we should call this group as 'others'? 
2) do you think we could use some padding or margins on the left side of the UI especially for the apply button? or center it a little bit. 

Overall it looks great to me. The functionality works as expected.

gavrielm:
For question 1, splice variant classes are categorized under the SNV/indel dt:

image


For question 2, I can look into adding more padding

server/src/mds3.init.js
```diff
@@ -2763,6 +2762,56 @@ function addDataAvailability(sid, sample2mlst, dtKey, mclass, origin, sampleFilt
...
+		// so sample cannot not pass the filter
+		return [false, false]
+	}
+	const passFilter = filter.in ? passFilterLst : !passFilterLst
+	return [passFilter, allTested]
+}
+
+// function to filter a sample based on its mlst and a filter item
+function filterByItem(filter, mlst) {
+	if (filter.type == 'tvslst') return filterByTvsLst(filter, mlst)
```
xzhou82, line 2789: great solution

xzhou82 (approved):
thank you so much. this is impressive. okay to merge now and continue with additional fixes:
1. document new arg in menu.js
2. no longer use outdated `geneVariantTermGroupsetting`; figure out new way in that when user selects geneVariant termsetting for barchart term2, do not directly jump to fillTW(), but show an intermediary menu with geneVariant  groupsetting UI that asks user to finalize the design first. this should apply to other places e.g. scatter colorTW etc

PR #3089, opened 2025-04-07
Modified hclust.R, edge.R and corr.R to set digits=NA, na = 'string'

robinpaul85:
## Description

Modified hclust.R, edge.R and corr.R to set digits=NA, na = 'string'

na='string' causes any "not a number" to be reported as string. This is similar to whats already being used for `server/utils/cuminc.R`

Very minor change:
    Some lines were incorrectly indented. This is now automatically fixed by my text editor. Maybe should consider common indentation techniques for all text editors like we have for js/ts files.

## Checklist

### DIFF
--- a/server/utils/edge.R
+++ b/server/utils/edge.R
@@ -283,4 +283,4 @@ if (dim(read_counts)[1] * dim(read_counts)[2] < as.numeric(input$mds_cutoff)) { 
 #cat("Time for generating final dataframe: ", as.difftime(final_data_generation_time, unit = "secs")[3], " seconds\n")
 
 # Output results
-toJSON(final_output, digits = NA) # Setting digits = NA makes toJSON() use the max precision. This from ?toJSON() documentation
+toJSON(final_output, digits = NA, na = "string") # Setting digits = NA makes toJSON() use the max precision. na='string' causes any "not a number" to be reported as string. This from ?toJSON() documentation
--- a/server/utils/corr.R
+++ b/server/utils/corr.R
@@ -2,7 +2,7 @@
 
 # Load required packages
 suppressWarnings({
-    library(jsonlite)
+  library(jsonlite)
 })
 
 # Read JSON input from stdin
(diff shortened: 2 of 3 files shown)


server/utils/edge.R
```diff
@@ -283,4 +283,4 @@ if (dim(read_counts)[1] * dim(read_counts)[2] < as.numeric(input$mds_cutoff)) {
 #cat("Time for generating final dataframe: ", as.difftime(final_data_generation_time, unit = "secs")[3], " seconds\n")
 
 # Output results
-toJSON(final_output, digits = NA) # Setting digits = NA makes toJSON() use the max precision. This from ?toJSON() documentation
+toJSON(final_output, digits = NA, na = "string") # Setting digits = NA makes toJSON() use the max precision. This from ?toJSON() documentation
```
creilly8, line 286: Would you please include the purpose for `na="string"` in the comment?

compbiolover (commented):
@robinpaul85 Your comment about having a prettier for R code: We should look into using this: https://styler.r-lib.org/. It follows the tidyverse style guide but we can also set our own style guide if we wanted.

server/utils/edge.R
robinpaul85: Done

robinpaul85:
> @robinpaul85 We should look into using Styler (https://styler.r-lib.org/). It follows the tidyverse style guide but we can also set our own style guide if we wanted.

I think we need to address this issue as a group. Somewhere in the project folder this needs to be set.

creilly8 (approved):

compbiolover:
> > @robinpaul85 We should look into using Styler (https://styler.r-lib.org/). It follows the tidyverse style guide but we can also set our own style guide if we wanted.
> 
> I think we need to address this issue as a group. Somewhere in the project folder this needs to be set.

No problem. I agree. I think group discussion on it would be good/needed. It would be nice if we could have it run at commit time if it detects an R file

PR #3107, opened 2025-04-09
add copy button for session url

gavrielm:
## Description

Copy button is now added to session url. Made some changes to `client/dom/copyButton.js` to support copying attribute value (e.g., `href`).

Go to session button -> share -> copy button will appear next to url. The copy button will now allow user to easily share session links on vizcom.

## Checklist


client/mass/sessionBtn.js
```diff
@@ -370,7 +371,10 @@ class MassSessionBtn {
 				.style('margin-bottom', '12px')
 				.html(`Click the link to recover this session. Bookmark or share this link.`)
 
-			const a = linkDiv.append('a').attr('href', url).attr('target', '_blank').html(res.id)
+			const a = linkDiv.append('a').attr('id', 'sessionLink').attr('href', url).attr('target', '_blank').html(res.id)
```
creilly8, line 374: This isn't necessary but may not show the link at all for vizcom? May be show just the copy button to avoid the user  error altogether. If that's worthwhile, use ```const hostURL = sessionStorage.getItem('hostURL')``` to determine if it's a vizcom or non pp site.

gavrielm: There is no user error on vizcom. When user clicks link, the session will open in new tab (but url in browser will not change). Its just that user has no way to share/copy the link. So copy button will allow for that.

creilly8: That's what I meant. If we know that user behavior will cause an issue, maybe it's best to remove the element entirely to avoid that specific issue. Again, not necessary but it would improve the user experience.

siosonel:
The copy button is too far away from the link, and the icon is not clear. Maybe there should be two buttons, `Copy link` and `Open link`.

Screenshot 2025-04-09 at 11 35 42 AM

client/mass/sessionBtn.js
siosonel: I think there should be two buttons, `Copy link` and `Open link`, then it's clear what will happen.

gavrielm: Ok sounds good

siosonel (commented):
I get this error in the console when I mouseover the check character after clicking on the copy button. Maybe fix with `if (tip) setTimeout(()=>tip.hide()...`

Screenshot 2025-04-09 at 1 40 41 PM

Screenshot 2025-04-09 at 1 39 36 PM

siosonel (approved):

siosonel:
all unit and integration tests passed locally

Issue #3119, opened 2025-04-09
Groups with 'other' not formatted for all plots and client side functions


karishma-gangwani:
when only one variable category is used to create a custom variable group vs. others, edgeR is run, having repeated column names in the count's matrix breaks. 

the issue is around line 55.

karishma-gangwani:
issue is at creating groups in custom variable grouping. When only one group is created, the samples do not become a subset of only group but are present in both 'group 1' as well as 'others'

see below, sample hypo102 is present in both groups. so the overlap breaks edgeR

![Image](https://github.com/user-attachments/assets/3d494504-9836-4e3f-9616-129777854f6b)

![Image](https://github.com/user-attachments/assets/d0de3116-2204-43bd-b96a-32116446d146)

PR #3175, opened 2025-04-17
Stop gap soln for 'others' grps not working for 'as is' samplelst configs

creilly8:
## Description
Addresses #3119 . 

Several plot configs use `samplelst.groups` returned from groups 'as is', without modifying the groups values array. This is problematic when a custom variable is generated from an 'others' group. This stop gap solution addresses the immediate need for plots to work. 

***Note***: 
1. This solution is to address an immediate need, not serve as a long term fix. If additional problems, bugs, etc. are discovered during testing, we need to consider the issue(s) in a long term fix instead of this PR. 
2. This solution **does not** work in TermdbTest. TermdbTest includes repeat samples so this fails. The solution works though in ihg, mbmeta, and ash. 
3. Leave commented out code. It will be helpful to keep the history immediately available when attempting a long term fix. 

## Checklist

### DIFF
--- a/client/mass/groups.js
+++ b/client/mass/groups.js
@@ -119,17 +119,19 @@ class MassGroups {
 
 		//when there is only one group and need to create a others group
 		if (groups.length == 1) {
-			// find the sample count in current cohort
-			let countSampleCount
-			if (this.state.termfilter.filter.lst.length)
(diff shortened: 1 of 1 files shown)


xzhou82:
alternative https://github.com/stjude/proteinpaint/pull/3194

Issue #3114, opened 2025-04-09
periodically clear python pickle file from cache folder


xzhou82:
following the fix of https://github.com/stjude/proteinpaint/pull/3109, gsea pickle file names are truly random now and will start to accumulate on gdc pp container (after next release to gdc)

may extend the bam slice auto clear logic

PR #3189, opened 2025-04-18
Delete pkl files from cache dir every 30 days

creilly8:
## Description
Closes issue #3114.

Every 30 days, on init(), pickle files will be removed from the cache directory. I choose 30 days as an arbitrary cutoff. We should also consider creating a specific dir for cached DA/DE/GSEA files in the future for a targeted approach. 

Test: 
1. Create pickle files by launching the gsea from [the DA spec](http://localhost:3000/testrun.html?dir=plots/diffAnalysis&name=diffAnalysis) -> select a gene set to create the image. -> Should see a .pkl file in `/Your/filepath/data/cache`. 
2. In `server/src/initGenomesDs.js`, change line 621 to `const numOfSaveDays = -1`. 
3. Restart your instance. Should see in the server log `'Pickle file deleted: [file name]'` and no .pkl files in `/Your/filepath/data/cache`. 

## Checklist

### DIFF
--- a/server/src/initGenomesDs.js
+++ b/server/src/initGenomesDs.js
@@ -58,6 +58,7 @@ export async function initGenomesDs(serverconfig) {
 	//DELETE THIS after process for deleting mass session files moved into production
 	serverconfig.cachedir_massSessionTrash = await mayCreateSubdirInCache('massSessionTrash')
 	deleteSessionFiles()
+	deletePklFiles()
 
 	serverconfig.cache_snpgt = {
 		dir: await mayCreateSubdirInCache('snpgt'),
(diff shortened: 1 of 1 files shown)


creilly8:
Also we might consider using `massSessionDuration` as a global cutoff since it can be set in the serverconfig.

xzhou82:
sorry didn't fully specify requirement. just deleting on launch is not surfficient. esp on gdc where a container may start with clean slate with empty cache. consider lots of hits to a container that can accumulate the file to a point of full disk.

this should be done on generating a pickle file, do actions to delete the old enough ones, or even newer ones if there are too many files. Edgar has implementation in bam.js that may be generalized

PR #3191, opened 2025-04-18
Implementation of new geneVariant tvs UI in global filter

gavrielm:
## Description

Edgar and I worked together on this PR to implement the new geneVariant tvs UI in the global filter. Each dt filter term (e.g., SNV/indel, CNV, Fusion) can be added one at a time to the global filter to filter samples based on their mutation status.

Can test by opening [TP53 barchart](http://localhost:3000/?mass={%22dslabel%22:%22MB_meta_analysis2%22,%22nav%22:{%22activeTab%22:1},%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22term%22:{%22type%22:%22geneVariant%22,%22gene%22:%22TP53%22}}}]}) -> add a global filter for TP53 -> somatic SNV/indel -> missense. The number of samples with a somatic missense mutation in the barchart should match the total number in the filtered subcohort.

Please test with different AND/OR combinations, different genes/dictionary terms, and different plots

## Checklist


karishma-gangwani:
Thanks Gav. So in that link, after I create my selections, it loads the bar chart according to the group selections, but when the variant filters are removed completely and applied, then no bar chart is rendered. Do you think that if no variant filter is assigned, when a user submits empty filters, it should go back to the default bar chart when no variant grouping is done?

client/filter/tvs.geneVariant.js
```diff
@@ -21,235 +24,267 @@ export const handler = {
...
+		tree: {
+			click_term2select_tvs(tvs) {
+				tvs.term.geneVariantTerm = structuredClone(term) // track the parent geneVariant term
+				self.opts.callback(tvs)
+			}
+		}
+	})
 
-	div.append('div').style('font-size', '1.2rem').text(self.tvs.term.name)
+	return
```
siosonel, line 57: @gavrielm  I think the commented out code should be deleted. I had kept this earlier in case it may be needed for something else, but it doesn't look like it'll be used again.

xzhou82 (approved):
thanks it works.
i added one integration test as an attempt to get some coverage on new code, and we can revise it later

PR #3193, opened 2025-04-18
fix: Added tabix support back for all non-HDF5 datasets

compbiolover:
## Description
This fix adds support for tabix (.gz) support in the error handling. The logic is that we first try to validate the file as HDF5. If it fails then it falls back to the tabix handling. The error message from the attempted HDF5 validation will be shown in your server log. It will also report the successful tabix validation

## Checklist

### DIFF
--- a/release.txt
+++ b/release.txt
@@ -1,1 +1,2 @@
-
+Fixes:
+- Added tabix support back for all non-HDF5 datasets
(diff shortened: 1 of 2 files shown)


creilly8 (dismissed):
All of my datasets loaded when I restarted my instance. Thanks for the fix.

compbiolover:
Thrilled to hear it! No problem

xzhou82 (approved):

PR #3170, opened 2025-04-17
Updated `DEanalysis.rs`, `edge.R`,  `topGeneByExpressionVariance.rs`, and Multiple Files related to the Volcano plot and its test data to use `gene_ids` and `gene_names` instead of `gene_symbols`

compbiolover:
## Description
Updated `DEanalysis.rs`, and `edge.R` to use gene_ids and gene_names. Removed gene_symbols support and used gene_names instead. Used `formateElapsedTime` for reporting the time taking for DE analysis in `termdb.DE.ts`

reminder to recompile rust after checking branch

## To Test
Go to [here](http://localhost:3000/?mass={%22dslabel%22:%22ALL-pharmacotyping%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:2},%22groups%22:[{%22name%22:%22Resistant%22,%22color%22:%22%23f09930%22,%22filter%22:{%22tag%22:%22filterUiRoot%22,%22in%22:true,%22type%22:%22tvslst%22,%22join%22:%22and%22,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Molecular%20subtype%22},%22values%22:[{%22key%22:%22ETV6-RUNX1%22},{%22key%22:%22ETV6-RUNX1-like%22}]}},{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Asparaginase_normalizedLC50%22},%22ranges%22:[{%22start%22:0.1,%22startinclusive%22:true,%22stopunbounded%22:true}]}}]}},{%22name%22:%22Sensitive%22,%22color%22:%22%233453ed%22,%22filter%22:{%22tag%22:%22filterUiRoot%22,%22in%22:true,%22type%22:%22tvslst%22,%22join%22:%22and%22,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Molecular%20subtype%22},%22values%22:[{%22key%22:%22ETV6-RUNX1%22},{%22key%22:%22ETV6-RUNX1-like%22}]}},{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Asparaginase_normalizedLC50%22},%22ranges%22:[{%22stop%22:0.1,%22startunbounded%22:true}]}}]}}]}) and test that DA analysis works as expected for both edgeR, limma, and wilcoxon

## Checklist

[...]

PR #3208, opened 2025-04-23
Adding DE to Termdb Test

robinpaul85:
## Description
Implementation to add DE app to termdb Test

1) The same set of samples and genes have been retained as in the fpkm HDF5 file so that later exporting of DE results to hierarchial clustering can be tested.
2) Added some checks so that edgeR script and wilcoxon script  exit gracefully when after filtering no genes remain in the counts matrix.
3) Still deciding appropriate number of genes for DE.

## Checklist


rust/src/DEanalysis.rs
```diff
@@ -664,6 +664,14 @@ fn main() {
                                 //println!("filtering time:{:?}", filtering_time.elapsed());
                                 //println!("filtered_matrix_rows:{:?}", filtered_matrix.nrows());
                                 //println!("filtered_matrix_cols:{:?}", filtered_matrix.ncols());
+                                if filtered_matrix.nrows() == 0 {
+                                    // Its possible after filtering there might not be any genes left in the matrix, in such a case the rust code must exit gracefully with an error.
+                                    panic!("Number of genes after filtering = 0, cannot proceed any further")
```
compbiolover, line 669: I like this. My only concern is I don't think panic would return output to a user so it may fail silently. We might want to instead encapsulate the messages about no genes or samples after filtering into json objects that get returned and get displayed to the user in the console

xzhou82:
should merge https://github.com/stjude/proteinpaint/pull/3170 first, then rebuild h5 in this branch

rust/src/DEanalysis.rs
compbiolover: @robinpaul85 Sorry this slipped by me. That sounds good

Issue #3210, opened 2025-04-23
DE enhancements


karishma-gangwani:
would be great if we can implement these: 

1) adding group labels on DE volcano on both sides similar to this example:

![Image](https://github.com/user-attachments/assets/50052b63-52ad-4b15-b9d0-37af71478cf4)

2) changing p-value cut off input from linear to log scale

PR #3226, opened 2025-04-25
DA volcano shows group name on top

creilly8:
## Description
Closes issue #3210. Works with very long term names and displays the user selected color, on the site and in the download.

## Checklist

### DIFF
--- a/client/plots/volcano/viewModel/VolcanoViewModel.ts
+++ b/client/plots/volcano/viewModel/VolcanoViewModel.ts
@@ -144,13 +144,13 @@ export class VolcanoViewModel {
 		return [
 			{
 				color: controlColor || this.settings.defaultSignColor,
-				label: this.config.samplelst.groups[0].name,
+				label: `${this.config.samplelst.groups[0].name} (control group)`,
 				x: plotDim.top.x,
 				y: plotDim.top.y
 			},
 			{
 				color: caseColor || this.settings.defaultSignColor,
-				label: this.config.samplelst.groups[1].name,
+				label: `${this.config.samplelst.groups[1].name} (case group)`,
 				x: plotDim.top.x,
 				y: plotDim.top.y + 18
 			}
--- a/client/plots/volcano/viewModel/VolcanoViewModel.ts
+++ b/client/plots/volcano/viewModel/VolcanoViewModel.ts
@@ -29,6 +29,8 @@ export class VolcanoViewModel {
 	maxLogPValue = 0
 	//Used in place of 0 p values that cannot be log transformed
 	minNonZeroPValue = 10e-10
+	//The x coord flush with the left side of the plot
+	plotX: number
 	readonly offset = 10
 	readonly bottomPad = 60
 	readonly horizPad = 70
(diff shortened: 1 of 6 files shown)


congyu-lu (commented):
image
I think for the legends of the two groups, we need to show that which color represents genes are upregulated/downregulated in which group, instead of showing which color represents which group.

creilly8:
> image I think for the legends of the two groups, we need to show that which color represents genes are upregulated/downregulated in which group, instead of showing which color represents which group.

@congyu-lu. I don't follow. The unregulated genes in the case group have a pos fc (case group color) and the downregulated genes in the case group have a neg fc (control group color). So what exactly would change?

congyu-lu:
> > image I think for the legends of the two groups, we need to show that which color represents genes are upregulated/downregulated in which group, instead of showing which color represents which group.
> 
> @congyu-lu. I don't follow. The unregulated genes in the case group have a pos fc (case group color) and the downregulated genes in the case group have a neg fc (control group color). So what exactly would change?

In the current legend, green/orange represents group A/B, but the blue/black/orange color in the volcano plot represents down-regulated/not significant/up-regulated.

creilly8:
> > > image I think for the legends of the two groups, we need to show that which color represents genes are upregulated/downregulated in which group, instead of showing which color represents which group.
> > 
> > 
> > @congyu-lu. I don't follow. The unregulated genes in the case group have a pos fc (case group color) and the downregulated genes in the case group have a neg fc (control group color). So what exactly would change?
> 
> In the current legend, green/orange represents group A/B, but the blue/black/orange color in the volcano plot represents down-regulated/not significant/up-regulated.
> 
> I think we could either (1). do it as this figure (showing which color represents genes are upregulated/downregulated in which group), or (2). clearly indicate which group is "control group" and show the color legend as down-regulated/not significant/up-regulated in "treatment group" compared with "control group".
> 
> image

The assumption is that a gene is required in one or both group terms to run the app. That's not the case. The first option isn't possible. I can label the groups though.

client/plots/volcano/viewModel/VolcanoViewModel.ts
```diff
@@ -98,50 +99,72 @@ export class VolcanoViewModel {
...
+				y2: this.settings.height + this.offset * 4
 			}
 		}
 	}
 
+	setTermInfo(plotDim, caseColor, controlColor) {
+		return [
+			{
+				color: controlColor || this.settings.defaultSignColor,
+				label: `${this.config.samplelst.groups[0].name} (control group)`,
```
xzhou82, line 147: should implement the given idea, ok not to draw the pair of triangle
just print group name on each half, so case label is on right, control label is on left. label too long? truncate
this way, no need for labels `(control grp)`

xzhou82:
setting -log10 cutoff to 4 doesn't work (reverts to 1)

creilly8:
> setting -log10 cutoff to 4 doesn't work (reverts to 1)

Thanks for catching. This addressed in https://github.com/stjude/proteinpaint/pull/3226/commits/64e33a04dadd01da6713a899eb663486a0c5792b.

xzhou82 (approved):
thanks

PR #3231, opened 2025-04-28
Converting Rust workspace to ESM and README Update

compbiolover:
## Description
Converting Rust workspace to ESM and README Update. `npx tsc` and `npx test` from `proteinpaint/server` are clear. All unit and integration tests pass. Server loads as expected. Rust function tested: top variably expressed genes, DA, summary plots like violin plots, box plots, etc (tests rust HDF5 loading code). All tests performed on All-pharma dataset.

## Checklist


rust/test/fisherTest.js
```diff
@@ -7,11 +7,12 @@ Run test script as follows (from 'proteinpaint/'):
 
 *********************************************/
 
-const tape = require('tape')
-const run_rust = require('@sjcrh/proteinpaint-rust').run_rust
+// Import necessary modules
+import { tape } from 'tape'
```
siosonel, line 11: I think this should be `import tape from 'tape'` like in other spec files, not using curly braces.

robinpaul85: @siosonel Can these test be run on CI on a nightly basis?

siosonel: It ran as a PR check after you pushed your last commit. In general, unit tests for a workspace will run if there are detected code changes in that workspace, as part of PR push and also after merging to master. It may also run as part of combined coverage CI nightly.

siosonel: You can also run it locally as `npm run test:unit` from the `rust` dir. Even if it passes, we should try to be consistent in how `tape` is imported. I think `import { tape } from 'tape'` means a separate instance of `tape` will run for the spec file that uses that syntax, so those different spec files will each have a pass/fail result and failure stderr emit that could stop the single-thread js runtime. In contrast, using `import tape from 'tape'` will use the same instance for different spec files, so the results are aggregated and only a single pass/fail tally is emittted.

compbiolover: @siosonel @robinpaul85 If we want those tests included I can try renaming them and pushing them

robinpaul85:
> ## Description
> Converting Rust workspace to ESM and README Update. `npx tsc` and `npx test` from `proteinpaint/server` are clear. All unit and integration tests pass. Server loads as expected. Rust function tested: top variably expressed genes, DA, summary plots like violin plots, box plots, etc (tests rust HDF5 loading code). All tests performed on All-pharma dataset.
> 
> ## Checklist
> [Check each task](https://github.com/stjude/proteinpaint/wiki/Pull-Request-Checklist) that has been performed or verified to be not applicable.
> 
> * [x]  Tests: added and/or passed unit and integration tests, or N/A
> * [x]  Todos: commented or documented, or N/A
> * [x]  Notable Changes: updated release.txt, prefixed a commit message with "fix:" or "feat:", added to an internal tracking document, or N/A

 for CI may be we should only test on `TermdbTest` ?

rust/test/fisherTest.js
robinpaul85: Yes, try renaming the files such that it fits `tape **/test/*.unit.spec.js` and then see if it runs on CI.

compbiolover: @siosonel Thanks for clarifying. I went back and removed the curly braces for single module imports. I didn't realize that workspaces like run_rust were treated differently so I am now putting the curly braces back for it. Now tape (and other single module imports) are just `import ... from ...` instead of `import {...} from ...`

compbiolover: @robinpaul85 I'll do that and try it out

siosonel: if those tests do not use TermdbTest, they will fail on CI, but may pass locally if you have the tested dataset in your local serverconfig.

rust/package.json
```diff
@@ -1,6 +1,12 @@
 {
 	"version": "2.120.1",
 	"name": "@sjcrh/proteinpaint-rust",
+	"type": "module",
+	"exports": {
+		".": {
```
siosonel, line 6: Please simplify to `".": "./index.js"`. we are not offering "require" alternative so no need to distinguish the "import" entry point and create a nested object.

compbiolover: Will simplify

rust/test/fisherTest.js
compbiolover: @siosonel They have test data that gets made inside their own test file (e.g. a dummy matrix is just made within the spec file itself). They pass on my local. Is that okay or will that also fail on github CI?

rust/package.json
siosonel: the syntax is 
```json
"exports": {
  ".": "./ndex.js"
}
```

rust/test/fisherTest.js
siosonel: it's okay if they make their own test file, you can try pushing and see if these renamed tests pass

rust/package.json
siosonel: it's okay, I fixed it myself just by using the usual `"main"` field, since it seems to have worked even not having the `"exports"` field. I can wait for the renamed wilcoxon and fisherTest to be pushed and run on CI.

compbiolover: @siosonel Awesome, thanks. I just pushed them and they worked. I still need to get the `readHDF5.rs` tests fully functioning (still a few failed tests because I was originally going to use some of my own simulated data and Xin said I could use TermdbTest data instead). We can go ahead and wrap up this pr here and I'll open a different pr for the `readHDF5.rs` tests. Don't want to get too many things in a single pr

siosonel (approved):

PR #3236, opened 2025-04-30
new block callback onAddRemoveTk and replace onsetheight usage

### LINKED ISSUE #3217: in container, GB main() not called on panning
[try running this link in container](http://localhost:3000/?mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22genomeBrowser%22,%22geneSearchResult%22:{%22chr%22:%22chr17%22,%22start%22:7666657,%22stop%22:7688274},%22snvindel%22:{%22details%22:{%22groups%22:[{%22type%22:%22info%22,%22infoKey%22:%22AF%22}]}}}]}); on panning, no variants show up  over new region

xzhou82:
## Description

closes #3217 
tested with `sjlife` and `peddep` links from block section of url.html. both works and can be saved to state and recover

sorry for delay in committing the fix. i think my changes are simpler and i reverted your previous changes hope you don't mind

this should be tested in container first
[gb manual test](http://localhost:3000/testrun.html?dir=plots&name=genomeBrowser) is fixed and all passed

## Checklist

### DIFF
--- a/client/plots/genomeBrowser.js
+++ b/client/plots/genomeBrowser.js
@@ -127,7 +127,6 @@ class genomeBrowser {
 		try {
 			const tklst = await this.generateTracks()
 			await this.launchBlockWithTracks(tklst)
-			this.hasActiveCoordChange = false
 		} catch (e) {
 			sayerror(this.dom.errDiv, e.message || e)
 			if (e.stack) console.log(e.stack)
@@ -158,7 +157,7 @@ class genomeBrowser {
 					dslabel: this.app.opts.state.vocab.dslabel,
 					onClose: () => {
 						// on closing subtk, the filterObj corresponding to the subtk will be "removed" from subMds3TkFilters[], by regenerating the array
-						if (!this.hasActiveCoordChange) this.maySaveTrackUpdatesToState('onClose')
+						this.maySaveTrackUpdatesToState()
 					},
 					// for showing disco etc as ad-hoc sandbox, persistently in the mass plotDiv, rather than a menu
 					newChartHolder: this.opts.plotDiv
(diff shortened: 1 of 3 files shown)


gavrielm:
Tested genome browser from sjlife, peddep, and ash in the container and all work. For the test script `client/plots/test/genomeBrowser.spec.js`, it should be renamed to `genomeBrowser.integration.spec.js` if you want it on CI.

siosonel:
> Tested genome browser from sjlife, peddep, and ash in the container and all work. For the test script `client/plots/test/genomeBrowser.spec.js`, it should be renamed to `genomeBrowser.integration.spec.js` if you want it on CI.

Since `genomeBrowser.integration.spec.js` uses `SJLife`, it's okay to keep the name as-is. Only end-to-end tests can handle private dataset; `unit.spec` and `integration.spec` files are expected to use TermdbTest.

@xzhou82 can this PR be marked as ready?

gavrielm:
Ok. I don't think `genomeBrowser.spec.js` is currently run as part of an end-to-end test, which means it is not run automatically. Should `TermdbTest` be used instead of `SJLife` so that the script can be used as an integration test?

siosonel:
> Ok. I don't think `genomeBrowser.spec.js` is currently run as part of an end-to-end test, which means it is not run automatically. Should `TermdbTest` be used instead of `SJLife` so that the script can be used as an integration test?

Yes, but the challenge may be that the minimal mutation/snp data in TermdbTest is too sparse to simulate genome brower panning where new variants actually exist and show up, it may be hard to find a position/zoom level that can simulate what's going on with sjlife. But this is just my guess.

xzhou82:
we don't have a test to cover gb panning yet. the tests around gb is quite lacking. thanks i will merge

PR #3288, opened 2025-05-08
Improved Rust linting via clippy

compbiolover:
## Description
Using a baseline JSON file and a python script to allow Rust linting to only consider new errors/warnings that are not present in the file. Switched to using `cargo clippy` which now allows access to Rust crates and dependencies. To have new files checked in our workspace they must be added to `Cargo.toml` with their own `bin` entry. As we fix the various warnings in the existing Rust code we can then create a new baseline JSON file in the future. Currently I allow files to be pushed even if we see new warnings as we have many warnings and a couple of files have errors. If compilation fails then we may not see all warnings in a file. Once we get the various warnings resolved and fixed I will disable the true flag and new warnings or errors caught by the pre-commit hook will be blocking.

## To Test
Make a slight modification to an existing Rust file in `rust/src` (e.g. in a fn call add `unused_var = "unused variable";` and create a new new Rust file in `rust/src` with `touch rust_lint_test.rs`. Open this file and add `fn main(){println!("Hello world");}`. You should see warnings that aren't able to be caught by the baseline. You should still be able to push your commit even with these new warnings. These unused variables should still be caught by CI and blocked as seen in my test commit [ee2d723](https://github.com/stjude/proteinpaint/pull/3288/commits/ee2d72385d45168705008ba3462d8b8eb93b66b4)

## Checklist


xzhou82:
is the baseline "json" file really needed? it's 2Mb

robinpaul85:
I often commit rust code with warnings and in rare cases compilation errors. It helps me work on multiple devices and avoids loosing a day's work. A better option would be blocking the PR automatically from getting merged until all compilation and warning errors are resolved?

xzhou82:
new strategy:
1. disable rust linting from precommit hook for now
2. share with team a command to manually run rust linting, and grep output to find out all warnings from a particular rust script
3. a team member dedicates some time to address all warnings from a script; test it to make sure still works and commit
4. cover all rust scripts so no warning/error from linting in master code
5. enable linting in precommit
this way no need to track the baseline file and add more complexity to have code that will manage that file.

xzhou82 (approved):
thanks

Issue #3283, opened 2025-05-08
use value interpolation in all constructed sql queries


siosonel:
Do not use `term_id='${tw.term.id}`, should be `term_id=?` and `values.push(tw.term.id)` in `termdb.sql.multivalue.js`,  `termdb.sql.date.js`.

Do similar fix for `termdb.sql.samplelst.js`, where `samplesString` is replaced with `?, ?, ?, ...` instead of directly using strings from the request query parameters.

Validate `anno_${term_type}` to whitelist that `['integer', 'float', ...].includes(term_type)` in `termdb.sql.numeric.js` and in `termdb.sql.js get_summary_numericcategories()`.

Validate that `v.key` only has alphanumeri characters (no whitespace, etc) in `termdb.filter.js getMultiValue()`.

There may be other sql statements that needs fixing with regards to value interpolation or validation.

PR #3291, opened 2025-05-08
Rewrote date and samplelst queries to remove risk of sql injection

airenzp:
## Description

Passed parameters for the query in the values. Solves partially issue #3283 as other queries need to be reviewed.

## Checklist

### DIFF
--- a/server/src/termdb.sql.samplelst.js
+++ b/server/src/termdb.sql.samplelst.js
@@ -8,15 +8,18 @@ export const sampleLstSql = {
 			if (!('in' in group)) group.in = true
 			samples = group.values.map(value => value.sampleId || value.sample)
 			const type = samples[0] ? ds.sampleId2Type.get(samples[0]) : ''
-			samplesString = samples.join(',') //later on need to unify the list handling in samplelst
-
+			samplesString = Array(samples.length).fill('?').join(',') //later on need to unify the list handling in samplelst
 			sql += `SELECT id as sample, ? as key, ? as value
 				FROM sampleidmap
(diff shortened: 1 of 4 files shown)


server/src/termdb.sql.samplelst.js
```diff
@@ -8,15 +8,18 @@ export const sampleLstSql = {
 			if (!('in' in group)) group.in = true
 			samples = group.values.map(value => value.sampleId || value.sample)
 			const type = samples[0] ? ds.sampleId2Type.get(samples[0]) : ''
-			samplesString = samples.join(',') //later on need to unify the list handling in samplelst
-
+			samplesString = Array(samples.length).fill('?').join(',') //later on need to unify the list handling in samplelst
 			sql += `SELECT id as sample, ? as key, ? as value
 				FROM sampleidmap
 				WHERE sample ${group.in ? '' : 'NOT'} IN (${samplesString})  `
```
xzhou82, line 14: do you think `(${samplesString})` should be interpolated?

airenzp: It is a string I am building without using any parameter, it will become something like "?,?,?" and the interrogation mark will be replaced by the values later, so there is no risk of injection. If you mean I could pass less parameters by creating a value that has the list already it sounds good to me, I could pass less values and question marks doing that. @siosonel what do you think?

siosonel: Each value should be interpolated separately, better-sqlite has to detect whether the substituted value is a string or number. Multiple values cannot be safely concatenated as a comma-separated string with only one `?` to substitute.

xzhou82 (approved):
thanks

PR #3293, opened 2025-05-09
validate term.type as used for anno_* table name

siosonel:
## Description

Validate term.type as used for anno_* table name as additional protection sql injection beyond if-else/case branching against term.type's in request payload.

Tested with
- http://localhost:3000/testrun.html?name=*.unit
- http://localhost:3000/testrun.html?name=*.integration

## Checklist

### DIFF
--- a/server/src/termdb.filter.js
+++ b/server/src/termdb.filter.js
@@ -4,6 +4,9 @@ import { dtTerms } from '#shared/common.js'
 import { getSnpData } from './termdb.matrix.js'
 import { filterByItem } from './mds3.init.js'
 
+// includes term.type=date, unlike in termdb.sql.numeric.js
+const annoNumericTypes = new Set(['integer', 'float', 'date'])
+
 /*
 nested filter documented at:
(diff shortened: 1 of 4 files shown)


gavrielm (approved):
Tested with different plots and filters and all work well

server/src/termdb.filter.js
```diff
@@ -4,6 +4,9 @@ import { dtTerms } from '#shared/common.js'
 import { getSnpData } from './termdb.matrix.js'
 import { filterByItem } from './mds3.init.js'
 
+// includes term.type=date, unlike in termdb.sql.numeric.js
+const annoNumericTypes = new Set(['integer', 'float', 'date'])
```
airenzp, line 8: Maybe we can import the numeric types set defined in shared/utils/src/terms.js and check against this list to avoid duplicates. Every time we add a new term we edit the terms.js file so it is easier to remember adding it to the numeric types if it is numeric.

airenzp (approved):
Thanks

PR #3299, opened 2025-05-09
Periodically delete gsea files on an interval

creilly8:
# Description
Addresses the second and partially the third items in issue #3247. This branch [passed on CI](https://github.com/stjude/proteinpaint/actions/runs/14937689058) before making this PR. 

Changes: 
- Files maybe deleted out of a cache subdirectory based on age and size 
- Allows the serverconfig to set `features.cacheMonitor`
- Deleted outdated code to delete only the gsea files. 
- New unit tests for cache init and deleting test files. 

Automated test: 
- `npx tsx server/src/test/CacheManager.unit.spec.js` from the `proteinpaint` dir
- can temporarily change to `const interval = 1000` (1 second) to be make the file move/deletion more observable in two Mac Finder windows that show `proteinpaint/.cache/test0` and `trash` subdirs.

Manual test:
- should see these logs initially when the server restarts
```bash
checking for cached gsea files to delete ...
No gsea cached files to delete
checking for cached massSession files to delete ...
No massSession cached files to delete
checking for cached massSessionTrash files to delete ...
deleted 0 of [...] massSessionTrash cached files (0 bytes deleted, [...] remaining)
```
- After 1 minute, should see `checking for cached gsea files to delete ...`, but there should be no cache checks for `massSession*` subdirs since the `skipMs` is 30+ days on those. Subsequent messages indicate whether there were files to delete and if files were deleted. 
- For further testing: 
    - Speed up or slow down the interval by adding the following to your `serverconfig.features` and changing the `intervalOverride`:
    ```
		"cacheMonitor": {
			"interval": 500, 
                        "subdirs": {
		           "gsea": {
				"maxAge": 10000,
				"maxSize": 5e9
                            }
			}
		}
    ```
    - Create a few pickle files by using the [gsea spec file](http://localhost:3000/testrun.html?dir=plots&name=gsea), selecting a gene set group, then a gene set. Change the gene set group a few times and more pickle files will be created. Then check the server console for messages about whether or not the file was deleted. 
    - Verify the files deleted by checking `your/file/path/cache/gsea`. I suggest checking in between intervals. 

## Checklist


server/src/test/cache.unit.spec.js
```diff
@@ -0,0 +1,86 @@
...
+
+const cache = new Cache(serverconfig.cachedir)
+
+/**************
+ helper functions
+***************/
+
+function makeTestFiles() {
+	for (const subdir of cache.defaultCacheDirs) {
+		const dirPath = path.join(serverconfig.cachedir, subdir.dir)
```
xzhou82, line 20: this test breaks on local because somehow `serverconfig.cachedir` value is "cache" rather than actual full path in my own serverconfig.json. and this breaks fs.promises.mkdir() at line 60 of Cache.ts

creilly8: @xzhou82, please try the latest changes. I have no issues locally when my `server/serverconfig.json` is set to the absolute file path we discussed.

xzhou82: still fails on my local

siosonel:
@creilly8 and @xzhou Instead of making suggested changes in PR comments, I found it easier to code `CacheManager.ts` and other changes to clarify and avoid recomputing cache options and default values. Note that I have not rewritten the `cache.unit.spec.ts`, in case this is not the direction we want to take.

To test manually, `touch /path/to/cachedir/massSession/test-000` and add this `serverconfig.features`
```json
          "cacheMonitor": {
               "interval": 5000,
              "monitorByDir": {
                "massSession": {
			"maxAge": 1000,
			"skipMs": 1000
		}
             }
          }
```

On server startup, you see logs like below. Every time a massSession file is added using `touch ..`, it will be deleted after 1 second.
```bash
checking for cached massSession files to delete ...
deleted 1 of 1 massSession cached files (0 bytes deleted, 0 remaining)
...
checking for cached massSession files to delete ...
No massSession cached files to delete

```

creilly8:
Edgar's version worked for me. I don't have any objections.

siosonel (dismissed):

siosonel (dismissed):

siosonel (approved):

siosonel (approved):

PR #3302, opened 2025-05-12
fix: use a saved jwt token from a previous successful password login,…

siosonel:
… if available for a protected dataset route

# Description

The bug fix 
- codes a more comprehensive handling of session jwt that is generated from password login. 
- ensures that a session jwt is reused across different server instance, by distinguish between login and session jwt's
- forces re-login if an expired session jwt is detected for `auth.type == 'basic'`

TODO: In a another PR/branch, separate client auth code from `dofetch.js` and write client-side auth tests.

Tested with:
- MB meta analysis 2: may also test expired token by forcing `if (time > payload.exp)` to true in `getJwtPayload()` in `server/src/auth.js`. This should cause the login UI to be always shown on page load, but there will be an expected data request/server middleware error after successful login in this test since the jwt will always be detected as expired
- SJLife termdb works, and also burden route:
- PNET

## Checklist


karishma-gangwani (approved):
Thanks a lot @siosonel. It works on my local instance. can we merge and release to prod and I can test??

congyu-lu (approved):
Thanks

PR #3305, opened 2025-05-12
Use R as base image in docker image build

gavrielm:
# Description

In this PR, the base image of the pp docker image is now an R image instead of a NodeJS image. The R image is pulled from docker hub and then NodeJS and all other dependencies are installed on top of this image. R packages are installed and validated in the image using the same install script that will be used locally (`R/utils/install.pkgs.R`). This script determines which packages to install from two new text files that track the R dependencies for our codebase: `R/utils/cran.pkgs.txt` and `R/utils/bioconductor.pkgs.txt`.

Additionally, input arguments have now been matched across ppbase, ppserver, and ppfull `docker buildx build` commands in `container/deps/build.sh`. Previously, arguments were mismatched, which caused each build run to rebuild the deps image from scratch, rather than use cached steps. As a result, the build workflow now finishes in much less time (~20min) than before (~90min).

Can test by building and running the deps image by following the instructions here: https://github.com/stjude/proteinpaint/wiki/Buiding-a-deps-image

## Checklist


xzhou82 (approved):
tested multiple R features on local and all worked. thanks

container/deps/Dockerfile
```diff
@@ -58,6 +59,28 @@ RUN apt-get update && apt-get install -y \
...
+    nvm install $NODE_VERSION && \
+    nvm alias default $NODE_VERSION && \
+    nvm use $NODE_VERSION && \
+    echo 'export NVM_DIR="$NVM_DIR"' >> /root/.bashrc && \
+    echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> /root/.bashrc
+
+ENV PATH="/root/.nvm/versions/node/v${NODE_VERSION}/bin:$PATH"
+
+# Verify node installation
+RUN node -v
```
siosonel, line 80: should combine these two version check steps to minimize image layers

gavrielm: Done

siosonel (approved):

siosonel:
looks good to me, was able to build after pack.sh and run integration tests by setting `host: 'http://localhost:3456` in `sjpp/public/index.html`.

PR #3330, opened 2025-05-16
Updates to workflows for handling new base image

gavrielm:
# Description

Updates to build and release workflows for handling new docker image that is now based on an R base image in a debian testing environment.

## Checklist


client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sjcrh/proteinpaint-client",
-  "version": "2.124.1",
+  "version": "2.125.0",
```
siosonel, line 3: Did this get bumped automatically or manually? I think the versions in the shared/types, client front workspaces should not be bumped, as there are no code changes that occurred in these workspaces.

shared/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sjcrh/proteinpaint-types",
-  "version": "2.124.0",
+  "version": "2.125.0",
```
siosonel, line 3: Did this get bumped automatically or manually? I think the versions in the shared/types, client front workspaces should not be bumped, as there are no code changes that occurred in these workspaces.

front/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@sjcrh/proteinpaint-front",
-	"version": "2.124.1",
+	"version": "2.125.0",
```
siosonel, line 3: Did this get bumped automatically or manually? I think the versions in the shared/types, client front workspaces should not be bumped, as there are no code changes that occurred in these workspaces.

aacic (approved):

front/package.json
aacic: It got bumped by the script. We can manually revert it, since the version 2.124.1 is also published, but I would avoid that. 

 We need to verify how the script works.

client/package.json
aacic: It got bumped by the script.

siosonel (approved):

front/package.json
siosonel: Actually, the version bump looks okay. There were code changes in the shared/types and client workspaces between the last release and before this PR branch, so that caused the automatic version bump. I just needed to remember that the version bump has to be reviewed based changes since the last release, and not relative to master which is what's shown in a PR diff.

PR #3342, opened 2025-05-20
Added save zoom icon to save the zoom state

airenzp:
# Description

Do not save zoom state by default but allow to save it on a session created from the save session icon added to the zoom icons

## Checklist

### DIFF
--- a/client/dom/control.icons.js
+++ b/client/dom/control.icons.js
@@ -500,6 +500,15 @@ export const icons = {
 		</svg>`
 		return getHolder(elem, _opts).html(svg)
 	},
+	save: (elem, opts = {}) => {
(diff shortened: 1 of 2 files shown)


airenzp:
I forgot to check if there is a transformation before saving. The error you saw is fixed now, it happens if there is no zoom/pan done. Regarding the function of the save icon, I think a way to work it out would be to be able to save a session from there directly. I was thinking about it from yesterday, will look into it. Saving an svg is an icon from the controls,  not just from the scatter but from all the plots, I would rather keep it like it is as it impacts other plots. Otherwise we should discuss it with @xzhou82 and @siosonel before doing any changes.

xzhou82 (approved):

PR #3360, opened 2025-05-22
add download icon to lollipop sample table

karishma-gangwani:
# Description
sample table in lollipop chart now shows download option. @airenzp and @creilly8 could you help test and crosscheck if there's other places in the codebase that need this param. @xzhou82 is the name of the downloaded file suitable? or do you want something different? 

@creilly8, this could be something to add to the newsletter. 

## Checklist

### DIFF
--- a/client/mds3/sampletable.js
+++ b/client/mds3/sampletable.js
@@ -70,7 +70,7 @@ export async function displaySampleTable(samples, args) {
 		return await make_singleSampleTable(samples[0], args)
 	}
 	const [columns, rows] = await samples2columnsRows(samples, args.tk)
-	const params = { rows, columns, div: args.div, resize: rows.length > 10, dataTestId: 'sjpp_mds3tk_sampletable' }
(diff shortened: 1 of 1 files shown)


xzhou82:
download file name is not acceptable and should reflect context, which has some complexity.
also need to check if downloaded contents are valid

karishma-gangwani:
> download file name is not acceptable and should reflect context, which has some complexity. also need to check if downloaded contents are valid

ok. Could you point out where the name should come from? And yes, you're right, it does not print data for some of the columns. Maybe @airenzp and @creilly8 could help/suggest?

airenzp:
Thanks @karishma-gangwani. We can go through the details of the content that needs to be downloaded tomorrow. The filename should not have spaces. And I think Xin wants to build a filename that reflects the content. I created another PR related replacing the icon with a button, that I feel is better integrated in the UI. Please check [here](https://github.com/stjude/proteinpaint/pull/3363)

PR #3361, opened 2025-05-22
gdc cnv tool launcher will auto hide the <input> after search is done

siosonel:
# Description

Manually tested in local GFF by choosing the demo mode. To test outside of GFF, force lines 90-92 to true in `lollipop.js` to simulate GFF demo mode.

Also tested to make sure that lollipop does not break with these changes. For example, genesearch does not get hidden/disabled.

DONE ~~Note the TODO to be able to freeze/disable the gene/coord input in genome browser during demo mode, otherwise the demo text about MYC may not match the displayed track if the user searches for a different gene/coord.~~

## Checklist

### DIFF
--- a/client/gdc/lollipop.js
+++ b/client/gdc/lollipop.js
@@ -86,7 +86,10 @@ export async function init(arg, holder, genomes) {
 		tip,
 		row: geneInputDiv,
 		callback: launchView,
-		geneSymbol: arg.geneSymbol
+		geneSymbol: arg.geneSymbol,
+		triggerSearch: arg.geneSymbol && arg.geneSearch4GDCmds3?.hardcodeCnvOnly == true,
+		hideInputBeforeCallback: arg.geneSearch4GDCmds3?.hardcodeCnvOnly == true,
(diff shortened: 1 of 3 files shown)


client/gdc/lollipop.js
```diff
@@ -86,7 +86,10 @@ export async function init(arg, holder, genomes) {
 		tip,
 		row: geneInputDiv,
 		callback: launchView,
-		geneSymbol: arg.geneSymbol
+		geneSymbol: arg.geneSymbol,
+		triggerSearch: arg.geneSymbol && arg.geneSearch4GDCmds3?.hardcodeCnvOnly == true,
+		hideInputBeforeCallback: arg.geneSearch4GDCmds3?.hardcodeCnvOnly == true,
+		disableInput: arg.geneSearch4GDCmds3?.hardcodeCnvOnly == true
```
xzhou82, line 92: this causes `<input>` to be disabled by default

siosonel: thanks for fixing

xzhou82 (approved):

PR #3441, opened 2025-06-17
Fix security vulnerability in legacy single cell

creilly8:
# Description

## Checklist


client/src/singlecell.js
```diff
@@ -207,12 +211,6 @@ or selected a gene for overlaying
 }
 
 async function init_view(obj) {
-	// TODO only load below if to do 3d
-	await add_scriptTag('/static/js/three.js')
```
xzhou82, line 211: can delete these files from public/static/js/?

PR #3453, opened 2025-06-23
Modified DataMapper to prevent invalid fusions and added unit test fo…

ItsVictorLP:
…r DataMapper

# Description
This PR updates the fusion filtering logic in DataMapper.map() to ensure that only fusion events with both chrA and chrB present in the reference genome are included. A corresponding unit test has also been added to verify that fusion entries with invalid chromosome names are properly excluded.

test in:
http://localhost:3000/testrun.html?dir=plots/disco/data&name=DataMapper.unit

## Checklist

### DIFF
--- a/client/plots/disco/data/DataMapper.ts
+++ b/client/plots/disco/data/DataMapper.ts
@@ -121,7 +121,7 @@ export default class DataMapper {
 					this.addData(dObject, dataArray)
 				}
 			} else if (dObject.dt == dtfusionrna || dObject.dt == dtsv) {
-				if (indexA != -1 || indexB != -1) {
+				if (indexA != -1 && indexB != -1) {
 					// show fusion event when at least one gene knonw
 					this.addData(dObject, dataArray)
 				}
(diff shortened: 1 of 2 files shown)


client/plots/disco/data/DataMapper.ts
```diff
@@ -121,7 +121,7 @@ export default class DataMapper {
 					this.addData(dObject, dataArray)
 				}
 			} else if (dObject.dt == dtfusionrna || dObject.dt == dtsv) {
-				if (indexA != -1 || indexB != -1) {
+				if (indexA != -1 && indexB != -1) {
 					// show fusion event when at least one gene knonw
```
xzhou82, line 125: this comment no longer applies. revise as `show sv/fusion event with valid A/B breakpoints`

xzhou82: at line 129 for dtcnv and dtloh, the chr index validity is not validated. see if it should be done

ItsVictorLP: This has been changed

PR #3468, opened 2025-06-25
cap the number of records

josephpowi:
# Description
set up the cap of records to 100,000

## Checklist

### DIFF
--- a/container/deps/GDC-GRIN2-README.md
+++ b/container/deps/GDC-GRIN2-README.md
@@ -1,0 +1,97 @@
+# GDC-GRIN2 Dependencies Image Build Guide
+
+This guide walks you through building a new dependencies image for the GDC-GRIN2 project.
+
+## Prerequisites
+
+- Docker Desktop installed and running
+- Access to the `sjpp` and `proteinpaint` repositories
+- WARP disabled (see step 4)
+
+## Build Process
+
+### 1. Generate Dataset File
+
+Navigate to the `sjpp` root directory and generate a new hg38 dataset:
+
+```bash
+cd sjpp
+npm run cjs
+```
+
+### 2. Copy Dataset to Container
+
+Copy the generated dataset file to the proteinpaint container:
+
+```bash
+cp dataset/cjs/gdc.hg38.js proteinpaint/container/dataset/
+```
+
+### 3. Prepare Dependencies Directory
+
+Create a temporary packaging directory:
+
+```bash
+cd proteinpaint/container/deps
+mkdir tmppack
+```
+
+### 4. Disable WARP
+
+**Important:** Ensure WARP is turned off before proceeding.
+
+### 5. Version and Build Dependencies
+
+Generate version information and build the dependencies:
+
+```bash
+# Still in proteinpaint/container/deps
+./version.sh deps
+./build.sh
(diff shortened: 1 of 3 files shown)


server/routes/gdc.grin2.run.ts
```diff
@@ -44,16 +48,59 @@ function parseJsonlOutput(rustOutput: string): any {
 				const data = JSON.parse(trimmedLine)
 
 				if (data.type === 'data') {
+					// Check if we've already reached the cap
+					if (isCapReached) {
+						console.log(`[GRIN2] Skipping file ${data.case_id} - record cap of ${MAX_RECORDS} already reached`)
+						continue // Skip processing this file
```
xzhou82, line 54: break but not continue to loop through rest of lines

server/routes/gdc.grin2.run.ts
```diff
@@ -44,16 +48,59 @@ function parseJsonlOutput(rustOutput: string): any {
...
+						console.log(
+							`[GRIN2] RECORD CAP REACHED: ${MAX_RECORDS} records processed. Subsequent files will be skipped.`
+						)
+					}
 				} else if (data.type === 'summary') {
 					// Final summary - all files processed
 					finalSummary = data
 					console.log(`[GRIN2] Download complete: ${data.successful_files}/${data.total_files} files successful`)
+					if (isCapReached) {
+						console.log(`[GRIN2] Processing stopped due to record cap of ${MAX_RECORDS}`)
```
xzhou82, line 101: later: instead of console log, add this to run status and display on client; important to let user know this

container/deps/GDC-GRIN2-README.md
```diff
@@ -0,0 +1,97 @@
+# GDC-GRIN2 Dependencies Image Build Guide
```
xzhou82, line 1: maybe okay to have a separate readme for the moment, but we really need a single set of instructions to build deps image, because this is not in anyway special for grin2, it is pp deps image running on all environments. thus we should improve existing build doc instead of a separate one like this

server/routes/gdc.grin2.run.ts
compbiolover: Just now checked this and with break we don't get a return object and it fails. With continue we cap at the number of records and analysis completes as expected. Tested specifically on the erroring dataset that failed due to to many records

container/deps/GDC-GRIN2-README.md
compbiolover: Once deadlines are passed I will fold this readme into the existing set of instructions

server/routes/gdc.grin2.run.ts
xzhou82: i see i didn't understand the rustOutput structure. fine for now. later the max_record cutoff should be done in rust instead, to avoid piping large volume of useless data out

xzhou82 (approved):
works and need to move on to deps image build and deployment. resume below after deadline:
1. combine readme into general wiki page and do not maintain a separate readme for grin2
2. change rust to maintain grin2 input data in RAM with 100K cutoff, rust outputs grin2 lesion data and data status. this avoids outputing large volume of unused data

PR #3490, opened 2025-07-01
Do not get child dt terms when filling geneVariant tw

gavrielm:
# Description

Child dt terms are no longer requested when filling geneVariant tw. This removes the need to repeatedly call `getCategories()` for geneVariant terms in oncomatrix, and thereby removes a performance bottleneck ([example](http://localhost:3000/example.gdc.matrix.html?maxGenes=3)).

## Checklist


siosonel (approved):
Works for me, approving for now.

However, in a separate PR, should improve the performance when clicking on matrix row gene label, where the `Loading ....` message is displayed for maybe 20+ seconds. I think the general menu should be shown right away, but a non-blocking categories request could be triggered in the background, since the categories would not be needed if the `Remove` or `Replace` option is selected instead of `Edit`. 

Also, clicking on the same row label should use a cached categories response, should see why a new response is triggered every time.

PR #3492, opened 2025-07-01
cache ssm and cnv assayAvailability during gdc.initCache

siosonel:
# Description

Please pull the same branch from the `sjpp` repo.

To test: 
- load http://localhost:3000/example.gdc.matrix.html?&cohort=Gliomas, should load much faster

## Checklist

### DIFF
--- a/server/src/mds3.init.js
+++ b/server/src/mds3.init.js
@@ -2816,16 +2816,19 @@ async function mayAddDataAvailability(sample2mlst, dtKey, ds, origin, q) {
 		const sampleFilter =
 			!dt.get && q.filter && q.filter.lst.length ? new Set((await get_samples(q.filter, ds)).map(i => i.id)) : null
 
-		for (const sid of dt.yesSamples) {
-			// sample has been assayed
-			// if sample does not have annotated mutation for dt
-			// then it will be annotated as wildtype
-			addDataAvailability(sid, sample2mlst, dtKey, 'WT', dt.origin, sampleFilter)
-		}
-		for (const sid of dt.noSamples) {
-			// sample has not been assayed
-			// annotate the sample as not tested
-			addDataAvailability(sid, sample2mlst, dtKey, 'Blank', dt.origin, sampleFilter)
+		// shorter loop than iterating over initial cache of yesSamples and noSamples for all cases
+		// console.log(2780, 'sample2mlst.size=', sample2mlst.size )
+		for (const sid of sample2mlst) {
+			if (dt.yesSamples.has(sid)) {
+				// sample has been assayed
+				// if sample does not have annotated mutation for dt
+				// then it will be annotated as wildtype
+				addDataAvailability(sid, sample2mlst, dtKey, 'WT', dt.origin, sampleFilter)
+			}
+			if (dt.noSamples.has(sid)) {
+				// sample has not been assayed, annotate the sample as not tested
+				addDataAvailability(sid, sample2mlst, dtKey, 'Blank', dt.origin, sampleFilter)
+			}
 		}
 	}
 }
--- a/server/src/gdc.initCache.js
+++ b/server/src/gdc.initCache.js
@@ -227,6 +227,12 @@ async function cacheMappingOnNewRelease(ds, version) {
 
 		await getCasesWithGeneExpression(ds, ref)
 		await getAnalysisTsv2loom4scrna(ds, ref)
+		if (ds.assayAvailability.byDt) {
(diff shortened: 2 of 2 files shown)


shared/types/src/dataset.ts
```diff
@@ -1553,6 +1553,7 @@ type DtAssayAvailabilityByOrigin = {
 
 type Mds3AssayAvailability = {
 	/** object of key-value pairs. keys are dt values */
+	set?: (size: number) => Promise<void>
```
xzhou82, line 1556: should properly declare this callback with input and output

server/src/mds3.init.js
```diff
@@ -2816,16 +2816,19 @@ async function mayAddDataAvailability(sample2mlst, dtKey, ds, origin, q) {
 		const sampleFilter =
```
xzhou82, line 2816: this doesn't consider q.filter0

xzhou82:
the matrix indeed loads much faster with the optimization. still it doesn't work
as a simple test, [myc gene barchart](http://localhost:3000/?mass={%22dslabel%22:%22GDC%22,%22nav%22:{%22header_mode%22:%22hidden%22},%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22term%22:{%22type%22:%22geneVariant%22,%22gene%22:%22MYC%22}}}]}) should show wildtype/not tested categories for both ssm and cnv

server/src/mds3.init.js
xzhou82: i think this is viable. will need a method to get samples passing both q.filter and q.filter0.
then can replace byDt[].get()

siosonel:
> the matrix indeed loads much faster with the optimization. still it doesn't work as a simple test, [myc gene barchart](http://localhost:3000/?mass=%7B%22dslabel%22:%22GDC%22,%22nav%22:%7B%22header_mode%22:%22hidden%22%7D,%22genome%22:%22hg38%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22term%22:%7B%22type%22:%22geneVariant%22,%22gene%22:%22MYC%22%7D%7D%7D%5D%7D) should show wildtype/not tested categories for both ssm and cnv

For the `myc` gene barchart, the issue is that `sid` was not found in either `dt.yesSamples` or `dt.noSamples`. I thought that caching all of the 45087 total cases is comprehensive, so I'm not sure why any case id would not be matched. Is there a case id mapping that should be used to find `sid` in `[yes|no]Samples`?

Screenshot 2025-07-02 at 4 21 53 PM

shared/types/src/dataset.ts
siosonel: done

server/src/mds3.init.js
siosonel: I'll need more background on how this should work

congyu-lu: **sample2mlst** initially only contains samples with alteration, **mayAddDataAvailability()** loops samples in  yesSamples and noSamples, 

1. if a sample is in yesSamples but not in sample2mlst, annotate the sample as 'WT'
2. if a sample is in noSamples but not in sample2mlst, annotate the sample as 'Blank'

siosonel: @congyu-lu All of the above makes sense. When you say `cache`, I assume you mean within the same request for mutation data and happens every time a user loads a matrix or another plot, right? In contrast, when I say `cache` in this PR, I mean the yesSamples and noSamples were requested/computed only once during server startup as part of `gdc.initCache.js`, and that same cache is reused for all subsequent plot data requests.

So, the improvement in this PR is there are no GDC API requests needed to recompute yesSamples and noSamples, making for faster-loading plots. However, I do understand that when `q.filter0` or `q.filter` is present, then it's possible that the loops in this PR branch may not set the availability data for cases that do not have variant/alteration data, since  the loop is done on `sample2mlst`. Please let me know if you think I misunderstood anything above.

congyu-lu: By **cache** I meant using **dt.term_id** or call **dt.get()** to cache **ALL** samples/cases into two sets: yesSamples and noSamples, which is done only once during server startup (mayValidateAssayAvailability). 

But then we stoped the caching for dt.get() because we wanted to let dt.get() to handle filter0/filter instead. we still cache *ALL** samples/cases into two sets: yesSamples and noSamples for dt.term_id (once during server startup by mayValidateAssayAvailability)

PR #3506, opened 2025-07-07
Radius and fusion opacity controls for disco plot with unit tests

ItsVictorLP:
# Disco control improvements 
allow user to set plot radius
support adjustable opacity for fusion arcs

Unit test in:
http://localhost:3000/testrun.html?dir=plots/disco/viewmodel&name=ViewModelMapper.unit

Sv file for testing:
http://localhost:3000/?genome=hg19&dslabel=PNET&disco=1&sample=SJBT032239_D1

## Checklist

### DIFF
--- a/client/plots/disco/Disco.ts
+++ b/client/plots/disco/Disco.ts
@@ -96,47 +96,48 @@ export default class Disco {
 			configInputsOptions.push(...filterMutationsGenesCheckbox)
 		}
 
-		const cnvConfigInputOptions = [
-			{
-				boxLabel: '',
-				label: 'CNV rendering type',
-				type: 'radio',
-				chartType: 'Disco',
-				settingsKey: 'cnvRenderingType',
-				title: 'CNV rendering type',
-				options: [
-					{ label: 'Heatmap', value: CnvRenderingType.heatmap },
-					{ label: 'Bar', value: CnvRenderingType.bar }
-				]
-			}
-		]
-
-		if (viewModel.cnvMaxValue !== 0 || viewModel.cnvMinValue !== 0) configInputsOptions.push(...cnvConfigInputOptions)
-
-			const dimensionOptions = [
+		if (viewModel.cnvMaxValue !== 0 || viewModel.cnvMinValue !== 0) {
+			const cnvConfigInputOptions = [
 				{
-					label: 'Radius',
-					title: 'Set the radius of the entire plot, between 300 and 1000 pixels.',
-					type: 'number',
+					boxLabel: '',
+					label: 'CNV rendering type',
+					type: 'radio',
 					chartType: 'Disco',
-					settingsKey: 'radius',
-					debounceInterval: 500,
-					step: 25,
-					min: 300,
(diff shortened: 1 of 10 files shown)


client/plots/disco/viewmodel/ViewModel.ts
```diff
@@ -44,15 +44,22 @@ export default class ViewModel {
 		this.genesetName = genesetName
 
 		this.width =
-			1.2 *
+		//length multiplier (2) must be great enough to allow canvas to expand beyond 
+		//the expansion of the discoplot. If not, some of the labels will render outside
+		//the canvas and result in cut labels.
+			2 *
```
creilly8, line 50: We can make this cleaner by figuring out the pixel length of the longest label. There's a function called `getMaxLabelWidth` in `client/dom/maxLabelWidth.ts` for this purpose. Please see if you can use that to calculate the max label space and add to the width * 1.2 and height * 2. Then this will always be correct, no matter the edge case.

client/plots/disco/fusion/FusionRenderer.ts
```diff
@@ -7,7 +7,13 @@ import { table2col } from '#dom/table2col'
 
 // TODO extract constants from this file.
 export default class FusionRenderer {
-	render(holder: any, fusions: Array<Fusion>) {
+	private opacity: number
+
+	constructor(opacity = 1) {
```
creilly8, line 12: If `opacityOverride` (which is really the opacity setting) is passed to `render()` and not used anywhere else, there's no need to add it to the constructor. 

As an aside, @xzhou82 - should users be able to set the opacity to 0 and hide the fusions, or should the min value be something like 0.01?

client/plots/disco/Disco.ts
```diff
@@ -113,6 +113,33 @@ export default class Disco {
 
 		if (viewModel.cnvMaxValue !== 0 || viewModel.cnvMinValue !== 0) configInputsOptions.push(...cnvConfigInputOptions)
```
xzhou82, line 114: for multi-line if statement i prefer to quote it
```javascript
if (...) {
   const variable= [
     aaa
     bbb
  ]
}
```

creilly8 (approved):

xzhou82 (approved):
thank you both it's very nice! comments:

issue: when hovering on outside gene labels, in the tooltip the cnv colorbox color is incorrect and not based on cnv color scale

minor fix: on hovering a cnv segment, the tooltip first row header can print `CNV` instead of `Copy number change` and be consistent with gene label hover tooltip

suggestions:
1. may lower minimum radius from 300 to 200
2. default radius may be dynamically computed based on number of rings. e.g. [this gdc sample has only mutation ring but no cnv](http://localhost:3000/?genome=hg38&dslabel=GDC&disco=1&sample=MBCProject_1667_T1_WES) and radius=300 looks too big. can use this rule: assess how many dt are present in provided `mlst` for mutation|cnv|loh. if only 1, use 200 for radius; if 2, use 250; if 3, use 300

Issue #3487, opened 2025-07-01
reusable helper wrapping native fetch


xzhou82:
hopefully can be reused for client & back and allow to get rid of ky/node-fetch/got

PR #3507, opened 2025-07-08
Issue.3487

siosonel:
# Description

Closes https://github.com/stjude/proteinpaint/issues/3487. Also partially addresses slow load times/performance degradation.

`memFetch()` is the only shared helper that is used and tested here. It is used to reuse the same fetch promise/response for GDC cohort cases, to avoid multiple active duplicate requests for assay availability across genes. Note that later, may prefer to request cohort cases once and pass it as argument while looping through different terms in `getData()`, instead of using `memFetch()` repetitively.

`ezFetch()` is meant to replace `ky` and `node-fetch` and can be used and tested in a different branch.

Tested with unit and integration tests, also http://localhost:3000/example.gdc.matrix.html?&cohort=Gliomas should load in about 15 seconds, compared to 1-2 minutes in master.

## Checklist


xzhou82 (dismissed):
thanks a lot

xzhou82 (approved):

PR #3528, opened 2025-07-14
Report improvements in the filter and in the UI

airenzp:
# Description

This fix allows to combine the global filter with the report filter in the child plots, so the filter is always in sync with both filters. @siosonel I thought the getCombinedTermFilter name was the one that made more sense to me as it refers to what the function does, combine the term filter with the local filter, but I can rename it as needed. It also redesigns the UI to use the blank spaces organizing the sections in rows

## Checklist


client/filter/filter.utils.js
```diff
@@ -215,3 +215,8 @@ export function negateFilter(f0, tag = 'filterUiRoot') {
 	}
 	throw 'cannot negate filter'
 }
+
+export function getCombinedTermFilter(appState, filter) {
```
xzhou82, line 219: maybe easier
```javascript
function getCombinedTermFilter(appState,filter) {
  if(!filter || filter.lst?.length==0) return appState.termfilter
  return {filter: filterJoin([ ... ]), filter0: xx}
}
```

and use it as:

const termfilter = getCombinedTermFilter( appState,parentConfig.filter)

client/filter/filter.utils.js
```diff
@@ -215,3 +215,9 @@ export function negateFilter(f0, tag = 'filterUiRoot') {
...
+	if (!filter) return appState.termfilter
+	if (appState.termfilter.filter) filter = filterJoin([appState.termfilter.filter, filter])
```
xzhou82, line 221: this will modify parentConfig.filter and may cause issues

airenzp: Fixed

PR #3531, opened 2025-07-14
Report section show/hide fix

airenzp:
# Description
When showing/hiding the section toggle between flex or none, not block. Otherwise the plots are missaligned when viewing them again.

## Checklist

### DIFF
--- a/client/plots/report/report.ts
+++ b/client/plots/report/report.ts
@@ -49,7 +49,7 @@ export class Report extends RxComponentInner {
 				const display = sectionDiv.style('display')
 				headerIcon.text(display === 'none' ? '▼' : '▲') //toggle the icon
 				//toggle the display of the plot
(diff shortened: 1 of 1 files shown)


xzhou82 (approved):
works well!
a design fault is that barchart width is hardcoded. should make it a parameter, okay to keep current width as default, but in report chart, apply a smaller width to achieve compact look

airenzp:
> works well! a design fault is that barchart width is hardcoded. should make it a parameter, okay to keep current width as default, but in report chart, apply a smaller width to achieve compact look

Thanks for the feedback. In this PR I just removed the margin. I dont know why the barchart width is that one and as I am not familiar with the details of this logic. As for the report, the width per section is more or less the screen width, if the plots do not fit horizontally they move to the next line.

PR #3547, opened 2025-07-21
Disco Issue.867.legend improvements

ItsVictorLP:
Addresses the first item under Phase 2-> Legend improvements -> "Show ui option to list chromosomes ..." in https://github.com/stjude/sjpp/issues/867. 

Changes: 
1. Add .hiddenChromosomes:[] to settings 
2. Allow user to select which chrs are not rendered in the disco plot. When selecting a chr that contains the only element in a ring, that ring will be removed.
3. Error UI no longer throws error when chrs are hidden
4. Default settings addresses possible `termdbConfig.queries.singleSampleMutation.discoPlot.skipChrM` flag.
5. Multi-checkbox control layout improved. Renders via flexbox for a cleaner appearance. 
6. Apply runpp() defined settings to disco launch func, thus applying to the state

Test :
1. [PNET example](http://localhost:3000/?genome=hg19&dslabel=PNET&disco=1&sample=SJBT032239_D1). Check the hidden chromosome controls. 
2. [GDC example](http://localhost:3000/?genome=hg38&dslabel=GDC&disco=1&sample=TCGA-85-7710-01A). Should see chrM hidden by default. 

## Checklist

### DIFF
--- a/client/plots/disco/Disco.ts
+++ b/client/plots/disco/Disco.ts
@@ -46,18 +46,7 @@ export default class Disco {
 
 	async init() {
 		const state = this.app.getState()
-		const genomeChrs = Object.keys(this.app.opts.state.args.genome.majorchr)
-		let settings = state.plots.find(p => p.id === this.id).settings
-
-		if (!settings.Disco.selectedChromosomes || !settings.Disco.selectedChromosomes.length) {
-			await this.app.dispatch({
-				type: 'plot_edit',
-				id: this.id,
-				config: { settings: { Disco: { selectedChromosomes: genomeChrs } } }
-			})
-			const updatedState = this.app.getState()
-			settings = updatedState.plots.find(p => p.id === this.id).settings
-		}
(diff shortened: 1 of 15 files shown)


client/plots/disco/Disco.ts
```diff
@@ -46,7 +46,18 @@ export default class Disco {
 
 	async init() {
 		const state = this.app.getState()
-		const settings = state.plots.find(p => p.id === this.id).settings
+		const genomeChrs = Object.keys(this.app.opts.state.args.genome.majorchr)
+		let settings = state.plots.find(p => p.id === this.id).settings
+
+		if (!settings.Disco.selectedChromosomes || !settings.Disco.selectedChromosomes.length) {
+			await this.app.dispatch({
```
xzhou82, line 53: this is a bit convoluted.
i verified in line 193 of `getState()` method definition, it has access to `this.app.opts.state.args.genome`. thus this step should be moved there without using `await this.app.dispatch()`

PR #3575, opened 2025-07-28
handle term-level authorization

siosonel:
# Description

Also pull and test with the  `sjpp/auth-filter-sites-3` branch.

This branch handles term-level authorization, a new level of access authentication after portal-wide `jwt` and dataset-level credentials by server route. In this way, server route access may be allowed, but access privilege may still be restricted based on the term(s) in the request payload as well as the roles that are specified in the `jwt`. 

Design:
-  `clietnAuthResults{}` and `ignoredTermIds:[]` are added to `req.query.__protected__` in `app.middleware.js`. Previously, only `sessionId` was added to this object. This means  only having to pass these auth-related properties only once, instead of through many server route handler code and nested functions. 
- `authApi.mayExtendqFilter()` is called at the beginning of`getData()`, once all  `terms[]` are known regardless of what they are called in the request payload, and thereby all queried terms are protected by default and may be screened by the optional `ds.cohort.termdb.getAdditionalFilter()`.  
- a server route code may push specific term.ids. to `ignoredTermIds[]`, so that a server route that aggregate data can bypass term checks.

To test:
- open http://localhost:3000/profile/ `-> Graphs`: should still see a fully rendered polar plot, but the `Data Dictionary` plot should not list `Sites` as a term
-  open http://localhost:3000/profile/?role=user: should now see `Sites` in the `Data Dictionary`, but clicking on it should only render 3 bars corresponding to what the user's jwt payload/site access level
- open http://localhost:3000/profile/?role=admin: should see all terms in the `Data Dictionary`, and more chart options

## Checklist

### DIFF
--- a/server/routes/filterTermValues.ts
+++ b/server/routes/filterTermValues.ts
@@ -62,6 +62,7 @@ function getList(samplesPerFilter, filtersData, tw) {
 async function getFilters(query, ds, genome, res) {
 	try {
 		//Dictionary with samples applying all the filters but not the one from the current term id
+		if (query.ignoredTermIds) query.__protected__.ignoredTermIds.push(...query.ignoredTermIds)
 		const samplesPerFilter = await getSamplesPerFilter(query, ds)
 		const filtersData = await getData(
 			{
--- a/server/routes/filterTermValues.ts
+++ b/server/routes/filterTermValues.ts
@@ -1,6 +1,7 @@
 import type { RouteApi } from '#types'
 import { FilterTermValuesPayload } from '#types/checkers'
 import { getData, getSamplesPerFilter } from '../src/termdb.matrix.js'
+import { authApi } from '../src/auth.js'
 
 /*
 Given a set of terms and filters per term, this route returns the list of samples that match each term filter.
@@ -55,11 +56,16 @@ function getList(samplesPerFilter, filtersData, tw) {
 	}
 	filteredValues.unshift({ label: '', value: '' })
 	filteredValues.sort((a, b) => a.label.localeCompare(b.label))
-
 	return filteredValues
 }
 
 async function getFilters(query, ds, genome, res) {
+	// safe to process this client-submitted query.filterByUserSites flag,
+	// which only affects aggregation levels (not revealed sample-level data),
+	// as performed by this route code, and since query.terms would still
+	// be matched against a dataset's hiddenTermIds[]
+	if (!query.filterByUserSites) authApi.mayAdjustFilter(query, ds, query.terms)
+
 	try {
 		//Dictionary with samples applying all the filters but not the one from the current term id
 		const samplesPerFilter = await getSamplesPerFilter(query, ds)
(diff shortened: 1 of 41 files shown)


server/routes/profileScores.ts
```diff
@@ -37,6 +37,8 @@ function init({ genomes }) {
 }
 
 async function getScores(query, ds, genome) {
+	query.__protected__.ignoredTermIds.push(query.facilityTW.term.id)
```
airenzp, line 40: We need to add this line too in profileFormScores that is a similar route that calculates the scores for the template plots

server/src/app.middlewares.js
```diff
@@ -113,11 +113,12 @@ export function setAppMiddlewares(app, doneLoading) {
         by doing this, route code is worry-free and no need to pass "req{}" to gdc purpose-specific code doing the API calls
         these *protected* contents are not used in non-gdc code
         */
-		req.query.__protected__ = {}
+		const __protected__ = { ignoredTermIds: [] }
```
airenzp, line 116: Can we rename __protected__ to __userData__ or __protectedUserData__ or something? Something that indicates what this variable stores

server/src/termdb.server.init.ts
```diff
@@ -612,6 +615,7 @@ export function filterTerms(req, ds, terms) {
 
 const defaultCommonCharts: isSupportedChartCallbacks = {
 	dictionary: () => true,
+	summary: () => true,
```
xzhou82, line 618: should delete it since there's no longer a `Summary Plot` chart button

server/routes/filterTermValues.ts
```diff
@@ -65,7 +65,8 @@ async function getFilters(query, ds, genome, res) {
 		const samplesPerFilter = await getSamplesPerFilter(query, ds)
 		const filtersData = await getData(
 			{
-				terms: query.terms
+				terms: query.terms,
```
airenzp, line 68: Here should be the same as in the scores, we dont filter as we show aggregated data.

server/routes/filterTermValues.ts
```diff
@@ -55,23 +56,30 @@ function getList(samplesPerFilter, filtersData, tw) {
...
-
 	return filteredValues
 }
 
 async function getFilters(query, ds, genome, res) {
+	// safe to process this client-submitted query.filterByUserSites flag,
+	// which only affects aggregation levels (not revealed sample-level data),
+	// as performed by this route code, and since query.terms would still
+	// be matched against a dataset's hiddenTermIds[]
+	if (!query.filterByUserSites) authApi.mayAdjustFilter(query, ds, query.terms)
```
xzhou82, line 67: since this is a profile/sjcares-specific param, i would like to know if this route is actually ds-specific. if so may rename the route name and document such

server/routes/termdb.config.ts
```diff
@@ -98,7 +98,8 @@ function make(q, req, res, ds: Mds3WithCohort, genome) {
 		defaultChartType: ds.cohort.defaultChartType,
 		invalidTokenErrorHandling: tdb.invalidTokenErrorHandling,
 		colorMap: tdb.colorMap,
-		defaultTw4correlationPlot: tdb.defaultTw4correlationPlot
+		defaultTw4correlationPlot: tdb.defaultTw4correlationPlot,
+		authFilter: req.query.filter
```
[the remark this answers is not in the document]
siosonel: added comment

airenzp (approved):
Thanks a lot!

PR #3581, opened 2025-07-30
Updates on the auth filter 

airenzp:
# Description
This branch is on top of auth-filter-sites-3. Worked on it with @siosonel. Here I passed q to get_samples where needed. Added misc fixes on the auth handling and supported the auth filter for sjcares. Please see corresponding branch in [sjpp](https://github.com/stjude/sjpp/pull/952/files). I see some routes are broken when running the tests @siosonel. 

## Checklist

### DIFF
--- a/server/src/termdb.matrix.js
+++ b/server/src/termdb.matrix.js
@@ -372,7 +372,8 @@ export async function getSamplesPerFilter(q, ds) {
 	q.ds = ds
 	const samples = {}
 	for (const id in q.filters) {
-		const filter = filterJoin([q.filter, q.filters[id]])
+		let filter = q.filters[id]
+		if (!q.ignoreUserFilter && q.filter) filter = filterJoin([q.filter, q.filters[id]])
 		const result = (await get_samples({ filter, __protected__: q.__protected__ }, q.ds)).map(i => i.id)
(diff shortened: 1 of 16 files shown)


server/src/termdb.matrix.js
```diff
@@ -372,7 +372,8 @@ export async function getSamplesPerFilter(q, ds) {
 	q.ds = ds
 	const samples = {}
 	for (const id in q.filters) {
-		const filter = filterJoin([q.filter, q.filters[id]])
+		let filter = q.filters[id]
+		if (!q.ignoreUserFilter && q.filter) filter = filterJoin([q.filter, q.filters[id]])
```
siosonel, line 376: this flag cannot be set by the client, it will allow a public role to sidestep restrictions on what they can see

airenzp: if a public user says he wants to ignore the user filter (this is not possible from the UI but lets say from the route) it will see the average of all the sites which is the default behaviour.  If he wants to apply the auth filter, that is based on the sites he has access to in the jwt, he will get an empty filter. Either way he wont see sensitive info.

siosonel (approved):

PR #3582, opened 2025-07-30
handle a dataset minimum sample size option  before returning data

siosonel:
# Description

Tested by opening a [TermdbTest barchart](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22diaggrp%22},%22term2%22:{%22id%22:%22genetic_race%22}}],%22nav%22:{%22activeTab%22:1}}). Uncomment the minSampleSizeForFilterCharts in  `server/dataset/termdb.test.js`, there should be an error saying `Error: less than 100 samples with data for any queried term`. Lowering that number to 50 should get rid of the error.

Should also be tested with regression plot and other charts. Also, not sure whether the error bar is enough or if the message should be more nicely displayed.

## Checklist

### DIFF
--- a/client/plots/barchart.js
+++ b/client/plots/barchart.js
@@ -213,7 +213,7 @@ export class Barchart {
 					settingsKey: 'dedup',
 					boxLabel: 'Yes',
 					getDisplayStyle: plot =>
-						this.chartsData.charts.find(c => c.serieses.length != c.dedupedSerieses.length) ? 'table-row' : 'none'
+						this.chartsData?.charts.find(c => c.serieses.length != c.dedupedSerieses.length) ? 'table-row' : 'none'
 				}
 			]
 			if (this.hasStats)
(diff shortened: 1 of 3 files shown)


siosonel:
> Similar issue with [divideBy](http://localhost:3000/?mass=%7B%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22plots%22:%5B%7B%22chartType%22:%22summary%22,%22term%22:%7B%22id%22:%22diaggrp%22%7D,%22term0%22:%7B%22id%22:%22sex%22%7D%7D%5D,%22nav%22:%7B%22activeTab%22:1%7D%7D) chart. Sample number in each chart should probably not go below sample size cutoff.

I think these are okay, the total sample size is still 50+, hiding bars doesn't make the remaining rendered data more identifiable. So, maybe should change the config option name to `minSampleSizeForAggregatedData` instead of `minSampleSizeForFilterCharts`, this is not necessarily about what's actively rendered in each chart.

xzhou82:
thank you very much
@gavrielm please apply this setting of 5 samples to sjlife, test it and release

Issue #3569, opened 2025-07-27
summary plot won't rerequest and update after deleting revised geneVariant term2


xzhou82:
[test link](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22id%22:%22diaggrp%22},%22term2%22:{%22term%22:{%22type%22:%22geneVariant%22,%22gene%22:%22TP53%22}}}],%22nav%22:{%22activeTab%22:1}})
at Edit, click TP53 pill > select "CNV: gain vs loss" > chart updates > click TP53 pill again > Remove, chart does not update

PR #3583, opened 2025-07-31
geneVariant fixes

gavrielm:
# Description

Fixes https://github.com/stjude/proteinpaint/issues/3569

Also, radio buttons for single gene/gene set are now hidden in matrix to restrict input to single gene. This is because matrix is already designed to add genes one at a time to a gene set.

Also, fixed `mayMapGeneName2coord()` in `server/src/mds3.init.js` to use coordinate from term.chr/start/stop instead of from term.name.

Searching geneVariant by coordinate in gdc is still unresolved (see issue [here](https://github.com/stjude/proteinpaint/pull/3562/files#r2231371367)).

## Checklist

### DIFF
--- a/client/termdb/handlers/geneVariant.ts
+++ b/client/termdb/handlers/geneVariant.ts
@@ -11,11 +11,14 @@ export class SearchHandler {
 		this.callback = opts.callback
 		this.dom.radiosDiv = opts.holder.append('div')
 		this.dom.searchDiv = opts.holder.append('div')
+		/*
+		retain ability of adding geneset as row in matrix, in addition to one gene per row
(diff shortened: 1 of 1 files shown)


client/termdb/handlers/geneVariant.ts
```diff
@@ -11,6 +11,11 @@ export class SearchHandler {
 		this.callback = opts.callback
 		this.dom.radiosDiv = opts.holder.append('div')
 		this.dom.searchDiv = opts.holder.append('div')
+		if (this.opts.app?.opts?.state?.tree?.usecase?.target == 'matrix') {
```
xzhou82, line 14: it will be useful to add a geneset as rows in matrix, e.g. to mark out samples mutated in dmr pathway, rather than flatten out to several dozen gene rows

PR #3605, opened 2025-08-06
Fixed report filter recovered from a session. Should use the current user filter.

airenzp:
# Description

When creating a report, overwrite previous filters. Auth filter should always be the starting filter. This solves the issue of recovering a report from a session with different user access and loading an invalid filter.

## Checklist

### DIFF
--- a/client/plots/report/report.ts
+++ b/client/plots/report/report.ts
@@ -139,7 +139,8 @@ export async function getPlotConfig(opts, app) {
 
 	try {
 		const config = app.vocabApi?.termdbConfig?.plotConfigByCohort?.default?.report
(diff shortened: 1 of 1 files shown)


siosonel (approved):
Thanks, good for now. 

I've thought about this last week, the long term fix is to let the server apply the auth filter on the request, no need to pass the auth filter to the client. I'll work on this later.

siosonel:
> exactly. server should not trust an auth filter provided by client

The auth filter is always applied in the backend middleware, so it's still safe. But the filter is automatically loosened if there are no hidden terms detected in the processed request payload, which helps preserve the expected behavior in profile dataset. So, it's a matter of how the server knows when to not automatically loosen, it depends on the chart type and other context.

PR #3609, opened 2025-08-07
apply min sample size check to specific termdb routes

siosonel:
# Description

To test:
- pull the `sjpp/protected-fix` branch
- open http://localhost:3000/example.auth.html?dslabel=SJLife&route=termdb and follow the instructions to set up a fake-jwt
- create a filter with <10 samples, then open charts: should be an error when the user is not logged in, no error when logged in

## Checklist

### DIFF
--- a/server/src/auth.js
+++ b/server/src/auth.js
@@ -16,7 +16,7 @@ const defaultApiMethods = {
 	getJwtPayload, // declared below
 	canDisplaySampleIds: (req, ds) => {
 		if (!ds.cohort.termdb.displaySampleIds) return false
-		return authApi.userCanAccess(req, ds)
+		return authApi.userCanAccess(req, ds, protectedRoutes.samples)
 	},
 	// these open-acces, default methods may be replaced by maySetAuthRoutes()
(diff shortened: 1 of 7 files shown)


server/src/auth.js
```diff
@@ -205,10 +205,11 @@ function mayReshapeDsCredentials(creds) {
 	}
 }
 
-// TODO: should create a checker function for each route that may be protected
+// TODO: should create a checker function for each route group that may be protected
 const protectedRoutes = {
 	termdb: ['matrix'],
-	samples: ['singleSampleData', 'getAllSamples', 'scatter', 'convertSampleId', 'getSamplesByName']
+	samples: ['singleSampleData', 'getAllSamples', 'scatter', 'convertSampleId', 'getSamplesByName'],
+	minSampleSize: ['/termdb/barsql', 'matrix', 'cuminc', 'survival', 'regression', 'scatter']
```
siosonel, line 212: @gavrielm these server routes correspond to charts that we expect to have a min sample requirement, and not to other routes like `/termdb/categories` that break filter construction if a min sample size is enforced

PR #3622, opened 2025-08-11
Added stats handling to the violin

airenzp:
# Description
Added stats handling to the violin. Added fix in the barchart to show the default color option if no multicolorBars

## Checklist


xzhou82 (approved):
thanks. i realized in violin plot using term2, the p-values are from wilcoxon test from pairwise comparison, rather than "association test" that refers to fisher exact from barcharts. later may rename the Config option as well as property name to "Show p-value table"

PR #3656, opened 2025-08-25
Added function to combine svgs into a pdf. 

airenzp:
# Description
Reopened PR adding libraries jsPDF and svg2pdf.js as discussed. This PR adds a function to combine svgs into a pdf. Used it in the scatter when dividing by. Later on may be used to combine other plots like in the report (or to support the download of the open plots in the portal). See corresponding PR in [sjpp](https://github.com/stjude/sjpp/pull/1008/files) updating package-lock.json. 

## Checklist

### DIFF
--- a/client/common/svg.download.js
+++ b/client/common/svg.download.js
@@ -2,6 +2,7 @@ import { select } from 'd3-selection'
 import { to_svg } from '../src/client'
 import { jsPDF } from 'jspdf'
 import 'svg2pdf.js'
+import { add } from 'ol/coordinate'
 
 export function downloadSingleSVG(svg, filename, parent) {
 	if (parent) {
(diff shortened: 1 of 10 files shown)


client/common/svg.download.js
```diff
@@ -112,3 +115,25 @@ export function downloadChart(mainGsel, svgName, styleParent = null) {
 
 	to_svg(svg, svgName, { apply_dom_styles: true })
 }
+
```
siosonel, line 118: I think these pdf-related functions should be in a separate file, `pdf.download.js`, to not affect the chunk size of the `svg.download.js` bundler. This way, other charts/plots that do not offer a pdf option do not have to trigger a larger chunk (808 KB) import. Later, can have `dom/downloadMenu.js` to offer different ways to download, and each menu option would dynamically import either svg.download, pdf.download, etc.

airenzp:
Added downloadMenu component to select between pdf/svg  and dynamic import as suggested.

PR #3669, opened 2025-08-29
Allowed to download pdf or svgs in the barchart

airenzp:
# Description

Allowed to download pdf or svgs in the barchart. This replaces the previous barchart download combining multiple svgs into one, as they appear overlapped and maybe for the user is better to have  each plot in an independent svg. Please check if you want this approach.

## Checklist

### DIFF
--- a/client/plots/barchart.js
+++ b/client/plots/barchart.js
@@ -12,6 +12,7 @@ import { getColors, mclass, plotColor } from '#shared/common.js'
 import { isNumericTerm } from '#shared/terms.js'
 import { roundValueAuto } from '#shared/roundValue.js'
 import { getCombinedTermFilter } from '#filter'
+import { DownloadMenu } from '#dom/downloadMenu'
 
 export class Barchart {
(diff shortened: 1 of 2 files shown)


siosonel:
I think some users may still prefer being able to download one svg file that combines multiple charts. So, the download menu may have these options: 
- if there is no divide by: `SVG`, `PDF`
- if there is divide by: `SVG (one file)`, `SVG (zipped files)`, `PDF`

For `SVG (one file)`, we'd need to fix the combined svg height/offset calculations when the charts wrap around/are laid out vertically. It still works when the charts are laid out horizontally.

airenzp:
Yes sounds good, but the function to combine the plots needs to be moved to the download menu and use the list of svgs provided. Now this implementation only works for the barchart. It will take some time from me to understand it and move it. I am hoping I can do that in another PR if possible. I want to be able to generate the report pdf first for sjcares.

xzhou82 (approved):

PR #3723, opened 2025-09-16
Added support for a default filter in a dataset

airenzp:
# Description

@congyu-lu and I worked on this draft PR showing a possible implementation for the login in panMB. The idea is to add to the dataset a method getFilter that returns a filter that needs to be applied always in the portal. If this method is provided authApi.mayAdjustFilter will apply this filter instead. You can test it by opening a dictionary term through the [user](http://localhost:3000/panMB/?role=user) link and comparing it to the [public](http://localhost:3000/panMB) link. The barcharts show a different number of samples. Note that the about tab does not reflect the real number of samples, as this calculation would need to apply the user filter too. See corresponding PR in sjpp.

## Checklist

### DIFF
--- a/server/src/auth.js
+++ b/server/src/auth.js
@@ -29,7 +29,8 @@ const defaultApiMethods = {
 	getHealth: () => undefined,
 	// credentialed embedders, using an array which can be frozen with Object.freeze(), unlike a Set()
 	credEmbedders: [],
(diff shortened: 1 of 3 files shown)


airenzp:
Closing PR as we'll use getAdditionalFilter

PR #3734, opened 2025-09-17
Clean up on stats and boxplot updates

airenzp:
# Description

In this PR I removed duplicated code on stats, fixed the variance calculation (should use n -1 always) and supported removing outliers in the boxplot. I also removed the need for a second request to caculate the stats as this can be calculated on the boxplot request and sent to client, and it  only needs to be done for the term plotted.  I see this second request is also done  in the barchart and the violin. The same logic applies in those cases, the stats can be done in the main request.  
I supported hiding outliers as I needed it to see the timelines in the sjcares report properly. There is a corresponding [PR](https://github.com/stjude/sjpp/compare/report-timeline?expand=1) showing boxplots on the sjcares report.

Default:

Screenshot 2025-09-17 at 11 50 24 AM

After hiding outliers:

Screenshot 2025-09-17 at 11 50 30 AM

Please check

## Checklist

### DIFF
--- a/client/plots/boxplot/BoxPlot.ts
+++ b/client/plots/boxplot/BoxPlot.ts
@@ -248,6 +248,9 @@ class TdbBoxplot extends RxComponent {
 			const settings = config.settings.boxplot
 			const model = new Model(config, this.state, this.app, settings)
 			const data = await model.getData()
+			console.log(data)
+			config.term.q.descrStats = data.descrStats
+
 			if (data.error) throw data.error
 			if (!data.charts || !Object.keys(data.charts).length) {
 				this.interactions!.clearDom()
--- a/client/plots/boxplot/BoxPlot.ts
+++ b/client/plots/boxplot/BoxPlot.ts
@@ -248,7 +248,6 @@ class TdbBoxplot extends RxComponent {
 			const settings = config.settings.boxplot
 			const model = new Model(config, this.state, this.app, settings)
 			const data = await model.getData()
-			console.log(data)
 			config.term.q.descrStats = data.descrStats
 
 			if (data.error) throw data.error
@@ -258,7 +257,6 @@ class TdbBoxplot extends RxComponent {
 				return
 			}
 			this.data = data
-
 			this.dom.charts.selectAll('*').remove()
 
 			// determine max label length across all charts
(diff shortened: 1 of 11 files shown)


server/routes/termdb.boxplot.ts
```diff
@@ -1,12 +1,9 @@
...
+import type { BoxPlotRequest, BoxPlotResponse, RouteApi, ValidGetDataResponse } from '#types'
 import { boxplotPayload } from '#types/checkers'
 import { getData } from '../src/termdb.matrix.js'
 import { boxplot_getvalue } from '../src/utils.js'
 import { sortPlot2Values } from './termdb.violin.ts'
-import { roundValueAuto } from '#shared/roundValue.js'
-import { getMean, getVariance } from '#shared/descriptive.stats.js'
-
-const minSampleSize = 5 // a group below cutoff will not compute boxplot
+import { summaryStats, getDescriptiveStats, summaryStatsFromStats } from '#shared/descriptive.stats.js'
```
gavrielm, line 6: I am wondering if there is a way to unify these 3 functions.

airenzp: getDescriptiveStats returns the stats properties in an object, so one can read them directly. For example to filter out the outliers I use stats.outlierMin. If we only use summaryStats and get an array of values to get the outlierMin I would need to find the property in the array first. And summaryStatsFromStats is to get the legend array once you have the stats. Please see the different uses in this PR to see how they are used.

client/plots/boxplot/BoxPlot.ts
```diff
@@ -14,19 +14,29 @@ import { BoxPlotInteractions } from './interactions/BoxPlotInteractions'
...
 	private useDefaultSettings = true
-	constructor(opts: TdbBoxPlotOpts) {
-		super()
+	parentId?: string
+	api: any
+
+	constructor(opts: TdbBoxPlotOpts, api: MassAppApi) {
+		super(opts)
 		this.opts = opts
+		this.api = api
```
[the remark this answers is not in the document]
airenzp: Please check with @siosonel, this property is passed now.

server/routes/termdb.boxplot.ts
gavrielm: Seems like `getDescriptiveStats` can serve as the single function. Don't think `summaryStats` is as useful, because the result is nested under `.values[]`. And if we get rid of `summaryStats`, we also don't need `summaryStatsFromStats`.

I may be naive about this, so correct me if I am wrong. And this would require a refactor, so can be done in a separate branch (I can help with this too).

xzhou82: i agree with Gav `descriptive.stats.js` can use with some rework

airenzp: Could you clarify what do you mean?

PR #3740, opened 2025-09-18
GRIN2:Miscellaneous Polish & Improvements

compbiolover:
# Description
Renamed `plot_grin2_manhattan` function to `plot_manhattan` and moved it  from `grin2PpWrapper.py` to its own file (`manhattan.py`). Slightly tweaked y-axis padding in the png image to keep points from getting clipped. Replaced matplotlib with tkinter for more control over plotting. Added download option for the svg to `plotManhattan`. Added click event to open genome browser for interactive genes in the plot. Commented out the many `mayLog` lines in the general grin2 route to make it cleaner.

# To Test
Go [here](http://localhost:3000/?mass={%22dslabel%22:%22ASH%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:-1},%22plots%22:[{%22chartType%22:%22grin2%22}],%22termfilter%22:{%22filter%22:{%22type%22:%22tvslst%22,%22join%22:%22%22,%22in%22:true,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Diagnosis%22},%22values%22:[{%22key%22:%22AML%22}]}}]}}}) and [here](http://localhost:3000/?gdcgrin2=1)
## Checklist


xzhou82:
thanks for the concern. plots/manhattan is definitely quick fix, and should be refactored as plots/volcano/

compbiolover:
> Gentlemen, from what I understand there is a mismatch of goals here. Its my understanding a future goal is allowing users to change the dimensions of the manhattan plot such as the height and width. This code pattern will make it difficult long term to introduce controls and saving those changes to the state. That also means users won't be able to share a link with a modified manhattan plot to someone else. It will always launch with default settings. Furthermore it will be difficult to introduce any unit testing for this plot now that the plot dimensions are mixed with the rendering code.
> 
> If the immediate goal is to get something working quickly, that's fine. But do be aware what I see conflicts with the stated long term goals. I'm concerned this is coding into a corner rather than taking advantage of every possible workflow, method, pattern, and system proteinpaint has to offer.

Thanks so much for the feedback. Will chat with you on Monday about addressing the issues you see so it can align with plots/volcano for easier maintainability in the future

PR #3749, opened 2025-09-19
fix: to enable blat, run gfServer as a child process

xzhou82:
# Description
this runs gfServer as a long-running child process inside the PP container
as alternative to `container/blat/Dockerfile` that runs gfServer in a separate container
this might facilitate enabling blat for GDC
## Checklist

### DIFF
--- a/server/src/initGenomesDs.js
+++ b/server/src/initGenomesDs.js
@@ -11,6 +11,8 @@ import { mds_init } from './mds.init.js'
 import * as mds3_init from './mds3.init.js'
 import { parse_textfilewithheader } from './parse_textfilewithheader.js'
 import { clinsig } from '../dataset/clinvar.ts'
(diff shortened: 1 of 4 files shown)


server/src/initGenomesDs.js
```diff
@@ -132,10 +134,8 @@ export async function initGenomesDs(serverconfig) {
 			g2.snp = g.snp
 		}
 		if (g.blat) {
-			if (!g.blat.host) throw '.blat.host missing for ' + g.name
-			if (!g.blat.port) throw '.blat.port missing for ' + g.name
-			// gfServer must start with full path to 2bit file
-			g2.blat = g.blat // enable blat
+			g2.blat = g.blat
+			initBlatServer(g2)
```
xzhou82, line 138: note this is not awaited on

PR #3755, opened 2025-09-22
Termdb bcf: TermdbTest.bcf.gz

josephpowi:
update Termdb bcf:
1. add dna_assay and tumor_DNA_WGS in FORMAT field.
2. add CLNSIG in INFO field.


josephpowi (commented):
two germline sample are back now

xzhou82 (approved):

PR #3768, opened 2025-09-25
Refactor on profile filtering. Passing cohort selected to getAdditionalFilter

airenzp:
# Description

Refactor on profile filtering. Passing cohort selected to getAdditionalFilter. Still testing as there are a lot of changes involved. That is why this is a draft.

## Checklist


airenzp:
Thanks for reporting this bug. Now I only add WITH if the filter is provided. Also, getSingleSampleData may return empty rows if no terms are displayed in the tree upon start, like in the example you reported. The check to see if no sample was found should be made on an independent query that runs before (if a filter is provided).

xzhou82:
thanks it no longer breaks now. i also verified the previous error can be caught by sampleview CI test, so let's make sure all integration tests pass

siosonel:
I reused checkAccessToSampleData in `getSingleSampleData()`, please test after also pulling the corresponding sjpp PR branch

PR #3771, opened 2025-09-26
define qualitative types and tw

siosonel:
# Description

The new `qualitative` types and classes will be used as a base class for categorical, snp, and other applicable term types. For now, in this branch, only `snp` is using the new defs/classes.  This approach is similar to the `numeric` types and classes that are used for integer, float, and other applicable terms. 

Tested by running all unit and integration tests. If this is approved, then I'll migrate `singleCellCellType` and `categorical` to this approach.

Note that there was an initial attempt to reuse`shared/types/terms/categorical.ts` types and `client/tw/categorical` classes for `snp`. That didn't work because `categorical` type definitions are already narrowed (e.g., `term.type = 'categorical'`), whereas it's okay to narrow a definition with type unions to a single included type like `snp`, which is the way that `numeric` type definitions work when reused for `geneExpression` and other term types.

## Checklist


gavrielm:
Open [snp barchart](http://localhost:3000/?mass={%22dslabel%22:%22SJLife%22,%22genome%22:%22hg38%22,%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22term%22:{%22id%22:%22rs1641548%22,%22chr%22:%22chr17%22,%22start%22:7671460,%22stop%22:7671461,%22name%22:%22rs1641548%22,%22ref%22:%22C%22,%22alt%22:[%22T%22],%22type%22:%22snp%22}}}],%22nav%22:{%22activeTab%22:1}}) -> set groupset -> apply -> throw this error:
image

PR #3836, opened 2025-10-20
fix the wilcoxon test rust code to make it work with new HDF5 format

josephpowi:
fix the wilcoxon test rust code to make it work with new HDF5 format

please pull and rebuilt rust code!!!


xzhou82 (dismissed):

xzhou82:
thanks a lot works. but the existing unit test is broken. could you look into it?
`~/dev/proteinpaint/rust % node test/DEanalysis.unit.spec.js`

josephpowi (commented):
fix unit test DEanalysis.unit.spec.js

PR #3892, opened 2025-11-04
Replacing `block.js` imports with genome browser dispatch

gavrielm:
# Description

Addressing issues listed in https://github.com/stjude/sjpp/issues/1071

- In `matrix.interactivity.js`, replaced `block.js` import with dispatch to `GB.ts`.
- In `barchart.events.js`, removed the `block.js` import as it does not seem to be used (please confirm).
- Removed `plots/variantBrowser.js` because it is only used by `launchBrowser()` in `matrix.interactivity.js` and `launchBrowser()` is not called anywhere.
- When launching genome browser from custom variable, the presubmission UI is now shown.

## Checklist


siosonel:
"In barchart.events.js, removed the block.js import as it does not seem to be used (please confirm)."
I didn't find any `select_to_gp` string in the codebase, so it doesn't seem like this barclick menu option is used anymore. There may have been a time when it was used, back when `term.isgenotype` was still supported, but that has been replaced with `term.type: 'geneVariant'`.

xzhou82 (approved):

PR #3906, opened 2025-11-11
GRIN2: Added lesion type indicator to top gene table. Other cleanups 

compbiolover:
# Description
Added lesion type indicators for cases where they are significant. Now using `axisstyle` for the d3 y-axis styling. Used the value of dtUsage in state to determine which q-value columns to search for instead of just iterating through all data type q-value columns.  Made a combined map as the q-value column titles and the names assigned to the legend are different. By combining them into a single map we make it more concise. 

# To Test
Go [here](http://localhost:3000/?mass={%22dslabel%22:%22ASH%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:-1},%22plots%22:[{%22chartType%22:%22grin2%22}],%22termfilter%22:{%22filter%22:{%22type%22:%22tvslst%22,%22join%22:%22%22,%22in%22:true,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Diagnosis%22},%22values%22:[{%22key%22:%22AML%22}]}}]}}}) and test that the various lesion significance indicators show up. Can use the table sort to make it fast to see specific lesion types

# To-do
Long term the colors coming from default settings will be used in rust backend code in a state specific manner so that only checked data type colors are sent to backend (right now rust just generates all colors) 

# Closes
GRIN2 roadmap item 3
## Checklist

### DIFF
--- a/client/plots/grin2/grin2.ts
+++ b/client/plots/grin2/grin2.ts
@@ -655,25 +655,25 @@ class GRIN2 extends PlotBase implements RxComponent {
 
 			// Find column indices for q-values
 			const columns = result.topGeneTable.columns
-			console.log('this:', this)
-			// Map dtUsage keys to their Q-value column labels
-			const dtToQValueColumn: Record<number, string[]> = {
-				[dtsnvindel]: ['Q-value (Mutation)'],
-				[dtfusionrna]: ['Q-value (Fusion)'],
-				[dtcnv]: ['Q-value (Copy Loss)', 'Q-value (Copy Gain)'],
-				[dtsv]: ['Q-value (Structural Variant)']
+
+			// Map dt types to their column labels and lesion types
+			const dtMapping = {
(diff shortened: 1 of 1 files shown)


client/plots/grin2/grin2.ts
```diff
@@ -642,32 +644,83 @@ class GRIN2 extends PlotBase implements RxComponent {
...
 
+			// Define lesion type colors and q-value threshold
+			const lesionTypeColors = this.state.config.settings.manhattan.lesionTypeColors
+			const qValueThreshold = this.state.config.settings.manhattan.qValueThreshold
+
+			// Find column indices for q-values
+			const columns = result.topGeneTable.columns
+
+			// Map dt types to their column labels and lesion types
+			const dtMapping = {
```
xzhou82: will suggest a better way to avoid all these duplicated definitions

PR #3997, opened 2025-12-04
GRIN2: Enabled Burger Menu 

compbiolover:
# Description
Enabled burger menu to allow user to customize plot width, height, and dot radius of the manhattan plot. This is done by using the cache file that is generated upon first run of GRIN2. When a user updates these values and the route is called again we simply check for the cache file and use it instead of re-running the full analysis. I put the newly used dom in `this.dom`. Kept`lastPlotResult` as a private property of the GRIN2 class. I didn't think it would be a good idea to store things like the full response value (e.g. plot png, timmings, etc) in the manhattan settings of state and instead compromised and did this instead. I noticed that if decimals were entered in the burger menu it would cause the re-render to fail. I sanitize the inputs and force decimal values to their floor value. Integrated the download functionality to use the burger menu download ui button instead of custom button. Setting to draft to address copilot feedback

# TODO
Cleanup grin2 client as it is getting large now

# To Test
Go to [tdbtest](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:{%22activeTab%22:1},%22plots%22:[{%22chartType%22:%22grin2%22}]}) and [ASHOP](http://localhost:3000/?mass={%22dslabel%22:%22ASH%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:-1},%22plots%22:[{%22chartType%22:%22grin2%22}],%22termfilter%22:{%22filter%22:{%22type%22:%22tvslst%22,%22join%22:%22%22,%22in%22:true,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Diagnosis%22},%22values%22:[{%22key%22:%22AML%22}]}}]}}}) and run analysis. Then change settings in burger menu. Plot should re-render very quickly. Also test the share session to make sure the settings of the burger menu carry over. 
## Checklist

### DIFF
--- a/client/plots/grin2/grin2.ts
+++ b/client/plots/grin2/grin2.ts
@@ -150,7 +150,7 @@ class GRIN2 extends PlotBase implements RxComponent {
 		})
 
 		// Normalize number inputs to the integer values we actually use on the backend.
-		// This and the backend sanitatization allow the server to not error if the user types in invalid values (e.g. decimals) and hang the re-render
+		// This and the backend sanitization allow the server to not error if the user types in invalid values (e.g. decimals) and hang the re-render
 		this.dom.plotControls
 			.selectAll('input[type="number"]')
 			.on('change.grin2IntSanitize', function (this: HTMLInputElement) {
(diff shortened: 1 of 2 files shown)


client/plots/grin2/grin2.ts
```diff
@@ -8,10 +8,23 @@ import { dtsnvindel, mclass, dtcnv, dtfusionrna, dtsv, proteinChangingMutations,
 import { get$id } from '#termsetting'
 import { PlotBase } from '#plots/PlotBase.ts'
 import { plotManhattan, createLollipopFromGene } from '#plots/manhattan/manhattan.ts'
+import { controlsInit } from '#plots/controls.js'
```
xzhou82: see if possible to move the controls component to manhattan.ts instead of grin2
thus the same controls will be available for gwas results, as grin2 ui won't be used for that

compbiolover:
Closing as it will be better to move controls into manhattan.ts in a future branch

PR #4034, opened 2025-12-17
Mutation count setting for geneVariant term

gavrielm:
# Description

Implemented a mutation count setting (any, single, multiple) for geneVariant term. To implement this, the wildtype mclass value needed to be treated separately from mutation mclass values; therefore now there is also a new genotype setting (mutated/altered vs. wildtype).

`dom/variantConfig.ts` now handles geneVariant config UI rendering for all dts, except for continuous CNV, which is handled by `dom/cnvConfig.ts`.

Added and updated both client-side and server-side tests. Please test further.

## Checklist

### DIFF
--- a/release.txt
+++ b/release.txt
@@ -1,1 +1,2 @@
-
+Features
+- Added mutation count setting to geneVariant term


xzhou82 (commented):
thank you so much works well
will go ahead and merge now with great trust :)

xzhou82 (approved):

PR #4037, opened 2025-12-17
GRIN2: Improved Gene tool tips to display q-value

compbiolover:
# Description
Now displaying q-value in manhattan tooltips and click tables instead of the -log10(q-values). Also added a check for if we have the scaled the y-axis and add a [capped] label to the y-axis to make that clear to users when viewing the plot. Also capped the number of genes that can be sent to the matrix. The limit is 100 genes. If a user selects more than 100 we simply truncate the list after the first 100. 

# To test
Go to both [tdbtest](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:{%22activeTab%22:-1},%22plots%22:[{%22chartType%22:%22grin2%22}]}) and [ASHOP](http://localhost:3000/?mass={%22dslabel%22:%22ASH%22,%22genome%22:%22hg38%22,%22nav%22:{%22activeTab%22:-1},%22plots%22:[{%22chartType%22:%22grin2%22}],%22termfilter%22:{%22filter%22:{%22type%22:%22tvslst%22,%22join%22:%22%22,%22in%22:true,%22lst%22:[{%22type%22:%22tvs%22,%22tvs%22:{%22term%22:{%22id%22:%22Diagnosis%22},%22values%22:[{%22key%22:%22AML%22}]}}]}}}). On tdbtest you should see no capped text on the y-axis label. On the ASHOP plot you should see the additional capped text.

# Closes
GRIN2 roadmap 6  
## Checklist

### DIFF
--- a/client/plots/manhattan/manhattan.ts
+++ b/client/plots/manhattan/manhattan.ts
@@ -233,7 +233,7 @@ export function plotManhattan(div: any, data: any, settings: any, app?: any) {
 		.attr('text-anchor', 'middle')
 		.attr('font-size', `${settings.fontSize + 4}px`)
 		.attr('fill', 'black')
-		.text(data.plotData.y_max >= 40 ? '-log₁₀(q-value) [scaled]' : '-log₁₀(q-value)')
+		.text(data.plotData.y_max >= 40 ? '-log₁₀(q-value) [capped]' : '-log₁₀(q-value)')
 
 	// Add png image
 	svg


client/plots/manhattan/manhattan.ts
```diff
@@ -233,7 +233,7 @@ export function plotManhattan(div: any, data: any, settings: any, app?: any) {
 		.attr('text-anchor', 'middle')
 		.attr('font-size', `${settings.fontSize + 4}px`)
 		.attr('fill', 'black')
-		.text('-log₁₀(q-value)')
+		.text(data.plotData.y_max >= 40 ? '-log₁₀(q-value) [scaled]' : '-log₁₀(q-value)')
```
xzhou82: good fix. call it `capped` instead of `scaled`?

compbiolover: Good point. Capped makes it more clear

PR #4096, opened 2026-01-13
Nodejs chatbot (WIP)

robinpaul85:
# Description

1) In this PR, the existing logic for the summary agent and classification agent has been translated from rust to nodejs.

2) In classification agent, add option to provide links to publications and other potential information specific to a dataset.

## Checklist

### DIFF
--- a/server/dataset/ai/termdb.test.json
+++ b/server/dataset/ai/termdb.test.json
@@ -5,10 +5,10 @@
   "charts": [
     {
         "type":"Classification",
-	"SystemPrompt":"The first original publication pertaining to Proteinpaint is https://www.cell.com/cancer-cell/fulltext/S1535-6108(20)30659-0?dgcid=raven_jbs_etoc_email. GenomePaint/Proteinpaint is an interactive visualization platform for whole-genome, whole-exome, transcriptome, and epigenomic data of tumor samples. Its design captures the inter-relatedness between DNA variations and RNA expression, supporting in-depth exploration of both individual cancer genomes and full cohorts. \n Link to survovorship portal https://aacrjournals.org/cancerdiscovery/article/14/8/1403/746507. It shows datasets comprising demographic, diagnosis, treatment, outcome, and genomic data from survivors of childhood cancer.",
+	"SystemPrompt":"The first original publication pertaining to Proteinpaint is https://www.cell.com/cancer-cell/fulltext/S1535-6108(20)30659-0?dgcid=raven_jbs_etoc_email. GenomePaint/Proteinpaint is an interactive visualization platform for whole-genome, whole-exome, transcriptome, and epigenomic data of tumor samples. Its design captures the inter-relatedness between DNA variations and RNA expression, supporting in-depth exploration of both individual cancer genomes and full cohorts. \n Link to survovorship portal https://aacrjournals.org/cancerdiscovery/article/14/8/1403/746507. It shows datasets comprising demographic, diagnosis, treatment, outcome, and genomic data from survivors of childhood cancer. Put weblinks inside HTML <a> tag so that links are clickable",
 	"TrainingData":[
-            {"question":"Show link to survivorship portal", "answer": {"type":"html","html":"https://aacrjournals.org/cancerdiscovery/article/14/8/1403/746507"}},
-	    {"question":"Show main publication for proteinpaint", "answer": {"type":"html","html":"https://www.cell.com/cancer-cell/fulltext/S1535-6108(20)30659-0?dgcid=raven_jbs_etoc_email"}},
+            {"question":"Show link to survivorship portal", "answer": {"type":"html","html":"<a href=\"https://aacrjournals.org/cancerdiscovery/article/14/8/1403/746507\">Visit weblink</a>"}},
+	    {"question":"Show main publication for proteinpaint", "answer": {"type":"html","html":"<a href=\"https://www.cell.com/cancer-cell/fulltext/S1535-6108(20)30659-0?dgcid=raven_jbs_etoc_email\">Visit weblink</a>"}},
             {"question":"compare tp53 expression between genders", "answer": {"type":"plot","plot":"Summary"}},
 	    {"question":"Show summary of all molecular subtypes for patients with age from 10 to 40 years", "answer": {"type":"plot","plot":"Summary"}},
 	    {"question":"Show correlation between diagnosis groups and ancestry", "answer": {"type":"plot","plot":"Summary"}},


server/routes/termdb.chat.ts
```diff
@@ -1,9 +1,16 @@
...
+import path from 'path'
+import type { ChatRequest, ChatResponse, RouteApi, DbRows, DbValue, SummaryType } from '#types'
 import { ChatPayload } from '#types/checkers'
-import { run_rust } from '@sjcrh/proteinpaint-rust'
+//import { run_rust } from '@sjcrh/proteinpaint-rust'
 import serverconfig from '../src/serverconfig.js'
 import { mayLog } from '#src/helpers.ts'
-import { run_python } from '@sjcrh/proteinpaint-python'
+//import { run_python } from '@sjcrh/proteinpaint-python'
+import got from 'got'
```
xzhou82: got has been removed. replace with `import {ezFetch} from '#shared'`

server/routes/termdb.chat.ts
```diff
@@ -56,122 +43,456 @@ function init({ genomes }) {
...
+		expose: 'export' as 'export' | 'all' | 'none' | undefined,
+		// Put the whole schema under a top‑level $ref (optional but convenient)
+		topRef: true,
+		// Turn off type‑checking for speed (set to true if you want full checks)
+		skipTypeCheck: true
+	}
+	const generator: SchemaGenerator = createGenerator(SchemaConfig)
+	const StringifiedSchema = JSON.stringify(generator.createSchema(SchemaConfig.type)) // This will be generated at server startup later
+	const words = prompt.split(/\s+/).map(str => str.toLowerCase()) // Split on whitespace and convert to lowercase
+	const common_genes = words.filter(item => genes_list.includes(item)) // The reason behind showing common genes that are actually present in the genedb is because otherwise showing ~20000 genes would increase the number of tokens significantly and may cause the LLM to loose context. Much easier to parse out relevant genes from the user prompt.
```
compbiolover: I think this logic needs to be more robust. If I enter a gene name with a . (e.g. TP53. vs. TP53) I get `invalid term id:`error.

server/routes/termdb.chat.ts
```diff
@@ -56,122 +43,456 @@ function init({ genomes }) {
...
+			)
 			const time2 = new Date().valueOf()
 			mayLog('Time taken for classification:', time2 - time1, 'ms')
+			mayLog('classResult:', classResult)
 
-			let ai_output_data: any
+			//let ai_output_data: any
 			let ai_output_json: any
-			if (classResult.route == 'summary') {
+			if (classResult == 'summary') {
```
compbiolover: This is likely outside the scope of this pr but I discovered that prompting with `Show me the karyotype of chr x` vs. `Show me the karyotypes of chr x` gives very different results. In the former it gives `Unknown classification value` while the latter give the expected summary chart of the the karyotypes for chrX. Probably should think about how to make it so the user doesn't have to be absolutely right on with term specification for it to work

server/routes/termdb.chat.ts
```diff
@@ -56,122 +43,456 @@ function init({ genomes }) {
...
-			const classResult = JSON.parse(await run_rust('query_classification', JSON.stringify(chatbot_input)))
+			const classResult = await classify_query_by_dataset_type(
+				q.prompt,
+				comp_model_name,
+				serverconfig.llm_backend,
+				apilink,
+				serverconfig.aiRoute
+			)
 			const time2 = new Date().valueOf()
 			mayLog('Time taken for classification:', time2 - time1, 'ms')
```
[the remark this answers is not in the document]
robinpaul85: Done

server/routes/termdb.chat.ts
```diff
@@ -56,122 +43,456 @@ function init({ genomes }) {
...
+			pp_plot_json.filter = validated_summary_type.simpleFilter
+		return { type: 'plot', plot: pp_plot_json }
+	}
+}
+
+function validate_filter(filters: any, ds: any): any {
+	if (!Array.isArray(filters)) throw 'filter is not array'
+	let invalid_html = ''
+	const localfilter = { type: 'tvslst', in: true, join: '', lst: [] as any[] }
+	if (filters.length > 1) localfilter.join = 'and' // For now hardcoding join as 'and' if number of filter terms > 1. Will later implement more comprehensive logic
```
compbiolover: If focusing only on AND logic for now might be good to include some logic in the validation function to ensure user queries with OR don't silently fail or silently get changed to AND. Could cause unexpected behavior

server/routes/termdb.chat.ts
```diff
@@ -56,122 +42,454 @@ function init({ genomes }) {
...
+				description: description[0].description,
+				values: values,
+				term_type: term_type
+			}
+			const stringified_db = parse_db_rows(db_row)
+			rag_docs.push(stringified_db)
+			db_rows.push(db_row)
+		}
+	})
+	return rag_docs
```
compbiolover: Before returning rag_docs shouldn't the db be closed?

server/routes/termdb.chat.ts
```diff
@@ -56,122 +42,454 @@ function init({ genomes }) {
...
+	const db = new Database(dataset_db)
+	// Query the database
+	const desc_rows = db.prepare('SELECT * from termhtmldef').all()
+
+	const description_map: any = []
+	// Process the retrieved rows
+	desc_rows.forEach((row: any) => {
+		const name: string = row.id
+		const jsonhtml = JSON.parse(row.jsonhtml)
+		const description: string = jsonhtml.description[0].value
```
compbiolover: Wouldn't we want to do  something like `const description: string = jsonhtml?.description?.[0]?.value ?? ''` in case any of the description fields are missing or empty?

server/routes/termdb.chat.ts
```diff
@@ -56,122 +42,454 @@ function init({ genomes }) {
...
-			//mayLog('ai_output_json:', ai_output_json)
-			res.send(ai_output_json as ChatResponse)
-		} catch (e: any) {
-			if (e.stack) console.log(e.stack)
-			res.send({ error: e?.message || e })
+async function parse_geneset_db(genedb: string) {
+	const db = new Database(genedb)
+	// Query the database
+	const desc_rows = db.prepare('SELECT * from codingGenes').all()
+	let genes_list: string[] = []
```
compbiolover: Wouldn't it be faster to use a set here? Something kind of like this? 
`    const genes = new Set<string>()
    for (const row of rows) {
      if (row.name) {
        genes.add(row.name.toLowerCase())
      }
    }`

shared/types/src/routes/termdb.chat.ts
```diff
@@ -34,3 +34,34 @@ export const ChatPayload: RoutePayload = {
...
+
+export type FilterTerm = CategoricalFilterTerm | NumericFilterTerm
+
+export type CategoricalFilterTerm = {
+	term: string
+	category: string
+}
+
+export type NumericFilterTerm = {
+	term: string
```
compbiolover: Did these values end up being aligned with the typescript definitions used throughout pp?

server/routes/termdb.chat.ts
robinpaul85: I would personally prefer it to fail which will make it easier to catch the problem. If the description is blank and the chatbot starts to hallucinate its much harder to debug why its hallucinating (due to lack of description). In my future PRs, I plan to add tests for each db which should catch any in compatibilities with the code and db before we deploy to production.

server/routes/termdb.chat.ts
robinpaul85: Done

server/routes/termdb.chat.ts
robinpaul85: I think it would be better to implement the and/or logic for filters and add appropriate training data. Right now it just coerces multiple filter terms into `and` .

PR #4198, opened 2026-02-16
add vafs to single sample json

josephpowi:
add vafs to single sample json

### DIFF
--- a/server/test/tp/files/hg38/TermdbTest/mutationpersample/3318
+++ b/server/test/tp/files/hg38/TermdbTest/mutationpersample/3318
@@ -14,7 +14,7 @@
   "vafs": [
    {
     "altCount": 15,
-    "name": "tumor DNA WGS",
+    "id": "tumor_DNA_WGS",
     "refCount": 13
    }
   ]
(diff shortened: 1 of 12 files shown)


server/test/tp/files/hg38/TermdbTest/mutationpersample/3318
```diff
@@ -12,7 +11,13 @@
   "origin": "germline",
   "pos": 7675088,
   "ref": "C",
-  "refCount": 13
+  "vafs": [
+   {
+    "altCount": 15,
+    "name": "tumor DNA WGS",
```
xzhou82: discussed with Jian and will change `name` into `id` with FORMAT field as value in accordance with Gav's prior work

xzhou82:
1. introduce `queries.snvindel.mafFilter` to termdbtest. we missed it in Gav's prior PR @gavrielm . make sure all tests pass, and may add some test coverage for mafFilter now it's in termdbtest
2. grin2 shares mtationpersample files with disco. to avoid breaking grin2, in grin2 ui, update maf filter part based on termdbConfig.queries.snvindel.mafFilter. need to decide if/how to implement logic and/or as in geneVariant tvs, or quick fix to hardcode only to use `mafFilter[0]`, and solve corresponding snvindel filtering in routes/grin2.ts;
3. if turns out too hard, simply disable maf filter in grin2 ui, so disco fix can be merged. fix up grin2 in next branch

xzhou82 (approved):

PR #4219, opened 2026-02-19
include WT/Not Tested in for tw.term.genes data requests

siosonel:
# Description

Fixes https://gdc-ctds.atlassian.net/browse/SV-2753. Pull and test with `sjpp/master`.

## Checklist

### DIFF
--- a/server/src/termdb.matrix.js
+++ b/server/src/termdb.matrix.js
@@ -130,7 +130,7 @@ async function getSampleData(q, ds, onlyChildren = false) {
 		} else {
 			if (!q.ds.mayGetGeneVariantData) throw 'not supported by dataset: geneVariant'
 			const maxGenesPerUser = ds.cohort.termdb?.maxGenesPerUser || 1000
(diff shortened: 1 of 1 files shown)


client/termdb/TermdbVocab.js
```diff
@@ -807,7 +807,8 @@ export class TermdbVocab extends Vocab {
 					// one request per term
 					terms: copies.map(this.getTwMinCopy),
 					filter,
-					embedder: window.location.hostname
+					embedder: window.location.hostname,
+					isSummary: opts.isSummary
```
xzhou82: what's the source of `isSummary:true`? 
or rather than this flag, how about adding `data.isMatrix=true` at matrix.data.js `const data = await this.app.vocabApi.getAnnotatedSampleData(opts, _data)`?

gavrielm:
@siosonel this will not fix the linked jira ticket. I will make a separate PR.

client/termdb/TermdbVocab.js
xzhou82: i see. feels tricky. better add notes

PR #4253, opened 2026-02-25
Tentative implementation for testing prompts in chatbot

robinpaul85:
# Description
Implementation for automatically testing test prompts.
All dataset tests can be run using this [URL](http://localhost:3000/testchat).
Can optionally also run select datasets using this [URL](http://localhost:3000/testchat?dslabel=TermdbTest)

## Checklist

### DIFF
--- a/server/src/test/routes/testchat.ts
+++ b/server/src/test/routes/testchat.ts
@@ -9,11 +9,12 @@ import assert from 'node:assert/strict'
 process.removeAllListeners('warning')
 
 export default function setRoutes(app, basepath, genomes) {
-	app.get(basepath + '/testchat', async (req, _res) => {
+	app.get(basepath + '/testchat', async (req, res) => {
 		// URL parameters:
 		// ?dataset=<label> - restrict testing to a specific dataset (e.g. ?dataset=TermdbTest)
 		const datasetFilter = req.query.dslabel as string | undefined
 		console.log('test chat page' + (datasetFilter ? ` (dslabel filter: ${datasetFilter})` : ''))
+		const results: { dataset: string; num_errors: number }[] = []
 		for (const genome of Object.values(genomes)) {
 			for (const ds of Object.values((genome as any).datasets)) {
 				if ((ds as any)?.queries?.chat) {
@@ -32,9 +33,11 @@ export default function setRoutes(app, basepath, genomes) {
 							'Tests complete for ' + label + '. Number of failed prompts: ' + num_errors
 						) // Show in red if any of the tests failed
 					}
+					results.push({ dataset: label, num_errors: num_errors })
 				}
 			}
 		}
+		res.send(results)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


siosonel:
@robinpaul85 @compbiolover @xzhou82  In order to minimize having test/spec code from prod code, and instead of importing test code in `server/src/app.ts`, it's better to do the following:
1. Create a test route file like `server/src/test/routes/chat.ts` or similar, so there will be a `/chat-tests` server endpoint. The route handler will run your tests.
2. Edit `server/src/servercoonfig.js` to add `chat.ts` to `testRouteSetters` array.
3. To run the tests, just load `http://localhost:3000/chat-tests`. It can reply with ok if passing, also emit the test results in the server logs.

A different way to do this is have code similar to `server/src/test/xfetch.integration.spec.js`, which calls `/ky-retry-test` and `//external-API-test?dslabel=GDC` to run tests. So the `/chat-tests` route handler doesn't have to run test, just has to respond, and the integration spec code will run the `tape` tests.

server/src/test/routes/testchat.ts
xzhou82: use of detailed if-else is counter-intuitive. just strickly require `test_result` to be deepEqual to expected result object, and get rid of all these if-else

server/src/test/routes/testchat.ts
```diff
@@ -9,23 +9,28 @@ import assert from 'node:assert/strict'
 process.removeAllListeners('warning')
 
 export default function setRoutes(app, basepath, genomes) {
-	app.get(basepath + '/testchat', async () => {
-		// (req.res) not currently used
-		console.log('test chat page')
+	app.get(basepath + '/testchat', async (req, _res) => {
+		// URL parameters:
+		// ?dataset=<label> - restrict testing to a specific dataset (e.g. ?dataset=all-pharmacotyping)
+		const datasetFilter = req.query.dataset as string | undefined
```
siosonel: In the vast majority of routes, this is called `req.query.dslabel`, it may help to be consistent for app middleware support

siosonel: What I mean is, the request payload should use `dslabel`, like `/testchat?dslabel=TermdbTest`, not `?dataset=TermdbTest`. There are some dslabel checks that's done in app.middleware.js, so it helps to be consistent with the parameter name.

robinpaul85: I have also added weblinks to the PR description and added the URL in the sjpp PR

server/dataset/ai/termdb.test.json
```diff
@@ -1,808 +1,1893 @@
...
-			}
-		}
-	],
-	"prebuiltPlots": [
-		{
-			"name": "TermdbTest TSNE",
-			"chartType": "sampleScatter"
-		}
-	]
+  "hasGeneExpression": true,
```
xzhou82: delete these flags and no longer duplicating info, since now runtime ds obj is available

xzhou82 (approved):

PR #4435, opened 2026-04-20
DA: Render volcano plots server-side as PNG + interactive top-N overlay

compbiolover:
# Description
The volcano plot previously rendered every dot (10k–30k+ per analysis) as an individual SVG <circle>, which put a lot of memory pressure on the client especially for DNA methylation analysis (13k+ promoters). This PR offloads full-scatter rendering to a new Rust binary (`volcano.rs`) that creates a PNG server-side, and narrows the client overlay to just the top-N most-significant dots — which are the only ones that need hover/click interactivity. The number of dots that get hover/click interactivity is sent by the client

### New

- `rust/src/volcano.rs` — single pass over input rows: classifies vs thresholds, draws stroke-only rings into a plotters BitMapBackend, encodes PNG via tiny-skia, and emits { png, plot_extent, dots, total_rows } on stdout. Sorts significant rows by p-value and truncates to `max_interactive_dots`. Exposes `min_nonzero_p` in `plot_extent` so the client can cap p=0 rows at the same y-coordinate the PNG drew them at.
- `server/src/renderVolcano.ts` — thin helper wrapping `run_rust('volcano', …)`; logs elapsed time per render via `mayLog` + `formatElapsedTime`.
- Shared types (shared/types/src/routes/*)
- New `VolcanoData<T>` generic bundling { dots, volcanoPng, plotExtent, totalRows }.
- Every response type nests its volcano payload under data: `VolcanoData<RouteEntry>` alongside route-specific metadata (sample sizes, method, etc.).
- New `VolcanoRenderRequest` (significance thresholds, pixel dimensions, dot colors, max interactive dots); required on all three request types — server always runs the renderer.
- Removed the legacy flat row-array data field from `DEResponse`, `DiffMethResponse`, `HasDataResponse`.

### Server

- Routes simplified: each one runs its existing stats pipeline, then pipes rows through `renderVolcano` and nests the result under data.
- Preserves `preAnalysis` short-circuit on `termdb.diffMeth`.

### Client

- `VolcanoModel.getVolcanoRender()` sends `significanceThresholds`, `pixelWidth/Height`, case/control group colors (up/down), `dotRadius`, and `maxInteractiveDots`. All color fields are normalized to #rrggbb via `d3-color` before sending so the Rust renderer's hex-only parser doesn't fall back on CSS names like 'red'.
- `VolcanoViewModel` adopts the server's `plotExtent` + `minNonZeroPValue` so overlay circles land exactly on their PNG counterparts; all rows in `data.dots` are significant by definition.
- `VolcanoPlotView.renderPlot()` inserts an SVG <image> at the plot rect and the existing `renderDataPoints` overlays interactive <circle>s on top, one per dot.
- Hover highlight uses a clone-on-enter pattern (`DataPointMouseEvents`) so the original circle never moves in the DOM — fixes stuck-highlight glitches in dense regions.
### New setting
`maxInteractiveDots` (default 5000) with a number input in `VolcanoControlInputs`. Caps the overlay; the PNG always shows every dot.

Applies to all three differential-analysis routes: `termdb/DE`, `termdb/diffMeth`, and `termdb/singlecellDEgenes`.

## Checklist


client/plots/volcano/VolcanoControlInputs.ts
```diff
@@ -64,6 +64,17 @@ export class VolcanoControlInputs {
...
 			},
+			{
+				label: 'Max interactive dots',
+				type: 'number',
+				chartType: 'volcano',
+				settingsKey: 'maxInteractiveDots',
+				title:
+					'Cap on the number of top-significant points the server returns as interactive overlay circles. The PNG still shows every dot.',
+				min: 0,
+				max: 100000,
```
creilly8: 100,000 seems excessive. Maybe 10,000 or 20,000 should be the hard limit.

compbiolover: Thanks, I lowered it to 20,000. I know the promoter example has nearly 14,000 significant promoters

xzhou82 (approved):

PR #4439, opened 2026-04-21
fix the sample hierarchy issue in lollipop

josephpowi:
fix the sample hierarchy issue in lollipop.

integration test error:
          not ok 1234 subtrack has fewer number of mutations than main track


xzhou82:
fixed https://github.com/stjude/proteinpaint/pull/4452

PR #4441, opened 2026-04-22
GSEA: Reworked GSEA to use cache id

compbiolover:
# Description
Fix GSEA after recent volcano changes by introducing a server-side DA cache.

The volcano plot was changed to server-render the scatter PNG and return only the top-significant interactive dots (capped at `maxInteractiveDots`). GSEA, which runs downstream of the DE analysis, was still expecting the full gene list on the client — so it either crashed (`response.data` is no longer an array) or silently ran on a truncated subset.

This PR mirrors the GRIN2 cache pattern: the DE route now writes the full `{gene_name, fold_change}` table to disk and returns a `cacheId`; GSEA consumes the cacheId instead of re-transmitting gene arrays.

## Changes

- **New `daAnalysis` cache subdir** registered in `CacheManager` (60-day TTL, `.tsv`). Named "da" rather than "de" because we now differential analysis on more than just gene expression
- **DE route** ([server/routes/termdb.DE.ts](proteinpaint/server/routes/termdb.DE.ts)) writes a TSV cache file (`da_<32hex>.tsv`) after edgeR, wilcoxon, and limma runs, and attaches `cacheId` to `VolcanoData`.
- **GSEA route** ([server/routes/genesetEnrichment.ts](proteinpaint/server/routes/genesetEnrichment.ts)) reads genes + fold_change from the cache when `cacheId` is present. CacheId is regex-gated (`/^da_[0-9a-f]{32}$/`) before it enters `path.join` to block traversal. Also adds a `fetchDE: true` short-circuit so the client can lazy-load the ranked gene list for the cerno detail plot.
- **GSEA client** ([client/plots/gsea.js](proteinpaint/client/plots/gsea.js)) stores `cacheId` + `genes_length` in `gsea_params` instead of raw arrays; POST body sends only `cacheId`. The cerno enrichment-curve detail plot lazy-fetches the ranked list once and caches it in memory (not plot state).
- **Types** ([shared/types/src/routes/termdb.DE.ts](proteinpaint/shared/types/src/routes/termdb.DE.ts), [shared/types/src/routes/genesetEnrichment.ts](proteinpaint/shared/types/src/routes/genesetEnrichment.ts)) updated: `VolcanoData.cacheId?`, `GenesetEnrichmentRequest.{cacheId?, fetchDE?}`, and `genes`/`fold_change` are now optional.

## Backward compatibility

The GSEA request still accepts inline `genes` + `fold_change` when `cacheId` is absent, so older mass-session files continue to work — they'll just re-run DE on load (same behavior as before). Sessions saved after this change carry a `cacheId` that may expire after 60 days; if so, the user reopens the plot to regenerate. Matches GRIN2 behavior.

## Test plan

- [x] Open a gene-expression volcano, launch GSEA with **blitzgsea** — table populates, header shows the full gene count (not `maxInteractiveDots`).
- [x] Same flow with **cerno** — click a pathway row, confirm the enrichment-curve detail plot renders.
- [x] `ls <cachedir>/daAnalysis/` shows a `da_<32hex>.tsv` with one row per DE gene; row count equals `totalRows` in the volcano response.
- [x] Delete the TSV then click a pathway — produces a clear error, not a silent empty result.

## Checklist

### DIFF
--- a/client/plots/gsea.js
+++ b/client/plots/gsea.js
@@ -1,5 +1,5 @@
 import * as d3axis from 'd3-axis'
-import { Menu, renderTable, table2col, axisstyle } from '#dom'
+import { Menu, renderTable, table2col, axisstyle, sayerror } from '#dom'
 import { dofetch3 } from '#common/dofetch'
 import { controlsInit } from './controls'
 import { getCompInit, copyMerge } from '#rx'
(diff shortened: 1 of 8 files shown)


server/routes/termdb.DE.ts
```diff
@@ -13,6 +14,15 @@ import { get_header_txt } from '#src/utils.js'
 import { formatElapsedTime } from '#shared'
 import { renderVolcano } from '../src/renderVolcano.ts'
 
+async function writeDaCache(geneData: GeneDEEntry[]): Promise<string> {
+	const cacheId = `da_${crypto.randomBytes(16).toString('hex')}`
+	const file = path.join(serverconfig.cachedir, 'daAnalysis', `${cacheId}.tsv`)
+	const lines = ['gene_name\tfold_change']
+	for (const g of geneData) lines.push(`${g.gene_name}\t${g.fold_change}`)
+	await fs.promises.writeFile(file, lines.join('\n'))
+	return cacheId
```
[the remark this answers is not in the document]
compbiolover: Is this a concern? Since we have the cache monitor starting at startup I assumed it was safe to assume it would always be running

client/plots/gsea.js
```diff
@@ -371,19 +368,43 @@ add:
...
+			} else if (self.settings.gsea_method == 'cerno') {
+				if (!self.rankedDE && self.config.gsea_params.cacheId) {
+					const deResp = await dofetch3('genesetEnrichment', {
+						body: {
+							genome: self.config.gsea_params.genome,
+							cacheId: self.config.gsea_params.cacheId,
+							fetchDE: true,
+							geneSetGroup: '-',
+							filter_non_coding_genes: false,
+							method: 'cerno'
```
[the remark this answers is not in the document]
compbiolover: Resolved by making it optional since only blitzgsea requires it

xzhou82 (approved):

shared/types/src/routes/termdb.DE.ts
```diff
@@ -85,6 +85,10 @@ export type VolcanoData<T extends DataEntry> = {
 	/** Rows that passed significance thresholds, before any maxInteractiveDots
 	 * truncation. Use this (not dots.length) for "% significant" stats. */
 	totalSignificantRows: number
+	/** Server-side cache ID for the full DE result (all rows, not just dots).
+	 * Downstream tools (e.g. GSEA) pass this back to the server instead of
+	 * re-transmitting the gene + fold_change arrays. */
+	cacheId?: string
```
siosonel, line 91: @compbiolover Will the caching work in a server farm? not sure if the comment above requires the server to already know or have access to cached results, or is there a way for a server to recompute or for the client to retry a request by retransmitting the cache "input" data

xzhou82: yes it will be an issue as it is. but we will fix it to enable recomputing when cache file is missing, as we did for bam caching. thanks!

compbiolover: Thanks @siosonel. I'll make the fix @xzhou82 mentioned

PR #4496, opened 2026-05-11
Lowering memory usage for top VE gene

ojmel:
Lowering memory usage for top variably expresssed genes by reading in genes from the hdf matrix one at a time

# Description

## Checklist

### DIFF
--- a/python/src/topVEgene.py
+++ b/python/src/topVEgene.py
@@ -30,24 +30,26 @@ def create_gene_variance_list(
         all_samples = hdf_data["samples"].asstr()[:]
         matrix = hdf_data["matrix"]
 
-        sample_to_idx = {sample: index for index, sample in enumerate(all_samples) if sample in sample_list}
-        missing = [s for s in sample_list if s not in sample_to_idx]
-        if missing:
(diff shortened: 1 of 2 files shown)


python/src/topVEgene.py
```diff
@@ -12,62 +12,71 @@
...
+    max_genes: int
+) -> list[str]:
     sample_list = [str(s) for s in sample_list]
     with h5py.File(filename, "r") as hdf_data:
         gene_names = hdf_data["item"].asstr()[:]
         all_samples = hdf_data["samples"].asstr()[:]
-        matrix = np.array(hdf_data["matrix"])
+        matrix = hdf_data["matrix"]
+
+        sample_indexes = [index for index, sample in enumerate(all_samples) if sample in sample_list]
```
josephpowi: It might be worth  checking  sample against a set instead of sample_list. It is minor for typical sizes, but could become real for large cohorts.

ojmel: Check the new lookup and tell me what you think

xzhou82 (approved):

PR #4501, opened 2026-05-12
Fix single-cell volcano: hover tooltips, group labels, GSEA tab, high res volcano render

compbiolover:
# Description

Four fixes total. Three to the single-cell cell-type (SCCT) tab of differential analysis:

- **Hover tooltips on significant dots.** `attachInteractions` was short-circuiting for `SINGLECELL_CELLTYPE`, so the cover-rect / quadtree / tooltip pipeline never bound. Dropped the early-return; the existing GE tooltip path renders correctly for SCCT data (uses `gene_name`, `fold_change`, `original_p_value`, `adjusted_p_value`) and `getActionMenuOpts` already returns `[]` for SCCT so the single-hit action menu shows just the info panel with no buttons.
- **Group labels at the top of the plot.** `setTermInfo` ignored SCCT and returned `undefined`. Added an SCCT branch: `Cluster {categoryName}` on the right (positive fold-change = up in cluster) and `Not in Cluster {categoryName}` on the left. No `sample_size1/2` exists for SCCT, so the labels intentionally omit the trailing `(n)`.
- **GSEA tab now runs.** GSEA `init()` only knew the `cacheId` path (DE/DM) and the `dapParams` path (proteome); SCCT has neither, so init threw "No DE cacheId returned from volcano model" and the tab silently died. Added an SCCT branch that fetches `termdb/singlecellDEgenes` **without** `volcanoRender` (so the route returns the full ranked gene array, not the threshold-passing `dots` subset) and stores `{ genes, fold_change, genes_length, genome }` inline on `gsea_params`. `render_gsea`'s existing inline-genes path (previously only reached by the legacy `singleCellPlot.renderGSEA`) handles it.
- **High-resolution volcano render** Previously the render would be pixelated on higher dpi displays. Now uses device pixel ratio in a manner like manhattan plot to render the plot in a high res way

## Test plan

- [x] In GDC, open a single-cell sample → Differential expression → pick a cluster:
  - [x] Hovering a significant dot shows the gene tooltip with fold-change + p-values
  - [x] Plot header shows `Not in Cluster N` (left) and `Cluster N` (right)
  - [x] Clicking the **Gene Set Enrichment Analysis** tab fires `termdb/singlecellDEgenes` and renders the pathway dropdown (no console error)
  - [x] Picking a pathway returns enrichment results
- [x] Confirm the GE volcano is unchanged: hover, term-info labels with sample sizes, and GSEA via the cacheId path all still work.
- [ ] Confirm the DAP volcano is unchanged.

## Checklist

### DIFF
--- a/client/plots/volcano/viewModel/VolcanoViewModel.ts
+++ b/client/plots/volcano/viewModel/VolcanoViewModel.ts
@@ -230,17 +230,20 @@ export class VolcanoViewModel {
 		}
 
 		if (this.termType == SINGLECELL_CELLTYPE) {
-			// Positive fold-change = up in the selected cluster, so the cluster
+			// The "group name" prefix is the SC config's `DEgenes.termId`
+			// (e.g. "Cluster" for GDC) — it's the column whose value
+			// `categoryName` identifies, so the dataset already names it.
+			// Positive fold-change = up in the selected group, so the group
 			// label goes on the right and the complement on the left.
-			const clusterLabel = `Cluster ${this.config.categoryName}`
+			const groupLabel = `${this.config.termId} ${this.config.categoryName}`
 			return {
 				y: plotDim.top.y + 10,
 				first: {
-					label: getLabel(`Not in ${clusterLabel}`),
+					label: getLabel(`Not in ${groupLabel}`),
 					x: 0
 				},
 				second: {
-					label: getLabel(clusterLabel),
+					label: getLabel(groupLabel),
 					x: this.settings.width
 				}
 			}
--- a/client/plots/volcano/viewModel/VolcanoViewModel.ts
+++ b/client/plots/volcano/viewModel/VolcanoViewModel.ts
@@ -203,7 +203,13 @@ export class VolcanoViewModel {
 		// caseColor: string,
 		// controlColor: string
 	) {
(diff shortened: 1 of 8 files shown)


client/plots/volcano/viewModel/VolcanoViewModel.ts
```diff
@@ -223,6 +229,23 @@ export class VolcanoViewModel {
 			}
 		}
 
+		if (this.termType == SINGLECELL_CELLTYPE) {
+			// Positive fold-change = up in the selected cluster, so the cluster
+			// label goes on the right and the complement on the left.
+			const clusterLabel = `Cluster ${this.config.categoryName}`
```
xzhou82: these group names should be defined by sc, supplied to da, but not hardcoded as "Cluster N" here

client/plots/volcano/viewModel/VolcanoViewModel.ts
```diff
@@ -223,6 +230,27 @@ export class VolcanoViewModel {
 			}
 		}
 
+		if (this.termType == SINGLECELL_CELLTYPE) {
```
xzhou82, line 233: would like to know if possible for SC and DAP to supply `this.config.samplelst.groups[]` as in line 259 to simplify

xzhou82 (approved):

PR #4537, opened 2026-05-19
Bound cacheOrRecompute pool, auto-register subdirs, expose busy state

compbiolover:
# Description

- **Bounded per-subdir compute pool.** `cacheOrRecompute`'s `pending` map was previously unbounded — a burst of distinct DE/DM/GSEA/GRIN2/topVE requests could spawn unbounded R/Python work and overload the server before disk cache populated. Each subdir now has a `maxPending: 5` cap (single source of truth in `utils/cacheOrRecompute.ts`). The 6th distinct request fails fast with HTTP 429 + `code: 'CACHE_BUSY'` + `"Cache compute pool is full. Please try again shortly."`. Same-key attachers still dedup for free — they do not consume a slot.
- **Single source of truth for cache subdirs.** `cacheJobPolicies` in `utils/cacheOrRecompute.ts` now carries each subdir's name and `maxPending`. `CacheManager` imports the object and auto-registers every entry into its eviction sweep. Adding a new analysis type is a one-line edit.
- **Timing instrumentation for cache reads/writes.** `writeJsonCache` and `tryReadJson` now `mayLog` per-call duration + payload size via `formatElapsedTime` and `fileSize`, gated by `serverconfig.debugmode`. Example: `cacheOrRecompute read de/abc….json (6.24 MB) in 32ms`.
- **GRIN2 outer limit aligned to 5** (`GRIN2_CONCURRENCY_LIMIT`) so both backpressure layers agree.

## Test plan

- [x] `npx tsx ./src/utils/test/cacheOrRecompute.unit.spec.ts` from `proteinpaint/server/` — all 35 tests pass (covers pool gating, busy rejection, retry-after-drain, dedup-doesn't-burn-a-slot, timing log shape, and silent-when-debugmode-off).
- [x] `npx tsx ./src/test/CacheManager.unit.spec.ts` from `proteinpaint/server/` — covers auto-registration of every `CACHE_OR_RECOMPUTE_SUBDIRS` entry plus the disable-attempt-throws guard.
- [x] `npx tsc --noEmit` clean on both server and client.
- [x] Manual UI: temporarily set `de.maxPending: 0` in `utils/types.ts`, open a volcano plot — confirm the window popup "Cache compute pool is full. Please try again shortly." bar appears with a plain Retry button. Click Retry — confirms it re-fires the same request (still busies at cap 0). Revert to `maxPending: 5` and confirm the plot loads normally.
- [x] Manual curl: `maxPending: 1`, fire 5 parallel `/termdb/DE` requests with distinct `min_count` values — confirm 1 returns 200 and the rest return HTTP 429 with `{"status":"error","code":"CACHE_BUSY",...}`.

## Checklist


server/src/utils/types.ts
```diff
@@ -1,8 +1,21 @@
...
+ * registers every entry automatically with the eviction policy below;
+ * `maxSize` falls back to CacheManager's 5 GB per-subdir default when
+ * omitted. To add a new analysis type, append a new entry here — no
+ * other file needs to be edited. */
+export const CACHE_OR_RECOMPUTE_SUBDIRS = {
+	de: { maxAge: day * 60, skipMs: halfDay, maxPending: 5 },
+	dm: { maxAge: day * 60, skipMs: halfDay, maxPending: 5 },
+	gsea: { maxAge: day * 60, skipMs: halfDay, maxPending: 5 },
+	grin2: { maxAge: day * 60, skipMs: halfDay, maxPending: 5 },
+	topve: { maxAge: day * 60, skipMs: halfDay, maxPending: 5 }
```
siosonel: If we use this for GDC cohort-maf and bam-related tools, I'm pretty sure 60 days is too long. For BAM slices, I remember we only cache for 5 minutes at most. So, there should be a way for a dataset to override these default values based on embedder portal requirements. Search for `serverconfig.features.gdc` to examples of override options that may need to be supported later.

compbiolover: Thanks, I added a note about this so in a future PR I (or whoever is making the change) will be aware. Also `CacheManager's` constructor at [app.ts:29-33] includes `serverconfig.features?.cacheMonitor` into its options, and the constructor at [CacheManager.ts:155] merges per-subdir overrides on top of the defaults from `CACHE_OR_RECOMPUTE_SUBDIRS`. (e.g. ```    "cacheMonitor": {
      "subdirs": {
        "gdcBamSlice": { "maxAge": 300000, "skipMs": 60000, "maxSize": 500000000 },
        "gdcMaf":       { "maxAge": 300000, "skipMs": 60000 }
      }
    }
    ```
  The only flag not supported is `maxPending` because ist is read by `cacheOrRecompute.ts` straight from the static `CACHE_OR_RECOMPUTE_SUBDIRS` import — not from `CacheManager.subdirs`. I could extend it so that this flag is also covered if we decide to deploy this to GDC

client/gdc/grin2.ts
```diff
@@ -2214,7 +2214,12 @@ async function getFilesAndShowTable(obj) {
 					.text('No significant genes found in the analysis.')
 			}
 		} catch (e: any) {
-			sayerror(obj.errDiv, e.message || e)
+			if (e?.code === 'CACHE_BUSY') {
+				sayinfo(obj.errDiv, e.message || e, {
```
xzhou82: on e.code=CACHE_BUSY, let's consider to simply `window.alert(e.message)`

server/src/utils/types.ts
xzhou82: this script is about types. how about moving these constants that are not types out into cacheOrRecompute.ts

server/src/utils/cacheOrRecompute.ts
xzhou82, line 100: need to validate `if(!CACHE_OR_RECOMPUTE_SUBDIRS[cacheSubdir]) throw ..` so line 101 doesn't break

server/src/utils/types.ts
```diff
@@ -1,8 +1,21 @@
...
+const day = 1000 * 60 * 60 * 24
+const halfDay = day / 2
+
+/** Subdirs of serverconfig.cachedir that the cacheOrRecompute module
+ * writes JSON cache files to. CacheManager imports this object and
+ * registers every entry automatically with the eviction policy below;
+ * `maxSize` falls back to CacheManager's 5 GB per-subdir default when
+ * omitted. To add a new analysis type, append a new entry here — no
+ * other file needs to be edited. */
+export const CACHE_OR_RECOMPUTE_SUBDIRS = {
```
xzhou82: can rename CACHE_OR_RECOMPUTE_SUBDIRS to `cacheJobPolicies` to be intuitive. maxAge and skipMs are not used by cacheorrecompute thus should be removed

client/gdc/grin2.ts
siosonel: Should not use `alert()`, it may cause automated tests to hang. The `sayInfo()` helper should use a standard dialog element as a modal.

server/src/utils/types.ts
compbiolover: Moved the constant to `cacheOrRecompute.ts`

server/src/utils/cacheOrRecompute.ts
compbiolover: Validated now

client/gdc/grin2.ts
compbiolover: Changed to `window.confirm` so that when the user clicks ok they submit the same request with the same inputs while clicking cancel simply exits. This makes it easy for the user to for example submit their request again in situations like DE or DM where we don't have a submit/run button ui like in GRIN2.

server/src/utils/types.ts
compbiolover: Renamed to `cacheJobPolicies` and removed maxAge and skipMs and let them be handled by `CacheManager.ts`

xzhou82 (approved):

PR #4543, opened 2026-05-22
Fix termcollection visibility

vivekavuthu:
This pull request improves the handling of `termCollection` visibility in the `divideTerms` function, ensuring that access control is enforced correctly for collections of terms. Previously, `termCollection` terms could be incorrectly dropped for users with restricted roles because they lack a scalar `id`. Now, visibility is determined by checking that all member term IDs are visible to the requesting role. Comprehensive unit tests are added to validate this behavior.

Enhancements to termCollection visibility logic:

* Updated `divideTerms` to determine visibility of `termCollection` terms by requiring all member IDs to be visible to the requesting role, rather than relying on a non-existent scalar `id`. If any member is not visible, the collection is dropped. 
* Added logic to fall back to `termlst[].id` when `termIds` is absent in a `termCollection`, and to drop collections with no resolvable member IDs ("fail-closed").

Unit test coverage:

* Added multiple tests to verify that `termCollection` terms are only visible when all member IDs are allowed, that the fallback to `termlst[].id` works, that empty collections are dropped, and that datasets without an `isTermVisible` hook are unaffected. 
* This pull request introduces significant improvements to the barchart plotting logic, especially around the normalization and ordering of categorical and numeric series, as well as enhancements to term visibility and error handling. The changes improve robustness in edge cases, ensure more predictable bar ordering, and update test coverage accordingly.

**Barchart Series Normalization and Ordering:**

* Added a `normalizeRefOrder` function in `barchart.js` to ensure that categorical and numeric series IDs and labels are consistently normalized, improving the stability and predictability of series ordering and hidden value handling. This normalization is now applied when updating chart settings and when determining which series are excluded or hidden. 
* Refined the logic for determining the order of bars and overlays in the chart by introducing declared and numeric value order detection, and by making the bar and overlay sorters more robust to key/label mismatches.

**Error Handling and Global Context:**

* Updated error logging in `getPlotConfig` to use a more robust global context detection (`Function('return this')()`) for environments where `window` may not be available, preventing runtime errors in non-browser contexts. 

**Test Suite Updates:**

* Revised integration tests to accommodate the new normalization logic, including more flexible assertions for bar counts and order, and avoiding legend click interactions that could be affected by normalization changes. Added a new test for series order when no hidden values are present. 

**Server-side Series Ordering:**

* Improved the `getOrderedLabels` function on the server to better handle terms with partial or missing `order` properties, ensuring that keys with explicit order are prioritized and sorted correctly.

**Term Collection Visibility:**

* Enhanced the `divideTerms` function to properly determine the visibility of term collections by checking the visibility of all member terms, rather than relying on the collection itself having an `id`. # Description

## Checklist

### DIFF
--- a/server/src/termdb.matrix.js
+++ b/server/src/termdb.matrix.js
@@ -468,7 +468,16 @@ export function divideTerms(q, ds) {
 		geneVariantTws = [],
 		nonDict = []
 	const isTermCollectionVisible = tw => {
-		const memberIds = tw.term?.termIds || tw.term?.termlst?.map(t => t?.id).filter(Boolean) || []
+		// Fail-closed against malformed input: a request that arrives with termIds or
+		// termlst as a non-array (string, object, etc.) is treated as having no resolvable
+		// members. Guarding here keeps a malformed payload from crashing the whole request
+		// with a TypeError when .every()/.map() is called on a non-array.
+		const raw = Array.isArray(tw.term?.termIds)
+			? tw.term.termIds
+			: Array.isArray(tw.term?.termlst)
+			? tw.term.termlst.map(t => t?.id)
+			: []
+		const memberIds = [...new Set(raw.filter(id => typeof id === 'string' && id))]
 		if (!memberIds.length) return false
 		return memberIds.every(id => ds.cohort.termdb.isTermVisible(q.__protected__, id))
 	}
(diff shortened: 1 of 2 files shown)


xzhou82 (approved):
please complete below in next pr
1. change to `isTermVisible(_, term)` that accepts term obj but not id. this generalize it to cover non-dict term in future
2. for termCollection payload, `tw.term.termlst` is always present for both dict and non-dict collections, and should be used. `tw.term.termIds[]` should be phased out

PR #4627, opened 2026-06-22
gdc cohort maf: add user agent header to gdcmaf.rs

josephpowi:


rust/src/gdcmaf.rs
```diff
@@ -201,24 +201,29 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         )) as Box<dyn std::error::Error>);
     };
 
+    let client = reqwest::Client::builder()
+        .user_agent("MyCustomApp/1.0")
```
siosonel: @josephpowi  This should use the `arg,headers` key-values that was supplied in line 77 of `gdc.mafBuild.ts`, so somehow loop through those, which would be equivalent to adding these headers
```rust
        .header("X-Forwarded-Agent",  ... ) // Custom header key and value
        .header("X-Forwarded-For",  ...) // 
```

rust/src/gdcmaf.rs
```diff
@@ -206,6 +206,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         let req_headers_clone = req_headers.clone();
         async move {
             let client = reqwest::Client::builder()
+                .user_agent("MyCustomApp/1.0")
                 .timeout(Duration::from_secs(60)) // 60-second timeout per request
                 .connect_timeout(Duration::from_secs(15))
                 .default_headers(req_headers_clone.clone())
```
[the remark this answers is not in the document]
josephpowi: fixed

rust/src/gdcmaf.rs
josephpowi: deleted `.user_agent` and add header in a look

josephpowi (commented):
deleted `.user_agent` and add header in a loop

josephpowi (commented):
1. disable idle connection reuse
2. lower the per-request timeout to 30s
3. lower the simultaneous conncetions from 20 to 10
4. Retry up to 3 times if the download times out.

siosonel (approved):

siosonel (approved):

PR #4634, opened 2026-06-23
GRIN2: CNV type selection

compbiolover:
# Description
GRIN2 previously had no way to choose which CNV file type to analyze. For datasets like **GDC**, a single case can have multiple CNV file types (e.g. *Masked Copy Number Segment* from SNP array with segment-mean values, vs *Allele-specific Copy Number Segment* with copy-number values), and the getter silently picked one default per case — mixing value semantics under a single set of thresholds, and classifying allele-specific (baseline-2) segments against log2-ratio thresholds.

This PR lets the user pick the CNV type via radio buttons, loads only the selected type, classifies it with the correct value baseline, and reports cases whose CNV was dropped because they lack the requested type.

## What changed

**Data model**
- New optional `ds.queries.singleSampleMutation.cnvTypes[]` (`{ id, label, valueType, dataType }`), exposed to the client via `termdb/config`. GDC declares two: masked (`segmean`) and allele-specific (`copyNumber`).
- New optional `cnvOptions.cnvType` (the selected type id) on the GRIN2 request.

**Server**
- `grin2/main.ts` resolves the selected type, passes it to `singleSampleMutation.get()`, and classifies with that type's `valueType` (selected type → ds-level `cnv.type` → `log2ratio` fallback). Non-`cnvTypes` datasets are unaffected.
- GDC getter (`mds3.gdc.js`): `getCnvFusion4oneCase` loads only the requested `data_type` and tracks whether the case had *any* CNV; `getSingleSampleMutations` flags `droppedCnvNoMatch` when a case has CNV but none of the requested type.
- GDC CNV parsing now stamps a per-segment `valueType`. The parsing was extracted into a pure, exported `parseGdcCnvFile()`.

**Drop semantics (by design)**: a mismatched case loses **only its CNV** contribution; its SNV/fusion/SV lesions are kept, so it still counts as "processed." The count surfaces as a gated **"Dropped (CNV type unavailable)"** row in the stats Summary.

**Stats**
- Loss/Gain sample counts are now reported **per lesion type** (samples with ≥1 loss vs ≥1 gain) instead of a single per-data-type count.

**Client**
- CNV row now renders for datasets exposing `cnvTypes` (GDC has no `queries.cnv`). Radio buttons are sourced from `cnvTypes`; switching rebuilds the threshold inputs for the new type's `valueType` (defaults/ranges differ). Selection is sent in the request and persisted across runs.
- Radios get per-option `data-testid="sjpp-grin2-cnvtype-<id>"`; existing testids unchanged.

## Testing

- Existing `grin2` unit suite (61 tests) covers `filterAndConvertCnv` per-`valueType` classification — still green.
- New unit tests for `parseGdcCnvFile()`: segment-mean stamps `valueType:'segmean'` (+ `chr` prefix handling); copy-number stamps `valueType:'copyNumber'`, emits a `dtloh` event only for strict one-allele loss, and leaves LOH unstamped; unknown header throws.
- `tsc` clean across shared/types, server, client; prettier clean.

**Manual validation** across GBM and breast (ductal+lobular) cohorts, both CNV types:
- Mutation lesion/sample counts are **byte-identical** between the two CNV types in each cohort (confirms the CNV-type switch never touches SNV).
- Drop count tracks data coverage (GBM allele-specific 74 → masked 244; breast 14 → 216).
- Drivers are stable and biologically faithful (GBM: EGFR/CDKN2A/B/PTEN/TP53; breast: CDH1, 11q13 amplicon, ERBB2), with amplicon magnitude correctly softening under log2-ratio thresholds.

## Checklist

### DIFF
--- a/server/src/grin2/main.ts
+++ b/server/src/grin2/main.ts
@@ -307,6 +307,10 @@ async function runGrin2(g: any, ds: any, request: GRIN2Request, signal?: AbortSi
 						['Processed Samples', processing.processedSamples!.toLocaleString()],
 						['Unprocessed Samples', (processing.unprocessedSamples ?? 0).toLocaleString()],
 						['Failed Samples', processing.failedSamples!.toLocaleString()],
+						// only shown when a cnv-type selection actually dropped some samples' cnv
+						...(processing.droppedCnvNoType
+							? [['Dropped (CNV type unavailable)', processing.droppedCnvNoType.toLocaleString()]]
+							: []),
 						['Total Lesions', processing.totalLesions!.toLocaleString()],
 						['Processed Lesions', processing.processedLesions!.toLocaleString()]
 					]
@@ -577,6 +581,7 @@ async function processSampleData(
 		totalLesions: number
 		processedLesions: number
 		unprocessedSamples: number
+		droppedCnvNoType: number
 		lesionCap?: number
 		lesionCounts?: {
 			total: number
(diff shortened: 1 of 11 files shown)


server/src/grin2/main.ts
```diff
@@ -588,12 +593,25 @@ async function processSampleData(
...
+	const cnvTypeDefs = ds.queries?.singleSampleMutation?.cnvTypes as
+		| { id: string; label: string; valueType: CnvType; dataType: string }[]
+		| undefined
+	const selectedCnvDef = request.cnvOptions?.cnvType
+		? cnvTypeDefs?.find(t => t.id === request.cnvOptions!.cnvType)
+		: undefined
+
+	// How this ds quantifies cnv values; drives gain/loss classification. The selected type wins; else the
+	// ds-level default (file-based datasets); else 'log2ratio' (legacy default, diploid baseline 0).
+	const cnvType: CnvType = selectedCnvDef?.valueType ?? ds.queries?.cnv?.type ?? 'log2ratio'
```
xzhou82, line 608: should not fall back to ds.queries?.cnv?.type
either cnv type is defined in singleSampleMutation{}, or use default log2ratio

xzhou82 (approved):

PR #4637, opened 2026-06-24
Attempt to fix GDC cohort MAF multi-file download failures on qa-int

compbiolover:
# Description
### Problem
Against GDC's **qa-int** API, the cohort MAF builder fails while **prod works fine**:
- Selecting **3 MAF files** returns a `.maf.gz` containing data from **only the 1st file**; the other 2 fail. (Each file URL downloads fine individually on qa-int.)
- Downloading **all WXS files** never finishes and is killed at 5 min by the watchdog in `rust/index.js`.

The 3-file case still produced a *finalized* gz, so the rust process wasn't crashing — the other downloads were being rejected/stalled. This is a **concurrent-connection / HTTP-client** issue against qa-int's stricter (slower, proxied/WAF'd, connection-limited) stack, not a parsing bug — and it was masked by mislabeled error reporting.

### Changes (one commit per step)
1. **Diagnostics** (`gdcmaf.rs`) — surface the *real* per-file error: HTTP status, version, `Content-Type`/`Content-Encoding`, body size, a byte/body preview, and the full transport error `source()` chain. Previously a failed body read (e.g. *"connection closed before message completed"*) was mislabeled `"Failed to decompress"`, which misdirected earlier fixes.
2. **Unit tests** (`gdcmaf.rs`) — first Rust coverage for this binary: `select_maf_col`, `format_error_chain`, `preview_bytes` (9 tests, no new deps).
3. **Headers** (`gdc.mafBuild.ts`) — the binary `/data/<uuid>` downloads now send **only** auth (`Cookie` / `X-Auth-Token`) plus a browser-like `User-Agent`. Stops forwarding the JSON-API headers (`connection: close`, `Content-Type`/`Accept: application/json`) that leaked from `getHostHeaders()` and are harmful on a gzip download against a strict proxy/WAF. `getHostHeaders()` is unchanged for the ky metadata calls.
4. **Connections** (`gdcmaf.rs`) — re-enable keep-alive reuse (drop `pool_max_idle_per_host(0)`, which forced a brand-new connection per file), lower default concurrency **10 → 3** (overridable via the input JSON `concurrency` field), and restructure the retry loop to wrap the whole download (send **+** body read) so mid-body drops are retried. Non-2xx and decompression failures are treated as non-transient.
5. **Watchdog** (`rust/index.js`, `gdc.maf.ts`) — make the 5-min `gdcmaf` kill timeout configurable via `serverconfig.features.gdcMafMaxElapsed` (default 300000ms), threaded through `stream_rust(..., { maxElapsed })`. Documented as a safety **backstop**, not the cause of the hang.

### Verification
- `cargo build --release` + `cargo clippy` clean; `cargo test --bin gdcmaf` → 9/9 pass; server `tsc --noEmit` + eslint clean.
- **Local build against prod**, 3 files: produces a correct merged MAF — all 3 cases present (8,136 + 7,332 + 13,533 = 29,001 data rows), single header, 140-column rectangular, valid gzip.
- **Still needed:** deploy to **qa-int** and run the 3-file and full-WXS cases. Step 1's diagnostics will confirm the fix or print exactly what remains (e.g. a 403/406 or connection reset), which would guide further tuning (e.g. `concurrency: 1`, or a descriptive vs browser UA).

### Notes
- Defensive content-aware decompression and a ragged-row panic guard in `select_maf_col` were scoped as a lower-priority follow-up and are **not** included here.

## Checklist

### DIFF
--- a/rust/src/gdcmaf.rs
+++ b/rust/src/gdcmaf.rs
@@ -30,6 +30,30 @@ struct ErrorEntry {
     error: String,
 }
 
+// Flatten a std::error::Error (e.g. reqwest::Error) and its `source()` chain into
+// a single string, so the real transport-level cause (e.g. "connection closed
+// before message completed", connection reset, HTTP/2 stream error) is visible in
+// the per-file error rather than only the top-level wrapper message.
+fn format_error_chain(e: &dyn std::error::Error) -> String {
+    let mut msg = e.to_string();
+    let mut src = e.source();
+    while let Some(s) = src {
+        msg.push_str(" -> ");
+        msg.push_str(&s.to_string());
+        src = s.source();
+    }
+    msg
+}
+
+// Render the first bytes of a response body for diagnostics: helps tell whether the
+// server returned gzip (magic 1f 8b), plain text, or an HTML/JSON error page.
+fn preview_bytes(content: &[u8]) -> String {
+    let n = content.len().min(64);
+    let hex: Vec<String> = content[..n].iter().map(|b| format!("{:02x}", b)).collect();
+    let ascii = String::from_utf8_lossy(&content[..n]);
+    format!("hex=[{}] ascii=\"{}\"", hex.join(" "), ascii.escape_default())
+}
+
 fn select_maf_col(d: String, columns: &Vec<String>, url: &str) -> Result<(Vec<u8>, i32), (String, String)> {
     let mut maf_str: String = String::new();
     let mut header_indices: Vec<usize> = Vec::new();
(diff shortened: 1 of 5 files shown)


siosonel (approved):

server/src/routes/gdc.maf.ts
```diff
@@ -35,6 +35,16 @@ const allowedWorkflowType = 'Aliquot Ensemble Somatic Variant Merging and Maskin
 // change to 400 so it won't limit number of files; should keep this setting as a safeguard; also it's fast to check file size (.5s in gdc.mafBuild.ts)
 export const maxTotalSizeCompressed = serverconfig.features.gdcMafMaxFileSize || 400000000 // 400Mb
 
+// watchdog kill threshold (ms) for the gdcmaf rust process; a safety backstop against a
+// hung/leaked download. Defaults to 5 minutes; overridable via serverconfig for slower
+// GDC environments (e.g. qa-int) where a legitimate large download may need more time.
+export const gdcMafMaxElapsed = serverconfig.features.gdcMafMaxElapsed || 300000 // 5 min
```
xzhou82: these two constants should be moved into gdc.mafBuild.ts

compbiolover: Moved to `gdc.mafBuild.ts` in [0f01dbf](https://github.com/stjude/proteinpaint/pull/4637/commits/0f01dbf08dca93ffa3797a3c51f91fb4a8634a15)

rust/src/gdcmaf.rs
```diff
@@ -338,3 +431,135 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     io::stderr().flush().expect("Failed to flush stderr");
     Ok(())
 }
+
+#[cfg(test)]
```
siosonel, line 435: @compbiolover is it common or best practice in Rust to have test code with src code (in the same file)? and how do you run these tests?

siosonel: From Andrew:
Claude specifically says that it is convention in rust
Short answer: yes, for unit tests this is the idiomatic Rust convention — and for [gdcmaf.rs](http://gdcmaf.rs/) specifically it's also the only clean option.
The Rust convention
Rust splits tests into two kinds, with different homes:
• Unit tests → same file, in a #[cfg(test)] mod tests block at the bottom. This is exactly what the official Rust Book recommends ([ch. 11](https://doc.rust-lang.org/book/ch11-03-test-organization.html)). The #[cfg(test)] attribute means the module is compiled only during cargo test — zero cost and zero size in the release binary. The whole point is that unit tests can reach private items (select_maf_col, format_error_chain, preview_bytes are all private fns), which a separate test crate cannot.
• Integration tests → separate tests/ directory at the crate root. These compile as a separate crate and can only see the public API. (edited)

PR #4639, opened 2026-06-24
CohortMAF: forward the user's real User-Agent to GDC downloads

compbiolover:
# Description
Send the **end user's real User-Agent** on the GDC cohort-MAF binary `/data/<uuid>` downloads, instead of the hardcoded browser User-Agent introduced with the qa-int connection fix. This way GDC's proxy/WAF (e.g. on qa-int) sees the actual client rather than a static string.

Follow-up to the qa-int multi-file MAF download fix (already merged into `release-2.191`).

## Changes
- Capture the user agent in the route handler: `req.headers['x-forwarded-agent'] || req.headers['user-agent']` — the forwarded agent set by an upstream proxy, falling back to the direct request's UA.
- Thread it into `buildMaf(...)` and set it on the download headers **only when present** (so no `User-Agent` is sent if neither is available, rather than an invalid `undefined`).
- Remove the hardcoded `gdcDownloadUserAgent` constant.

## Checklist

### DIFF
--- a/server/src/routes/gdc.mafBuild.ts
+++ b/server/src/routes/gdc.mafBuild.ts
@@ -7,13 +7,6 @@ import { maxTotalSizeCompressed } from './gdc.maf.ts'
 import { mayLog } from '#src/helpers.ts'
 import serverconfig from '#src/serverconfig.js'
 
-// GDC's stricter environments (e.g. qa-int) sit behind a proxy/WAF that can reject requests
-// from header-less HTTP clients. Identify the file downloads with a normal browser User-Agent.
-// (prod works without it; this is a defensive measure for qa-int and can be swapped for a
-// descriptive UA if a browser-like agent turns out not to be required.)
-const gdcDownloadUserAgent =
-	'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36'
-
 // watchdog kill threshold (ms) for the gdcmaf rust process; a safety backstop against a
 // hung/leaked download. Defaults to 5 minutes; overridable via serverconfig for slower
 // GDC environments (e.g. qa-int) where a legitimate large download may need more time.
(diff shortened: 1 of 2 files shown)


server/src/routes/gdc.mafBuild.ts
```diff
@@ -46,7 +39,11 @@ export function init({ genomes }) {
 			const ds = g.datasets.GDC
 			if (!ds) throw 'hg38 GDC missing'
 			const q: GdcMafBuildRequest = req.query
-			await buildMaf(q, res, ds)
+			// the end user's real User-Agent: set by an upstream proxy as x-forwarded-agent,
+			// otherwise the direct request's user-agent. Forwarded onto the GDC file downloads
+			// so GDC's proxy/WAF sees the actual client rather than a header-less rust client.
+			const userAgent = req.headers['x-forwarded-agent'] || req.headers['user-agent']
```
siosonel: this has to use `req.headers['X-Forwarded-Agent']`, the letter casing should match exactly what is set by `trackReqHeaders()` option by a dataset such as `sjpp/ppgdc/active/dataset/gdc.hg38.ts`

compbiolover: Ah, okay, thanks. I didn't realize it was case sensitive

siosonel: correction - I see that you are using the `req` argument directly, which should mean there will be no `X-Forwarded-Agent` key-value.

siosonel (approved):

siosonel:
tested locally by both Edgar and Andrew:
```text
Chrome: userAgent Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36

Safari: userAgent Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Safari/605.1.15

Firefox: userAgent Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:145.0) Gecko/20100101 Firefox/145.0
```

PR #4650, opened 2026-06-29
Make launched Disco plots session-saveable

ItsVictorLP:
# Description
Disco plots launched ad-hoc from matrix/mds3 were created in an isolated PlotApp and therefore their interactive state (filters, CNV source selection, etc.) could not be saved into the MASS session state.

The change aims to allow launched Disco plots to become ordinary MASS plots so their args and subsequent user edits are persisted and recoverable in saved sessions.
## Checklist

### DIFF
--- a/client/plots/gb/GB.ts
+++ b/client/plots/gb/GB.ts
@@ -212,6 +212,7 @@ class TdbGenomeBrowser extends PlotBase implements RxComponent {
 	getOpts() {
 		const opts = {
 			genome: this.app.opts.genome,
+			app: this.app,
 			vocabApi: this.app.vocabApi,
 			debug: this.app.opts.debug,
 			plotDiv: this.opts.plotDiv,
--- a/client/mds3/sampletable.js
+++ b/client/mds3/sampletable.js
@@ -31,7 +31,7 @@ feedSample2selectCallback
 .block
 .div
 .tid2value={}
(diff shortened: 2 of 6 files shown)


client/plots/gb/GB.ts
```diff
@@ -212,6 +212,7 @@ class TdbGenomeBrowser extends PlotBase implements RxComponent {
 	getOpts() {
 		const opts = {
 			genome: this.app.opts.genome,
+			app: this.app,
```
xzhou82, line 215: at [link](http://localhost:3000/?mass={%22dslabel%22:%22TermdbTest%22,%22genome%22:%22hg38-test%22,%22nav%22:{%22activeTab%22:-1},%22plots%22:[{%22chartType%22:%22genomeBrowser%22,%22snvindel%22:{%22shown%22:true},%22geneSearchResult%22:{%22geneSymbol%22:%22p53%22}}]}) click any mutation > disco and it errors out. suspect it's due to this as vocab needs to do structuredClone on the arg.

ItsVictorLP: I cannot access the link due to an invalid dslabel:
Screenshot 2026-07-06 at 1 11 57 AM
I got a new computer and might be missing some datasets, I did my best to fix this assuming DataCloneError. I also do not have access to hpc to download the missing datasets

PR #4657, opened 2026-06-30
Gene data type implementation in omnisearch, improved regular bins feature in chatbot and number of token check 

robinpaul85:
# Description

1) Improved `regular bins` in summary skill. Improved scaffolding of ` regular bins` information into their respective scafoolds.

2) Added a simple check to ensure that length of prompt is lower than the context limit of the model. Needs a small change in serverconfig format. Will need to update prod PP serverconfigs once this branch is merged and released to production. 

3) In omni search, added support for various gene types (SNV/indel, CNV, SV/fusion, DNA methylation and gene expression). 

## Checklist

### DIFF
--- a/client/mass/chat.ts
+++ b/client/mass/chat.ts
@@ -6,6 +6,7 @@ import { dofetch3 } from '#common/dofetch'
 import type { ChatRequest, ChatResponse } from '#types'
 import { sayerror } from '../dom/sayerror.ts'
 import { select } from 'd3-selection'
+import { fillTermWrapper } from '#termsetting'
 
 const MIN_PROMPT_LENGTH_FOR_OMNISEARCH = 3 // Set a minimum prompt length for omnisearch to trigger
 const MAX_PROMPT_LENGTH_FOR_OMNISEARCH = 15 // Set a maximum prompt length for omnisearch to trigger
@@ -24,6 +25,7 @@ class MassAiChatBot implements RxComponent {
 	noResult: any
 	isChat: any
 	hasGeneExp: any
+	hasGeneVariant: any
 
 	constructor(opts: any) {
 		this.type = MassAiChatBot.type
@@ -33,6 +35,13 @@ class MassAiChatBot implements RxComponent {
 		this.opts.targetType = this.opts.targetType ? this.opts.targetType : 'Dictionary Variables'
 		this.isChat = this.app.getState().termdbConfig?.queries?.chat // Storing if chat is supported by the dataset for easy access in other methods
 		this.hasGeneExp = this.app.getState().termdbConfig?.queries?.geneExpression // Whether the dataset supports gene expression, gating gene search in omnisearch
+		this.hasGeneVariant = false // Whether the dataset supports gene variant, gating gene search in omnisearch
+		if (
+			this.app.getState().termdbConfig?.queries?.snvindel ||
+			this.app.getState().termdbConfig?.queries?.cnv ||
+			this.app.getState().termdbConfig?.queries?.svfusion
+		)
+			this.hasGeneVariant = true
 		setRenderers(this) // needed so that this.showTerms, noResult, clear work
 	}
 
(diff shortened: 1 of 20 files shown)


client/mass/chat.ts
```diff
@@ -56,7 +82,37 @@ class MassAiChatBot implements RxComponent {
 			return
 		}
 		const cohortStr = this.getState(this.app.getState()).cohortStr
-		const data = await this.app.vocabApi.findTerm(prompt, cohortStr, this.opts.usecase, this.opts.targetType)
+		// Search dictionary variables and (when supported) gene names for gene expression and
+		// DNA methylation in parallel. Each gene search returns only when the dataset supports it.
+		const [data, geneExpressionHits, geneVariantHits] = await Promise.all([
```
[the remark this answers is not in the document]
robinpaul85: Comment outdated

client/mass/chat.ts
```diff
@@ -46,17 +144,81 @@ class MassAiChatBot implements RxComponent {
...
 	}
 
+	// Resolve which gene data types this dataset supports by asking the server (termdb/chat endpoint,
+	// getGeneDataTypes) for its available data types, instead of inspecting termdbConfig.queries on the
+	// client. Sets the flags that gate gene search in the omnisearch, and builds the per-variant-type
+	// options (label + dt candidates) for each variant data type the dataset defines. On failure, gene
+	// search is left disabled rather than breaking the omnisearch.
+	async setDataTypes() {
+		try {
+			const data: any = await dofetch3('termdb/chat', {
```
xzhou82: bad approach. don't run this request. chat route will package all data types available for each gene hit
when we deal with gene panel in future, datatype availability isn't uniform for all genes

server/src/chat/routeAPIcall.ts
```diff
@@ -2,13 +2,49 @@ import type { LlmConfig } from '#types'
...
+		return {
+			type: 'text',
+			text: 'LLM model configuration is missing a valid modelName value in serverconfig.'
+		} as MsgToUser
+	}
+	if (tokenLengthResult.limitExceeded) {
+		return {
+			type: 'text',
+			text: `Prompt exceeds maximum token length for the selected model. Token count for the prompt: ${tokenLengthResult.tokenCount}. Max token limit for ${model} is ${tokenLengthResult.maxTokens}.`
+		} as MsgToUser
```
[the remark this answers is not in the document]
robinpaul85: Solved issue by using `const model = modelOverride?.modelName ?? llm.model?.modelName`

client/mass/chat.ts
robinpaul85: Fixed server route code such that in a single server side request it determines match to dictionary variables and genes. If there is a match to a gene, it determines what data types are available for this dataset. Next, it determines for each gene in the genome, whether those ds level gene types are actually available for those genes. Currently, it does not do any filtering at the gene level, but the function has been created which will do the gene=level filtering when we add support for gene specific assays such as gene panels.

server/src/chat/routeAPIcall.ts
```diff
@@ -1,14 +1,58 @@
...
+	if (isMsgToUser(tokenLengthResult)) return tokenLengthResult
+	if (tokenLengthResult.limitExceeded) {
+		return {
+			type: 'text',
+			text: `Prompt exceeds maximum token length for the selected model. Token count for the prompt: ${tokenLengthResult.tokenCount}. Max token limit for ${model} is ${tokenLengthResult.maxTokens}.`
+		} as MsgToUser
+	}
 	let response: string | MsgToUser
 	if (llm.provider === 'SJ') {
 		// Local SJ server
```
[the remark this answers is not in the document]
robinpaul85: `huggingface` is only used for embedding model providers, but our code currently does not use embedding models so this is ok. Later for bigger datasets, we may use embedding models that is when this might be useful.

shared/types/src/routes/termdb.chat.ts
```diff
@@ -32,26 +32,58 @@ export type PlotResponse = {
...
 	EmbeddingProviderApiToken?: string
-	modelName: string
+	model: ModelConfig
 	embeddingModelName: string
 	/** Whether to load the embedding model locally (via transformers.js) or call a remote API. Defaults to 'api'. */
 	embeddingModelAccess?: 'local' | 'api'
-	/** Smaller model to use for LLM classification fallback. Defaults to modelName if not set. */
-	classifierModelName?: string
+	/** Smaller model to use for LLM classification fallback. Defaults to `model` if not set. */
+	classifierModelName?: ModelConfig
```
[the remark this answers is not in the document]
robinpaul85: Done

server/src/chat/search.ts
```diff
@@ -0,0 +1,153 @@
...
+		cnv: Boolean(ds.queries?.cnv),
+		svfusion: Boolean(ds.queries?.svfusion)
+	}
+}
+
+/** Handle a single mass omnisearch request: search dictionary variables and genes together, without
+ * invoking the AI chat pipeline. Dictionary search reuses the dataset's term-search query (the same
+ * path as the termdb findterm route); gene search uses searchGeneNames() below. */
+export async function runOmnisearch(q: any, req: any, ds: any, genome: any): Promise<OmnisearchResult> {
+	const prompt = typeof q.prompt == 'string' ? q.prompt.trim() : ''
```
xzhou82, line 26: if prompt is not non-empty str, it should not do any search here. thus no need for if(prompt) at line 30

server/src/chat/search.ts
```diff
@@ -0,0 +1,153 @@
...
+	// resolve a default genomic coordinate for genes that need a genome browser (DNA methylation) so the
+	// client can seed the browser track without a separate genelookup request.
+	const genes: GeneMatch[] = geneNames.map(gene => {
+		const dataTypes = getGeneDataTypesForEachGene(ds, gene, datasetDataTypes)
+		const coord = dataTypes.dnaMethylation ? getGeneCoord(genome, gene) : null
+		return { gene, dataTypes, coord }
+	})
+
+	// Will later add support for other NonDict terms such as genesets etc.
+	return { lst: terms, genes }
```
xzhou82: return `{terms,genes}` and do not use lst key which is ambiguous

robinpaul85: I have replaced `lst` with `dictionaryTerms` to make it more intuitive. Later we may add support for `genesets` in which we can just append a new field to this object. I hope this is ok.

server/src/chat/search.ts
robinpaul85: Done. Client side already does a min and max length check for the search input. In case its does not fall within these thresholds server side is not invoked.

xzhou82 (approved):

PR #4670, opened 2026-07-06
use uiLabels in violin and boxplot interactions menus

siosonel:
# Description

Fixes https://gdc-ctds.atlassian.net/browse/SV-2754

Tests:
- open http://localhost:3000//example.gdc.correlation.html?noheader=1&cohort=TCGA-LGG
- use numeric terms as primary and overlay variables
- click on `Violin` tab, then click on a series label, there should be a `List cases` option, instead of `List samples`, click that option, the table should have a `Case` column title instead of `Sample`
- click on `Boxplot` tab and do similar clicks/interactions as with the violin plot
- also tested locally with all unit and integration tests

## Checklist

### DIFF
--- a/client/plots/boxplot/interactions/BoxPlotInteractions.ts
+++ b/client/plots/boxplot/interactions/BoxPlotInteractions.ts
@@ -1,6 +1,7 @@
-import type { BoxPlotDom, LegendItemEntry, BoxPlotConfig } from '../BoxPlotTypes'
+import type { BoxPlotDom, LegendItemEntry } from '../BoxPlotTypes'
 import type { MassAppApi } from '#mass/types/mass'
 import type { RenderedPlot } from '../view/RenderedPlot'
+import type { TdbBoxplot } from '../BoxPlot.ts'
 import { ListSamples } from '#dom/summary/ListSamples'
 import { filterJoin, getFilterItemByTag } from '#filter'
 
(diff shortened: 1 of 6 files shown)


client/plots/boxplot/interactions/BoxPlotInteractions.ts
```diff
@@ -114,7 +117,7 @@ export class BoxPlotInteractions {
 	}
 
 	getPlotConfig() {
-		return this.app.getState().plots.find((p: BoxPlotConfig) => p.id === this.id)
+		return this.boxplot.state.config
```
siosonel, line 120: @creilly8 @gavrielm I updated the constructor argument for `BoxPlotInteractions`, so that the up-to-date state.config is available directly.  Note that `BoxPlot.getState()` may compute state properties that may not be available from the `app.state.plots` entry, not sure if there are other plots that make similar  potentially incorrect assumption regarding plot state or state.config.

gavrielm (approved):
Works for me. In a later PR, can implement a dynamic approach for labeling the violin/boxplot based on sample type, as is done in `setSampleType()` in `client/plots/barchart.js`

xzhou82:
[gene exp scatter](http://localhost:3000/?mass={%22dslabel%22:%22GDC%22,%22genome%22:%22hg38%22,%22nav%22:{%22header_mode%22:%22hidden%22},%22termfilter%22:{%22filter0%22:{%22op%22:%22in%22,%22content%22:{%22field%22:%22cases.disease_type%22,%22value%22:%22Gliomas%22}}},%22plots%22:[{%22chartType%22:%22summary%22,%22term%22:{%22term%22:{%22type%22:%22geneExpression%22,%22gene%22:%22idh1%22},%22q%22:{%22mode%22:%22continuous%22}},%22term2%22:{%22term%22:{%22type%22:%22geneExpression%22,%22gene%22:%22egfr%22},%22q%22:{%22mode%22:%22continuous%22}}}]}) still shows sample

PR #4686, opened 2026-07-10
Integration test fixes for ListSamples and GSEA

creilly8:
# Description

Changes: 
1. The timeout for the ListSamples integration tests are increased and `test.end()` was replaced with `test.plan()`. `test.plan()` ought to mostly prevent a concurrently running test throwing a `.end() is already called error`. 
2. The GSEA integration test was updated to use blitzgsea since cerno is not enabled on CI. 
3. The old `plots/test/gsea.spec.ts` is deleted. 

Tests pass for me locally and[ on CI](https://github.com/stjude/proteinpaint/actions/runs/29107565853#artifacts).

## Checklist


client/dom/summary/test/ListSamples.integration.spec.ts
```diff
@@ -80,7 +80,8 @@ tape('\n', function (test) {
 })
 
 tape('ListSamples.getData() for term=numeric returns the correct data object', async test => {
-	test.timeoutAfter(1000)
+	test.timeoutAfter(3000)
```
siosonel, line 83: Do we have an idea of what caused the `/termdb?for=matrix` response to take longer? I'm wondering if some recent backend change is causing a bottleneck that may be obvious when listing samples for datasets that >10000 samples like SJLife.

creilly8: I'm not sure. The slowest test times were for the gene variant and/or gene expression terms.

xzhou82 (approved):

PR #4740, opened 2026-07-17
Fix repository_dispatch: use github-script instead of peter-evans@v4

srisivanicharan:
# Description
Replaces the failing peter-evans/repository-dispatch@v4 step which triggers the downstream deploy in stjude/sj-pp  with an equivalent actions/github-script@v7 step. Fixes the Request body length does not match content-length header error that was breaking the Trigger deploy in sj-pp step.
## Checklist

### DIFF
--- a/.github/actions/publish-ppfull-ppserver-image/action.yml
+++ b/.github/actions/publish-ppfull-ppserver-image/action.yml
@@ -69,11 +69,34 @@ runs:
         echo "front_version=$(node -p "require('./front/package.json').version")" >> "$GITHUB_OUTPUT"
         echo "server_version=$(node -p "require('./server/package.json').version")" >> "$GITHUB_OUTPUT" 
 
-    - name: Run deploy action
-      uses: peter-evans/repository-dispatch@v4
+    # Trigger the downstream deploy in stjude/sj-pp via a repository_dispatch event.
+    # Replaces peter-evans/repository-dispatch@v4: its bundled @octokit/request@8.4.1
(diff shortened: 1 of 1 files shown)


.github/actions/publish-ppfull-ppserver-image/action.yml
```diff
@@ -69,11 +69,34 @@ runs:
...
-    - name: Run deploy action
-      uses: peter-evans/repository-dispatch@v4
+    # Trigger the downstream deploy in stjude/sj-pp via a repository_dispatch event.
+    # Replaces peter-evans/repository-dispatch@v4: its bundled @octokit/request@8.4.1
+    # tags string bodies with duplex:"half", which Node 24's undici rejects as
+    # "Request body length does not match content-length header".
+    - name: Trigger deploy in sj-pp
+      # Only fire on non-release branches (release/prerelease deploys are handled elsewhere).
       if: ${{ !startsWith(github.ref_name, 'release') && !startsWith(github.ref_name, 'prerelease') }}
+      uses: actions/github-script@v7
```
siosonel: Wny not at least v8? From https://github.com/actions/github-script. I see `Version 8 of this action updated the runtime to Node 24`

srisivanicharan: We should definitely have v8 since it is meant for Node 24 where as V7 is for Node 20. Thank you for letting me know

siosonel (approved):

PR #4744, opened 2026-07-20
GDC: Move GDC server code out of server into ppgdc

compbiolover:
# Description
Companion to the ppgdc/sjpp PR (stjude/sjpp#1426) — **merge that first**, since it
carries the ppgdc side that consumes what this PR removes.

## What

Relocates all GDC-specific server code out of `@sjcrh/proteinpaint-server`. The
GDC dataset now owns its own query implementations, wired through the same
dataset hooks the MMRF dataset already uses — no more `.gdcapi` / `src == 'gdcapi'`
branches in generic server code.

`server/src/mds3.gdc.js` (2944 lines) and `server/src/mds3.gdc.filter.js` (120)
were 100% GDC-specific and are deleted here; they move to `ppgdc/active/dataset/gdc/`
in the paired PR. Net: **−3,700 lines** from the server package.

## Why

A GDC-only change used to require a full proteinpaint release + ppserver image
publish before it could ship. With the code in ppgdc, GDC iterates on its own
cadence. This also makes GDC structurally identical to MMRF (both pure hook
implementations), which is the groundwork for treating API-backed datasets as a
first-class kind alongside sqlite ones.

## How

- Dead code removed first (zero-behavior-change prep): unused imports/exports,
  a dead `app.unorg.js` import, `validate_variant2sample`, an unused
  `validate_query_geneCnv` + its private helper.
- Every inbound edge converted to a dataset-supplied getter. Generic server code
  now checks `typeof q.get == 'function'` and skips its own wiring; the GDC
  dataset pre-sets the getter. Applies to snvindel by{range,isoform}, m2csq,
  geneExpression, singleSampleMutation, singleCell data/DEgenes, convertSampleId,
  ssm2canonicalisoform, termid2totalsize2, variant2samples, and the
  assay-availability sample-set query.
- Server-internal utilities the GDC code still needs (`xfetch`, `mayLog`,
  `renderVolcano`, `combineSamplesById`) are injected via `dsHelpers`, the
  existing one-directional server→dataset channel.
- `shared/utils/src/gdcCase.ts`: `flattenCaseByFields` had three copies across
  two repos. Consolidated to one shared definition (details in the paired PR);
  this is the shared half.

## Scope note

`gdc_validateGeneExpression` (singleCell geneExpression) is still GDC code living
in `singleCell/samplesRoute.ts` — it was defined locally and never imported the
moved file, so it didn't block this move. It's the next thing to relocate.

## Testing

- `npm run test:unit` (server): 1549/1549
- `shared/utils` tests: 381/381 (incl. the moved gdcCase tapes)
- `tsc` clean (server + root)
- Manually exercised against live GDC: oncomatrix, lollipop, sunburst, disco,
  survival + gene expression, GRIN2, CNV tool (filtered + unfiltered). MMRF
  correlation + survival charts verified for the shared-helper change.

## Checklist


server/src/singleCell/samplesRoute.ts
```diff
@@ -103,8 +102,8 @@ export async function validate_query_singleCell(ds: any, genome: any): Promise<v
 	}
 
 	// validate required q.data{}
-	if (q.data.src == 'gdcapi') {
-		gdc_validate_query_singleCell_data(ds, genome) // todo change to ds-supplied q.data.get()
+	if (typeof q.data.get == 'function') {
+		// ds supplied getter
 	} else if (q.data.src == 'native') {
```
xzhou82, line 107: after migrating gdc to getter, no longer needs to support src='native'

xzhou82 (approved):

PR #4767, opened 2026-07-24
GDC: Finish geting GDC off the .gdcapi/src flags and drop the dead useFilter0 path

compbiolover:
# Description
This is the server side of finishing the GDC-into-[ppgdc](https://github.com/stjude/sjpp/pull/1448) move. After this, GDC works the same way MMRF does: the dataset supplies all its own query getters and the server has no idea it's GDC. `rg gdcapi server/src` comes back with a single comment explaining why a param is named `_genome`. `rg "dslabel.*'GDC'" client server/src` comes back with nothing.

Four things happen here.

**The last bit of GDC query code leaves the server.** `gdc_validateGeneExpression()` (the single-cell gene-expression getter) moves into ppgdc's `singleCell.ts` and gets wired up through `initQueries()` like the other singleCell getters. With it gone I could delete the `geneExpression.src == 'gdcapi'` branch and every other `.gdcapi` / `src == 'gdcapi'` check, including an inert one in `topTermsByType.ts` that GDC never actually reached.

**The bespoke GDC types are gone, and so is `src`.** `SingleCellDataGdc`/`Native`/`Base` collapse into one `SingleCellData`, `SingleCellGeneExpressionGdc`/`Native` into `SingleCellGeneExpression`, `SingleCellDEgeneGdc` into `SingleCellDEgenes`; `GDCSingleCellPlot`, the `GdcApi` flag mixin (plus its five inert `gdcapi?: boolean` fields, which nothing read anywhere) and the entirely-unused `GdcFilter0` are deleted. `src` is removed outright rather than made optional — from `singleCell.data`, `singleCell.geneExpression`, `singleSampleMutation` and `metaboliteIntensity`. All four now dispatch purely on whether the ds supplied a `get()`, which is what `singleSampleMutation`'s own TODO had been asking for.

What the union used to encode at compile time is now enforced at init with a clear message instead: `plots[]`, `plot.folder`, `geneExpression.folder`, `singleSampleMutation.folder`. I hoisted the plot check into its own `validateDataPlots()` — `validateSamples()` runs *before* `validateDataNative()` and reads `plot.folder` itself, so a guard left in the latter would have been dead for the path that actually reaches it first. `dataset.ts` nets −19 lines and 11 types become 4; `samplesRoute.ts` nets −64.

**The remaining `dslabel == 'GDC'` branches on the client become dataset config.** Five of them — the mutated-cases limit in `TermdbVocab`, `remove_empty_sample` and dictionary-term-first sorting in matrix, the boxplot Documentation button, and the gb block-recreate path (which was `['GDC','MMRF']`) — now read `termdbConfig` flags, following the `maxAnnoTermsPerClientRequest` precedent already in that file. Both matrix flags default off, since `matrix.sort.js` warns in-line about not disturbing the published PNET figure. Server-side, the `ds.__gdc.doneCaching` gate duplicated in `app.middlewares.js` and `termdb.cluster.ts` becomes an `init.notReadyMessage()` ds hook, resolving the `// TODO: use generalized ds properties` that sat on it.

That hook can fail open in a way the hardcoded check couldn't — drop it from the ds file and the gate silently disappears — so `mds3.init.js` now refuses to launch a `hasNonblockingSteps` dataset that doesn't define it.

**The redundant useFilter0 assay-availability path is gone.** GDC was resolving assay-availability sample sets through its own `useFilter0`/`getSampleSet` branch in `filterSamples4assayAvailability()`. That whole branch, dt-term stripping and all, does the same thing as the generic `ds.cohort.termdb.filterSamples` hook GDC already implements — `filter2GDCfilter` handles the dt terms the same way at the root, and both throw on nested geneVariant. So GDC just falls through to the generic path now, like MMRF. Net −46 lines.

### Two fixes found while wiring this up
- **SC experiment links were hardcoded to the GDC portal.** `SCViewModel.ts` had `dslabel == 'GDC'` → `https://portal.gdc.cancer.gov/files/…`, a regression against the legacy `singleCellPlot.js`, which does it generically via `termdbConfig.urlTemplates.scrnaExperimentId` — a template GDC already declares. Reverted to the template; the two unit tests now key off the template instead of the dslabel.
- **`sample2gene2expressionBins` was only ever seeded by the built-in getter**, but `termdb.getDefaultBins.js` indexes it unconditionally. Once GDC's getter left the server package nothing seeded it for GDC. Now seeded for every dataset.

Also, `MetaboliteIntensityQuery`/`…Native` turned out to be completely unreferenced while `Mds3Queries.metaboliteIntensity` used a stripped inline `{src, file}` — so `get`, `find`, `samples` and `metaboliteIntensity2bins` were untyped on the live path. Collapsed to one properly-typed `MetaboliteIntensityQuery`.

Small stuff along the way: renamed orphaned `genome` params to `_genome`, fixed the singleCell README (it documented a `DEgenes.src` dispatch that hadn't existed in a while), a test fixture, and two dead comments referencing types that no longer exist.

### Testing
- tsc clean across shared/types, server, client. `test:unit` 1637/1637 server, 3130/3130 client.
- New `samplesRoute.unit.spec.ts` covers the init guards the type union used to enforce. Mutation-checked: deleting the `validateDataPlots()` call fails exactly the two intended assertions.
- Ran it against live GDC: single-cell plots (data / DE genes / gene expression) all render, and GRIN2 on a breast/ductal-lobular cohort came back identical to master (2,829 samples, 133,952 lesions, same top genes and p/q values). That's the path that used to go through useFilter0.

### ⚠️ Needs to ship with the ppgdc image
This drops the compatibility flags that let an older ppserver self-wire GDC, and it now also moves GDC's client behaviors and its caching gate into ds-declared flags. An older ppgdc image on this server would lose them — including the "server has not finished caching the case IDs" gate. The ppgdc image has to go out with this release, not after it. Companion PR is stjude/sjpp `gdc.sjpp.8`. Publish the ppserver image and point `ppgdc/Dockerfile`'s `ARG VERSION` at it in the same series.

## Checklist


siosonel:
which file got split into 6 modules?

compbiolover:
> which file got split into 6 modules?

It was `queries.ts`

shared/types/src/dataset.ts
```diff
@@ -89,8 +89,6 @@ type bcfMafFile = {
 }
```
xzhou82, line 89: can drop `useFilter0?: true` from lines 2088 and 2102

compbiolover: Removed in [e729867](https://github.com/stjude/proteinpaint/pull/4767/commits/e72986708e87589b052e9c2689fbcdecaffa6c3e)

xzhou82 (approved):

PR #4817, opened 2026-08-06
wrap all ds init steps in a try-catch block

siosonel:
# Description

This branch catches dataset preliminary validation errors like missing file (likely due to misspelled file paths or names). 

Tested with:
- have at least 2 dataset entries, then misspell one of the entries (such as `./dataset/gdc/index.ts-000`) in `serverconfig.json`, the restart should not crash in this branch but crashes in master
- do the above but with only 1 dataset entry: the server should crash, because there is no other dataset to load so the validation error should stop the startup. For GDC and MMRF container where there is only one dataset loaded, we want the server to crash in this case to trigger devops notification.

## Checklist

### DIFF
--- a/server/src/initGenomesDs.js
+++ b/server/src/initGenomesDs.js
@@ -431,6 +431,7 @@ export async function initGenomesDs(serverconfig, opts = {}) {
 		*/
 		g.datasets = {}
 		for (const d of g.rawdslst) {
+			let ds
 			// wrap ds init execution in a try-catch, to not crash when at least 1 dataset loaded successfully
 			try {
 				/*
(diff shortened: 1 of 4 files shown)


server/src/initGenomesDs.js
```diff
@@ -497,6 +512,29 @@ export async function initGenomesDs(serverconfig, opts = {}) {
...
+					the dataset js file is imported and evaluated, such as when its exported function throws.
+					An untracked ds must therefore be replaced with a stub, so that
+					- mayRetryInit() below does not crash on an undefined ds/ds.init, which would defeat this try-catch
+					- the failure is still tracked and reported in the startup log, optional slack notification,
+					  and the /healthcheck response, instead of the dataset silently disappearing
+				*/
+				if (!trackedDatasets.includes(ds)) {
+					// the error happened before the ds object was ready to be init'ed, so retrying init() against
+					// it would not be meaningful; use a stub that fails without retries, and reuse the same label
+					// so that mayRetryInit() also deletes any reference under g.datasets{}
```
siosonel, line 526: this is good edge-case coverage

xzhou82 (approved):

PR #4822, opened 2026-08-08
fix: stopping unrelated mass actions (like closing a sandbox) from aborting in-flight plot requests

xzhou82:


siosonel:
@xzhou82 all plots that use `this.vocabApi` must be an instance of a class that extends `PlotBase`, since the PlotBase constructor will create `this.vocabApi`, otherwise it will be missing.

client/rx/src/ComponentApi.ts
```diff
@@ -64,6 +66,15 @@ export class ComponentApi {
 		if (!self.type) self.type = __Class__.type
 		self.app = opts.app
 		self.api = this
+		// Give every component a vocabApi whose requests are signaled by this component's abort
+		// controller, instead of the app-level one that AppApi.dispatch() may abort on behalf of
+		// all components. A component that uses it will have its pending requests canceled when
+		// its own state changes, but not when an unrelated action, such as closing another plot's
+		// sandbox, is dispatched. PlotBase assigns its own in its constructor, which runs above,
+		// so do not overwrite it.
```
siosonel, line 74: The preferred, more direct approach is for a plot class to extend PlotBse, which creates `this.vocabApi`. Below creates an instance property that is more surprising, it's not obvious how this property got added, whereas it's more obvious when the parent class' constructor adds this.vocabApi(). There may be other useful properties or methods that will be inherited from or added by PlotBase.

PR #4834, opened 2026-08-11
fix: GDC BAM slice download will not save 403 error text as a .bam file

xzhou82:
tested on local: http://localhost:3000/?gdcbamslice=1&gdc_id=00115367-8271-47b9-a987-fa1c59a7bd65&gdc_ssm=R132H&stream2download=1

to avoid chrome CORS, launch a temp chrome by disabling cors to test above url
```
% open -na "Google Chrome" --args --disable-web-security --user-data-dir=/tmp/chrome-nocors
```

### DIFF
--- a/client/gdc/bam.js
+++ b/client/gdc/bam.js
@@ -169,6 +169,11 @@ export async function bamsliceui(
 	// show block & bam tk
 	const blockHolder = holder.append('div').style('display', 'none')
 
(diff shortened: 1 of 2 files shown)


siosonel:
After clicking on an Entity ID, I got an `Access Alert ... not authorized` message.  This message does not disappear when I click on a different ID - should the `Submit` button be re-shown or not?

siosonel (approved):

PR #4853, opened 2026-08-13
fix: matrix row-label editing is re-enabled. for geneVariant rows, al…

xzhou82:
…low per-row tw.q customizaton

# Description

these tentative changes allow to create multiple matrix rows like 'kras G12D', 'kras G12V' each using its own tw.q
please evaluate

## Checklist


siosonel:
@xzhou82 EDIT: I found the variant filer under groups ~~The editing of a gene label works. Is there also supposed to be new edit UI inputs to limit the gene to specific isoforms/variants?~~

siosonel (approved):
Approving for now. But I think `valueFilter` should be used for any term type, for terminology consistency. Having a `variantFilter` for a geneVariant term is like having `categoryFilter` for categorical term or `numericFilter` for numeric terms, the field is really filtering the values that are rendered, regardless of what the value is specifically called for a given term type.

PR #4922, opened 2026-08-27
Fix sc integration test and support e2e test fixes

creilly8:
# Description

The server-side code for the single cell assumes that all file names will be the sample ID, not the patient. This simple fix renames the UMAP and gene expression file to a sample ID. This change fixes the commented out SC integration test and [broken TermdbTest sc e2e tests](https://github.com/stjude/sjpp/pull/1519). 

## Checklist


client/plots/scatter/test/scatter.integration.spec.ts
```diff
@@ -609,7 +609,7 @@ tape('Single cell scatter properly renders when colorTW = scct term', function (
 				{
 					chartType: 'sampleScatter',
 					colorTW: getScctTw(),
-					singleCellPlot: { name: 'UMAP', sample: { sID: '1_patient' } }
+					singleCellPlot: { name: 'UMAP', sample: { sID: '2646' } }
```
ojmel: Would it be possible to put this sampleID some place central like in singlecell scatter types and then just import it to all these places?

ojmel (approved):

client/plots/scatter/test/scatter.integration.spec.ts
creilly8: @ojmel, I created a new `shared/utils/src/testData.ts` to shared test variables, type, and functions between the client and e2e tests. You can use this file in the future to do the same.

